PR #1, opened 2022-12-16
Try to stay in 100 columns

mattkretz:
Limiting to 100 (plus some slack) typically removes the need for vertical scrolling, making it more readable in GitHub patch reviews, copy&paste to Mattermost, etc. and generally makes reading those long lines easier.

To actually make broken long lines easier to read, enable AlignAfterOpenBracket.

To find function names quickly enable AlwaysBreakAfterReturnType. Especially with long return types (or decl-specifiers) it otherwise can take way too long to find the function name.

Additionally, keep template heads on their own lines. I would equally prefer to have requires clauses on their own lines, but we'd need clang 15.

ChangeLog:

	* .clang-format: Change rules to insert readable breaks in otherwise long lines.
	* include/graph.hpp: Re-apply clang-format.
	* include/typelist.hpp: Ditto.
	* src/main.cpp: Ditto.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -16,7 +16,7 @@
 ---
 Language: Cpp
 AccessModifierOffset: -4
-AlignAfterOpenBracket: DontAlign
+AlignAfterOpenBracket: Align
 AlignConsecutiveAssignments: AcrossEmptyLines
 AlignConsecutiveDeclarations: AcrossEmptyLines
 AlignEscapedNewlines: DontAlign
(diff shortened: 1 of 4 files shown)


wirew0rm:
> Additionally, keep template heads on their own lines. I would equally prefer to have requires clauses on their own lines, but we'd need clang 15.

According to https://github.com/restyled-io/restylers/blob/v0.222.0/clang-format/Dockerfile this is already the default for restyled.io, so we could update our .restyled.yaml to use the newest configuration.

PR #2, opened 2022-12-16
Remove merged output ports

mattkretz:
This is the correct default because the goal of merging nodes is to improve performance. If all intermediate results were returned (passed on) then the optimization would be greatly reduced.

ChangeLog:

	* src/main.cpp (main): Adjust for different return type of merged node. (duplicate): New node type that copies its input to `Count` outputs.
	* include/graph.h (any_node): Move into detail namespace. Implement the concept, though the check for process_one does not compile yet. (node): Make default constructor protected (never to be constructed directly). (merged_node): Remove connected output port from output_ports. (merged_node::process_one): Move the result from the connected output port from apply_left into apply_right and remove it from the tuple returned from the function.

### DIFF
--- a/include/graph.hpp
+++ b/include/graph.hpp
@@ -8,9 +8,6 @@
 #include <tuple>
 
 namespace fair::graph {
-template<typename T>
-concept any_node = true;
-
 enum class port_direction { in, out };
 
 template<port_direction D, int I, typename T>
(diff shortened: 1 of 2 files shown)


include/graph.hpp
```diff
@@ -66,35 +63,50 @@ class node {
...
+                                 meta::remove_at<InId, typename Right::input_ports::typelist>>>,
+            make_output_ports<meta::concat<meta::remove_at<OutId, typename Left::output_ports::typelist>,
+                                           typename Right::output_ports::typelist>>>;
 
     Left  left;
     Right right;
 
     template<std::size_t... Is>
-    constexpr auto
+    [[gnu::always_inline]] constexpr auto
```
RalphSteinhagen, line 103: This may not be portable for all compilers... just our favorite ones. ;-)

include/graph.hpp
```diff
@@ -66,35 +63,50 @@ class node {
...
     apply_left(auto &&input_tuple, std::index_sequence<Is...>) {
         return left.process_one(std::get<Is>(input_tuple)...);
     }
 
     template<std::size_t... Is, std::size_t... Js>
-    constexpr auto
+    [[gnu::always_inline]] constexpr auto
```
RalphSteinhagen, line 109: as above

RalphSteinhagen (approved):
nice! As discussed: if the info of one merged/removed port is required, it's information could be duplicated and explicitly exposed by a simple 1-in-2-out copy block.

PR #3, opened 2022-12-16
Add generic process_batch with automatic explicit SIMD

mattkretz:
The new node::process_batch function calls process_one for processing as many elements as given via the input ranges. The function expects all input ranges to be of the same size. This is encoded as a pre-condition in the function body.

The first argument to process_batch is a buffer that provides simple span<return_type> access. This is still an *incorrect* solution since return_type is a std::tuple of all output port types. A correct solution provides access to a span per type in the return_type tuple.

If all input and output ranges are contiguous ranges, and if all input and output port types are vectorizable (in the std::experimental::simd definition), then process_batch calls process_one with simd<T> arguments instead of T. This is still missing a correct constraint on the process_one function, testing whether it actually works with simd arguments.

ChangeLog:

	* src/main.cpp (adder::process_one): Accept simd<T, Abi> or T.
	(scale::process_one): Accept simd<T, Abi> or T.
	* include/graph.h (detail::precondition): New.
	(port_data): New.
	(node::process_batch_simd_epilogue): New.
	(node::process_batch): New.
	(merged_node::process_one): Accept simd<T, Abi> or T.
	* include/typelist.h (meta::transform_types): New.
	(meta::transform_value_type): New.
	(meta::reduce): New.
	(typelist::construct): New.

### DIFF
--- a/include/graph.hpp
+++ b/include/graph.hpp
@@ -8,9 +8,6 @@
 #include <tuple>
 
 namespace fair::graph {
-template<typename T>
-concept any_node = true;
-
 enum class port_direction { in, out };
 
 template<port_direction D, int I, typename T>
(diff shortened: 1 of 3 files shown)


mattkretz:
see my next PR

PR #5, opened 2022-12-16
added lambda_wrapper to help concept constraint

RalphSteinhagen:
N.B. return type isn't constraint but using std::same_as<decltype(...)> seemed wrong


RalphSteinhagen:
fix to be entered in different PR

PR #7, opened 2022-12-19
Add generic process_batch

mattkretz:
This adds `node::process_batch` showing how to build a generic implementation that builds on specific `process_one` implementations. The implementation detects whether `process_one` can be compiled with `stdx::simd` and whether all inputs and outputs are suitable, in which case it will explicitly vectorize the loop.

### DIFF
--- a/include/typelist.hpp
+++ b/include/typelist.hpp
@@ -235,9 +235,9 @@ struct typelist {
     static constexpr inline bool each_convertible_from = (std::convertible_to<Other, Ts> && ...);
 
     template<typename F, typename Tup>
-    requires(sizeof...(Ts)
-             == std::tuple_size_v<
-                     std::remove_cvref_t<Tup>>) static constexpr auto construct(Tup &&args_tuple) {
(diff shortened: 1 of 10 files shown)


RalphSteinhagen (approved):
@mattkretz very impressive and nice implementation that works across gcc (guaranteed full SIMD functionality), clang (guaranteed when using stdlibc++) and Emscripten (compiler likely generates SIMD instructions)! Kudos! :+1:

include/typelist.hpp
```diff
@@ -189,8 +233,18 @@ struct typelist {
 
     template<typename... Other>
     static constexpr inline bool each_convertible_from = (std::convertible_to<Other, Ts> && ...);
-};
 
+    template<typename F, typename Tup>
+        requires(sizeof...(Ts) == std::tuple_size_v<std::remove_cvref_t<Tup>>)
+    static constexpr auto
+    construct(Tup &&args_tuple) {
+        return std::apply(
```
ivan-cukic, line 241: Can this be converted to std::invoke?

PR #12, opened 2022-12-23
Small refactor of the graph facilities

ivan-cukic:


include/graph.hpp
```diff
@@ -127,13 +127,6 @@ concept node_can_process_simd
               } -> detail::any_simd<typename Node::return_type>;
           };
 
-// Workaround bug in Clang:
-// std::is_constructible should be a valid template argument for the template template parameter
-// `template <typename> class T`, but because std::is_constructible<T, Args...> has a parameter
-// pack, Clang considers the program ill-formed. As so often, another indirection solves the
-// problem. *Sigh*
-template <typename T>
-using is_constructible = std::is_constructible<T>;
```
RalphSteinhagen, line 136: Is this not necessary any more?

RalphSteinhagen (approved):
Looks good! :+1:

include/graph.hpp
ivan-cukic: No, changed the place where it was used to template <template <tpyename *...* >> so that it is not needed to create these /bugfixes/ for all non-unary meta functions.

PR #13, opened 2022-12-29
flow-graph additional edge, dyn_port, and basic block mechanics

RalphSteinhagen:
* exposed buffer via get_buffer() in buffer_[reader, writer]
  N.B. should be safe since the internal buffer implementation's life-time is controlled via a shared_ptr
* added basic block and edge API definitions
   N.B. further changes should be made once the `class port` is integrated witht the block `class node` implementation

N.B. regarding life-time/ownership: the `dyn_port` is a reference wrapper implementing runtime type polymorphism that is only needed during the initial inter-block-connections setup. The ports are expected to be strongly referenced by the block API and the buffer are already strongly referenced by the output and input ports (via shared_ptr).

Thanks a lot to @ivan-cukic for the iteration on the design and object lifetime management concept as well as the pre-review!

Co-authored-by: Ivan Čukić <ivan.cukic@kdab.com>


include/circular_buffer.hpp
```diff
@@ -217,24 +217,22 @@ class circular_buffer
...
-        alignas(kCacheLine) DependendsType&             _read_indices;
         alignas(kCacheLine) std::size_t                 _size;
         alignas(kCacheLine) std::vector<U, Allocator>*  _data;
         alignas(kCacheLine) ClaimType*                  _claim_strategy;
 
     public:
         buffer_writer() = delete;
         buffer_writer(std::shared_ptr<buffer_impl> buffer) :
-            _buffer(buffer), _is_mmap_allocated(_buffer->_is_mmap_allocated), _read_indices(buffer->_read_indices),
+            _buffer(buffer), _is_mmap_allocated(_buffer->_is_mmap_allocated),
```
ivan-cukic, line 231: `_buffer(std::move(buffer))`

include/circular_buffer.hpp
```diff
@@ -243,29 +241,29 @@ class circular_buffer
             _is_mmap_allocated = _buffer->_is_mmap_allocated;
             _size = _buffer->_size;
             _data = std::addressof(_buffer->_data);
-            _read_indices = _buffer->_read_indices;
             _claim_strategy = std::addressof(_buffer->_claim_strategy);
 
             return *this;
         }
 
+        [[nodiscard]] constexpr BufferType get_buffer() const noexcept { return circular_buffer(_buffer); };
```
ivan-cukic, line 249: We don't usually use `get_` on getters, but I'm fine either way for this function.

include/graph_contracts.hpp
```diff
@@ -0,0 +1,444 @@
...
+            assert(false && "invalid static_cast from 'void* -> WriterType*' in update_reader_internal()");
+            return false;
+        }
+        return true;
+    }
+    friend dyn_port;
+
+    public:
+        constexpr port() = default;
+        constexpr port(std::string_view port_name, std::int16_t priority = 0, std::size_t n_history = 0,
```
ivan-cukic, line 143: It is a pessimization to convert to string_view  and then back to std::string when one passes a rvalue std::string for port name.

Receive by value, and std::move into the member.

include/graph_contracts.hpp
```diff
@@ -0,0 +1,444 @@
...
+                       std::size_t min_samples = 0U, std::size_t max_samples = SIZE_MAX) noexcept:
+            _port_name(std::string(port_name)), _priority{priority}, _n_history(n_history),
+            _min_samples(min_samples), _max_samples(max_samples) {
+            static_assert(PortName.empty(), "port name must be exclusively declared via NTTP or constructor parameter");
```
ivan-cukic, line 147: How does this assert work if it is meant to check emptiness of "constructor parameter"?

include/graph_contracts.hpp
```diff
@@ -0,0 +1,444 @@
...
+        }
+
+        [[nodiscard]] constexpr static port_type_t type() noexcept { return PortType; }
+        [[nodiscard]] constexpr static port_direction_t direction() noexcept { return PortDirection; }
+
+        template<bool enable = !PortName.empty()>
+        [[nodiscard]] static constexpr std::enable_if_t<enable,decltype(PortName)> static_name() noexcept { return PortName; }
+
+    [[nodiscard]] constexpr supported_type pmt_type() const noexcept { return T(); }
```
ivan-cukic, line 156: Indentation

include/graph_contracts.hpp
```diff
@@ -0,0 +1,444 @@
...
+                return MAX_SAMPLES;
+            }
+        }
+
+        [[nodiscard]] constexpr connection_result_t resize_buffer(std::size_t min_size) noexcept {
+            if constexpr (IS_INPUT) {
+                static_assert(IS_INPUT, "can only resize output buffers");
+            } else {
+                try {
+                    if constexpr (IS_INPUT) {
```
ivan-cukic, line 196: This condition is always false as we're in the `else` branch of `if constexpr (IS_INPUT)`

include/graph_contracts.hpp
```diff
@@ -0,0 +1,444 @@
...
+
+        void setBuffer(gr::Buffer auto buffer) {
+            if constexpr (IS_INPUT) {
+                _ioHandler = std::move(buffer.new_reader());
+                _connected = true;
+            } else { _ioHandler = std::move(buffer.new_writer()); }
+        }
+
+        [[nodiscard]] ReaderType& reader() noexcept {
+            if constexpr (PortDirection == port_direction_t::OUTPUT) {
```
ivan-cukic, line 216: `not IS_INPUT`?

include/graph_contracts.hpp
```diff
@@ -0,0 +1,444 @@
...
+                static_assert(helper::always_false<T>, "reader() not applicable for outputs (yet)");
+            }
+            return _ioHandler;
+        }
+
+        [[nodiscard]] WriterType& writer() noexcept {
+            if constexpr (PortDirection == port_direction_t::INPUT) {
```
ivan-cukic, line 223: Wrong condition (acc to the static assert below)

include/graph_contracts.hpp
```diff
@@ -0,0 +1,444 @@
...
+
+        // internal runtime polymorphism access
+        [[nodiscard]] virtual bool update_reader_internal(void* buffer_other) noexcept = 0;
+    };
+
+    std::unique_ptr<model> _accessor;
+
+    template<Port T>
+    class wrapper final : public model {
+    using PortType = std::decay_t<T>;
```
ivan-cukic, line 277: Indentation

include/graph_contracts.hpp
```diff
@@ -0,0 +1,444 @@
...
+    std::vector<dyn_port> _input_ports{};
+    std::vector<dyn_port> _output_ports{};
+    setting_map _exec_metrics{}; //  →  std::map<string, pmt> → fair scheduling, 'int' stand-in for pmtv
+
+public:
+    explicit block(std::string_view name) : _name(name) {};
+
+    [[nodiscard]] std::string_view name() const noexcept { return _name; }
+    void work() noexcept { /* to be implemented in derived classes */ }
+    [[nodiscard]] std::vector<dyn_port>const & input_ports() const noexcept { return _input_ports; }
```
ivan-cukic, line 342: Not a fan of giving access to the underlying structure to user.  IMO, this should return std::span, and all modifications should go through proper API, not through vector's API.

include/graph_contracts.hpp
```diff
@@ -0,0 +1,444 @@
...
+    [[nodiscard]] setting_map const &exec_metrics() const noexcept { return _exec_metrics; }
+
+//    { t.connect(src_port, dst_block, dst_port) } -> std::same_as<std::vector<connection_result_t>>;
+};
+static_assert(Block<block>);
+
+
+class edge {
+    block& _src_block;
+    dyn_port& _src_port;
```
ivan-cukic, line 357: I'm now even less of a fan of references for member than I was before. If a block or a port goes away, this ends up dangling. If dyn_port contained shared_ptr instead of unique_ptr, this could have been a proper value (or a weak pointer).

include/graph_contracts.hpp
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    [[nodiscard]] constexpr int32_t weight() const noexcept { return _weight; }
+    [[nodiscard]] constexpr std::string_view name() const noexcept { return _name; }
+    [[nodiscard]] constexpr bool connected() const noexcept { return _connected; }
+    [[nodiscard]] connection_result_t connect() const noexcept { return connection_result_t::FAILED; }
+    [[nodiscard]] connection_result_t disconnect() const noexcept { return _dst_port.disconnect(); }
+};
+
+[[nodiscard]] connection_result_t add_edge(std::vector<edge>& edges, block& src_block, dyn_port& src_port, block& dst_block, dyn_port& dst_port, int32_t weight = 0, std::string_view name = "unnamed edge") noexcept {
```
ivan-cukic, line 384: Similar comment to above - edges should not be in a plain vector, IMO this needs a class with a proper API

include/graph_contracts.hpp
RalphSteinhagen: This check is to check that the name is exclusively defined either by the NTTP or the constructor, but not both.

include/graph_contracts.hpp
RalphSteinhagen: Thought of explicitly stating that it is an output.... strictly speaking there are three states: INPUT, OUTPUT, ANY

The later is only used for query arguments.

include/graph_contracts.hpp
RalphSteinhagen: done

include/circular_buffer.hpp
RalphSteinhagen: done

include/graph_contracts.hpp
RalphSteinhagen: done

include/graph_contracts.hpp
RalphSteinhagen: removed dead branch

include/graph_contracts.hpp
RalphSteinhagen: fixed the string description

include/graph_contracts.hpp
RalphSteinhagen: done

include/graph_contracts.hpp
RalphSteinhagen: changed and shielded the access via explicit `add_port(..)` and `port<INPUT(...)` accessor that can be queried via string and/or id (N.B. should be deprecated for user-api because indices are error prone). Also having no removal operation should make it safer in general w.r.t. initialisation and life-time management.

include/graph_contracts.hpp
RalphSteinhagen: changed to shared_ptr and internal std::size_t-based ID

include/graph_contracts.hpp
RalphSteinhagen: done.

include/circular_buffer.hpp
RalphSteinhagen: changed to `buffer()` as suggested.

include/graph_contracts.hpp
ivan-cukic: Ok, I misread it

ivan-cukic (approved):

PR #14, opened 2022-12-30
basic benchmarking facilities

RalphSteinhagen:
follows style and re-uses components from

UT: C++20 μ(micro)/Unit Testing Framework by Kris Jusiak (SPDX-License-Identifier: BSL-1.0)

original example, see [here](https://github.com/boost-ext/ut/blob/7c667d99716eb25db3fabb0d9114c8ad46ad7464/example/benchmark.cpp).

The simple use only requires:
```cpp
using namespace boost::ut;
using namespace benchmark;

"string creation1"_benchmark = [] {
    std::string created_string{"hello"};
    do_not_optimize(created_string);
};

"using marker"_benchmark.repeat<10>() = [] (MarkerMap<"source", "sink1", "sink2", "finish">& marker) {
    std::barrier start(3, [&marker] { marker.at<"source">().now(); }); // 1 source + 2 sinks
    std::barrier finish(3, [&marker] { marker.at<"finish">().now(); }); // 1 source + 2 sinks
    std::jthread source([&start, &finish] {
        start.arrive_and_wait();
        finish.arrive_and_wait();
    });
    std::jthread sink1([&start, &marker, &finish] {
        start.arrive_and_wait();
        std::this_thread::sleep_for(std::chrono::microseconds(500));
        marker.at<"sink1">().now();
        finish.arrive_and_wait();
    });
    std::jthread sink2([&start, &marker, &finish] {
        start.arrive_and_wait();
        std::this_thread::sleep_for(std::chrono::microseconds(2));
        marker.at<"sink2">().now();
        finish.arrive_and_wait();
    });
    source.join();
    sink1.join();
    sink2.join();
};
```

with the output presently looking something like:
image
 
 Will add a few more features next year:
   * [x] multiple iterations & computations of statistics (i.e. r.m.s., min, max, median) and basic latency histograms
   * [x] intermediate latency marker that can be "sprinkled" in the test code (N.B. essentially taking time stamps) -> detect in concurrent code which thread/function is called first.
   * [ ] aim: correcting for looping/noop time measurement instrumentation bias
   * further ideas:
      * cycle/asm instruction counting
      * cache misses (N.B. see [here](https://man7.org/linux/man-pages/man2/perf_event_open.2.html) for details)

### DIFF
--- a/bench/benchmark.hpp
+++ b/bench/benchmark.hpp
@@ -18,10 +18,10 @@
 
 #if __has_include(<unistd.h>)  && __has_include(<sys/ioctl.h>)  && __has_include(<sys/syscall.h>) && __has_include(<linux/perf_event.h>)
 #define HAS_LINUX_PERFORMANCE_HEADER
-#include <unistd.h>
+#include <linux/perf_event.h>
 #include <sys/ioctl.h>
 #include <sys/syscall.h>
-#include <linux/perf_event.h>
+#include <unistd.h>
 #endif
 
 namespace benchmark {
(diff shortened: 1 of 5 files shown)


RalphSteinhagen:
added optional [Linux performance counter](https://man7.org/linux/man-pages/man2/perf_event_open.2.html) support:

image

bench/benchmark.hpp
```diff
@@ -0,0 +1,835 @@
...
+
+        struct {
+            std::size_t pass{};
+            std::size_t fail{};
+        } _asserts{};
+        TPrinter _printer{};
+
+    public:
+        constexpr reporter &operator=(TPrinter printer) {
+            _printer = static_cast<TPrinter &&>(printer);
```
ivan-cukic, line 762: std::move?

bench/benchmark.hpp
```diff
@@ -0,0 +1,835 @@
...
+        } _benchmarks{};
+
+        struct {
+            std::size_t pass{};
+            std::size_t fail{};
+        } _asserts{};
+        TPrinter _printer{};
+
+    public:
+        constexpr reporter &operator=(TPrinter printer) {
```
ivan-cukic, line 761: IMO, assignment should set all fields. When one reads
```
a = x;
b = x;
```
it would be intuitive to assume `a == b`

ivan-cukic (commented):
Didn't comment on formatting

bench/bm_case1.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+    "string creation3"_benchmark = [] {
+        std::string created_string;
+        created_string = "hello";
+        do_not_optimize(created_string);
+    };
+
+    "string creation4"_benchmark.repeat<10'000>() = [] {
+        std::string created_string;
+        created_string = "hello";
```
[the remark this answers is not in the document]
RalphSteinhagen: These are just tests to show the many ways how not to declare a `std::string`. :laughing: 

They are just place holders and will be replaced by the graph benchmarks once we are ready to produce some vertical stacks.

bench/benchmark.hpp
RalphSteinhagen: It does, the `_printer` is the only relevant state. The other structs are just temporary private variables that are allocated once during runtime.

bench/benchmark.hpp
RalphSteinhagen: Noted, this was copied from the original UT source...  some may say that std::move is a glorified 'cast'.

Will also shift more of the printout mechanics from above to this class to better separate benchmark state/result from it's visualisation for future alternate representations. Just realised this when tinkering with some of the upstream UT code.

RalphSteinhagen:
view after update:
image

wirew0rm (approved):
very nice :-)
Did some small fixes on the CMake as discussed offline, otherwise lgtm! :+1:

PR #19, opened 2023-01-20
CI: use ubuntu toolchains ppa for gcc 12.2.0

wirew0rm:
Uses the ubuntu-toolchain [PPA](https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/ppa) which provides newer versions of gcc than the ones from the regular repositories.

Switches from `gcc 12.1.0-2ubuntu1~22.04` (the default on ubuntu 22.04LTS jammy) to `gcc 12.2.0-3ubuntu1~22.04`.

The newest ubuntu kinetic 22.10 would also provide gcc 12.2.0-3ubuntu1 and the ubuntu lunar development version provides gcc 12.2.0-10ubuntu1 but there is no offiicial github actions runner image for these releases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,7 @@ jobs:
     - name: Install gcc-12
       if: matrix.configurations.compiler == 'gcc'
       run: |
+        sudo add-apt-repository ppa:ubuntu-toolchain-r/ppa # provides newer gcc 12.2.0 instead of 12.1.0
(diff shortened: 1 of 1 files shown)


wirew0rm:
seems to work: https://github.com/fair-acc/graph-prototype/actions/runs/3966376661/jobs/6797110078#step:9:5

@ivan-cukic once this is merged, ci will run on gcc 12.2.0

PR #20, opened 2023-01-24
update single header generation script and add ci

wirew0rm:
Automates the generation of single file headers for every commit pushed to the main branch.

The CI will always create a pseudo merge commit in the single-header branch, such that the branch will contain a parallel history to the main branch where each commit links the previous singleheader as its first parent and the commit from main as its second parent.

All local includes must either use quote includes or add a ` localinclude` comment after the include.

Because of the way the combine script works, if a file gets included transitively and directly in a file, the transitive include has to come first.
``` c++
// if those 2 lines are switched, the include logic will not work
#include <a.hpp> // localinclude
#include <b.hpp> // localinclude
```
if `a.hpp` contains
```c++
#include <b.hpp> // localinclude
```

See [single-header branch](https://github.com/fair-acc/graph-prototype/blob/a20bac56f86d514aab8ce19971bc72609b89c8bf/README.md) and the compiler explorer example linked there.


RalphSteinhagen (approved):
Great! needs to be tested with the next merged PR. :+1:

PR #21, opened 2023-01-24
removed noexcept from connect(...)

RalphSteinhagen:
... since there is an exception thrown if the nodes are not registered

Signed-off-by: Ralph J. Steinhagen <r.steinhagen@gsi.de>


RalphSteinhagen:
pushed w/o review as it is a quick fix

PR #24, opened 2023-01-31
added optional BENCHMARK_NO_PERF_COUNTER

RalphSteinhagen:
... suppressing the Linux specific CPU/Kernel performance metrics even if the headers are available.

The existing version is already guarded for use in non-Linux environments but this can be set when it is known in advance that the user does not have sufficient rights to execute those kernel calls.

### DIFF
--- a/bench/benchmark.hpp
+++ b/bench/benchmark.hpp
@@ -16,7 +16,7 @@
 #include <fmt/color.h>
 #include <fmt/format.h>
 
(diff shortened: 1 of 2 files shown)


wirew0rm (approved):
LGTM.

Since the difference seems to only be only during setup, I wonder if an additional runtime flag (environment variable?) wouldn't be even more useful to quickly verify that the perfomance counters do not influence the benchmark results without having to recompile. But for now this solves the immediate problem.

PR #28, opened 2023-02-02
Allow ports to be defined as class member variables

ivan-cukic:


RalphSteinhagen (approved):
Thanks a lot @ivan-cukic ! LGTM. :+1: 

N.B. Most of the changes/iterations were done iteratively via discussing this in jitsi!

PR #29, opened 2023-02-03
reordered/removed simd header duplicates

RalphSteinhagen:
to enable single-header-only feature again


wirew0rm (approved):
:+1:  :crossed_fingers:

PR #32, opened 2023-02-05
Make graph owner of all nodes

ivan-cukic:
- nodes are created with graph::make_node function
- it is no longer needed to specify the graph when making connections

### DIFF
--- a/bench/bm_case1.cpp
+++ b/bench/bm_case1.cpp
@@ -380,8 +380,8 @@ inline const boost::ut::suite _runtime_tests = [] {
 
     {
         fg::graph flow_graph;
-        auto *src = flow_graph.make_node<test::source<float>>(N_SAMPLES);
-        auto *sink = flow_graph.make_node<test::sink<float>>();
+        auto &src = flow_graph.make_node<test::source<float>>(N_SAMPLES);
+        auto &sink = flow_graph.make_node<test::sink<float>>();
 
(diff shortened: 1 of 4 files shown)


RalphSteinhagen (approved):
LGTM and as discussed! :+1: 

TODO: We need to iterate with the GR 4.0 folks and @mormj how to handle the interface between block/node and uses of the API other than the graph/scheduler.

mormj:
I'm guessing from the python API we would have to do something like:

```python
fg = gr.graph()
blk1 = blocks.my_block(fg)
blk1 = blocks.my_block(fg)
```
where the `__call__` operator wraps `graph::make_node<my_block<...>>`

RalphSteinhagen:
@mormj sorry for missing context: @ivan-cukic and I had in the jitsi was regarding pointer vs. reference return types when registering/connecting nodes, e.g.
```cpp
auto& node = graph.make_node<multiply>(42.0);
````
in terms of life-time management. The question/discussion was about -- other than the graph container -- which components need to have access to the specific 'node'.

@mormj we briefly discussed/hinted at it but the open questions are how do nodes/blocks communicate, for example:
  *  hardware errors (cannot initialise/connect physical device, device errors)
  *  get/set of new settings and notifying that settings for a given node have changed (change via node, or indirectly via graph)
  *  registering/having call-backs in case blocks need UI/user input
  *  ...

There are several ways to do it ... should be simple, safe, and intuitive for a large user base...  :thinking:

PR #33, opened 2023-02-06
CI: use edlund/amalgamate for singleheader generation

wirew0rm:
Uses https://github.com/edlund/amalgamate to generate the single-header files instead of the custom logic based on the `gcc -H` output, which was brittle for nested headers.


RalphSteinhagen (approved):
LGTM!

PR #36, opened 2023-02-08
Added SIMD support to the work function

ivan-cukic:


RalphSteinhagen (approved):
Thank you @ivan-cukic for putting all the work in this clean-up iteration. Looks good!

As discussed the mixing of SIMD-operation with different scalar types and SIMD-widths would need to be discussed and re-evaluated with @mattkretz, but I think for now the present state is sufficient.

PR #6, opened 2022-12-19
CI: enable emscripten build

wirew0rm:
Add's a CI run compiling to emscripten and running the resulting wasm binary from node.

PR #38, opened 2023-02-12
refactoring and basic low-level library documentation

RalphSteinhagen:
... this is the first iteration to provide a minimum of basic low-level descriptions of the different library components. 
Target audience: experienced software engineers and library maintainers.

The more verbose, detailed, and general description should end up in either the README.md in the root directory or dedicated user-level documentation ...

Additions and further PR are welcome.


include/node.hpp
```diff
@@ -89,9 +88,93 @@ output_ports(Self *self) noexcept {
...
+ * references to its input and output 'ports' that can be zero, one, or many, depending on the use case.
+ * As the base class for all user-defined nodes, it implements common convenience functions and a default public API
+ * through the Curiously-Recurring-Template-Pattern (CRTP). For example:
+ * @code
+ * struct user_defined_block : node<user_defined_block> {
+ *   IN<float> in;
+ *   OUT<float> out;
+ *   // implement one of the possible work or abstracted functions
+ * };
+ * ENABLE_REFLECTION(user_defined_block, in, out);
```
ivan-cukic, line 102: The documentation is missing for the other nice macro which would work for the user-defined block with several template parameters like the one below (but with member ports).

include/node.hpp
```diff
@@ -427,43 +508,153 @@ class merged_node : public node<merged_node<Left, Right, OutId, InId>, meta::con
...
+ *    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠛⢶⢷⣿⡿⠛⠛⠛⢿⣦⣄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠁⠈⠊⠂⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+ *    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⢄⠀⠀⠀⠀⠀⠀⠉⠛⡿⠿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+ *    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠃⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
+ *
+ * The following macros are helpers to wrap around the existing refl-cpp macros: https://github.com/veselink1/refl-cpp
+ * The are basically needed to do a struct member-field introspections, to support
+ *   a) compile-time serialiser generation between std::map<std::string, pmt::pmtv> <-> user-defined settings structs
+ *   b) allow for block ports being defined a member fields rather than as NTTPs of the node<T, ...> template
+
+ * Their use is limited to the namespace scope where the block is defined (i.e. not across .so boundaries) and will be
```
ivan-cukic, line 653: Maybe explain the namespace issue when using the macro with an example how to define a block which lives in a namespace.

include/README.md
```diff
@@ -0,0 +1,399 @@
...
+The following is a brief introduction to the library architecture for people already vaguely familiar with GNU Radio,
+graph-based algorithms, or those simply looking for a quick-intro intro of the concepts and the 'where-is-what
+implemented'.
+For the 'how-it-is-implemented' please consult the provided links, code-references, and feel welcome to make PRs to
+continuously improve this document.
+
+## Overall Architecture
+
+![graph_core_structure_diagram_dark](./assets/graph_core_structure_dark.png#gh-dark-mode-only)
+![graph_core_structure_diagram_light](./assets/graph_core_structure.svg#gh-light-mode-only)
```
wirew0rm, line 12: Some questions/clarifications/suggestions:
- image differentiates between source/node/sink, the code doesn't. At the same time the domain argument looks like a template parameter, making it look a bit like code.
- data converters are visualized as blocks, which side (or both) call their work function or are they not really blocks? maybe add a bullet point for data_converter below?
- I'm not really sure I'm interpreting the lower(scheduler) part of the image correctly:
    - what do the arrows represent (calls/dataflow/?) Maybe add a verb to arrows or have different arrows for different semantics.
    - subgraph3 which is not present in the flowgraph below the nodes while the gpu scheduler is below the main scheduler, between the blocks and the compute.
    - I would expect the main scheduler to first call source1 (all other blocks seem to be topologically ordered)
    - subgraph3 which is not present in the flowgraph below the nodes while the gpu scheduler is below the main scheduler, between the blocks and the compute. (I know, it's optional, but I have the feeling that if it was in the graph I could maybe understand it better. Maybe having it greyed out in both the flow- and scheduler-graph would make it easier to understand, without overloading the image?)

wirew0rm (approved):
Very nice, I think this is very helpful for getting into the code. I found one small misstake and added some comments on the image trying to make it more accessible.

include/README.md
```diff
@@ -0,0 +1,399 @@
...
+  The number, type, and ordering of input and arguments of `process_one(..)` are defined by the port definitions.
+* **case 1b** - non-decimating N-in->N-out mechanic providing bulk access to the input/output data and automatic
+  handling of streaming tags and settings changes (to-be-completed):
+  ```cpp
+  template<typename T, typename R>
+  struct user_defined_block : node<user_defined_block, IN<T, 0, N_MAX, "in">, OUT<R, 0, N_MAX, "out">> {
+    T _factor = T{1};
+    // constuctor setting _factor etc.
+
+    [[nodiscard]] constexpr auto process_one(std::span<const T> input, std::span<R> output) const noexcept {
```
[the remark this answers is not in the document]
RalphSteinhagen: done.

include/README.md
RalphSteinhagen: > At the same time the domain argument looks like a template parameter, making it look a bit like code.

This will (eventually) be a node NTTP arguement that user can specialise to implementa block that works on the CPU, GPU, etc...

RalphSteinhagen: > data converters are visualized as blocks

Yes, they are blocks/node whose `work()` function needs to be explicitly called by the scheduler.

RalphSteinhagen: Re: subgraph3

Yes, this was in the plot (grouping blocks #2 and #6), but seems to cut into the buffer representation and overall overloaded the picture a bit.

Regarding scheduling: we have some more graph topologies that are more usefull to explain the topology and ordering of the schedule... will add this later to the section `Collection of Ideas and known Scheduling Algorithms` also explaining why there cannot be a one-fits-all scheduler. This would also require gant-type charts that one would see in a profiler etc.... requires a bit more graphical design.

For this PR and picture this would IMO a bit too much.

wirew0rm: all fine, figures are always some kind of tradeoff between completeness and complexity, i guess. I was just trying to document the points where I felt that I first got it wrong or took longer to understand.

> >data converters are visualized as blocks
>
> Yes, they are blocks/node whose work() function needs to be explicitly called by the scheduler.

ok that makes sense, so the domain boundary would be at one of its ports and the implementer of the converters can decide to which flowgraph it belongs to, in the shown example, the cpu/main scheduler would transfer the data to the gpu buffer so the data_converter would be a `node<CPU,cpu_buffer_port, gpu_buffer_port>`.

include/node.hpp
RalphSteinhagen: done... added a commented and fully documented w/ examples where the macros are being declared.

include/node.hpp
RalphSteinhagen: done below next to the macro definitions.

RalphSteinhagen: done next to the macros.

ivan-cukic (approved):

PR #40, opened 2023-02-13
added node<T>::process_bulk(std::span<>...) support

RalphSteinhagen:
Added missing support for non-decimating N-in->N-out mechanic providing bulk access to the input/output data and automatic handling of streaming tags and settings changes:

  ```cpp
  template<typename T, typename R>
  struct user_defined_block : node<user_defined_block, IN<T, 0, N_MAX, "in">, OUT<R, 0, N_MAX, "out">> {
    T _factor = T{1};
    // constuctor setting _factor etc.

    [[nodiscard]] constexpr auto process_bulk(std::span<const T> input, std::span<R> output) const noexcept {
        std::ranges::transform(input, output.begin(), [this](const T& elem) { return elem * _factor; });
    }
  };
  ```
  This could be also the basis for most other interpolating/decimating blocks.

### DIFF
--- a/bench/bm_case1.cpp
+++ b/bench/bm_case1.cpp
@@ -11,41 +11,45 @@
 
 #include <vir/simd.h>
 
-namespace fg = fair::graph;
+namespace fg                           = fair::graph;
 
-inline constexpr std::size_t N_ITER = 10;
+inline constexpr std::size_t N_ITER    = 10;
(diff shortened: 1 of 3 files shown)


ivan-cukic (approved):

wirew0rm (approved):
Looks good!

PR #41, opened 2023-02-15
Use do-not-optimize inline-asm for more realistic results

mattkretz:
1. If an inline-asm statement clobbers all memory then we don't see realistic benchmark measurements because the optimizer is inhibited from doing certain optimizations. The result is very different depending on the order. The constexpr cases suffered most because the optimizer was inhibited in every single iteration.

2. The source produces only zero-initialized samples. The optimizer was able to see that. And it may or may not have used that knowledge, rendering all previous results as quasi-random. Use inline asm to only inhibit const-prop, nothing else.

ChangeLog:

	* bench/benchmark.hpp (do_not_optimize): Optimal do-not-optimize implementation. Do not invalidate all memory in the process.
	* bench/bm_test_helper.hpp (source::process_one): Taint the constructed sample using benchmark::do_not_optimize.

### DIFF
--- a/bench/benchmark.hpp
+++ b/bench/benchmark.hpp
@@ -36,25 +36,37 @@ namespace benchmark {
 
 #if defined(__GNUC__) or defined(__clang__) and not defined(_LIBCPP_VERSION)
 
+#if defined(__x86_64__) or defined(__i686__)
+#define SIMD_REG "x,"
+#else
+#define SIMD_REG
+#endif
+
     template<typename T>
     void
     do_not_optimize(T const &val) {
-        // NOLINTNEXTLINE(hicpp-no-assembler)
-        asm volatile("" : : "r,m"(val) : "memory");
+        if constexpr (sizeof(T) >= 16 || std::is_floating_point_v<T>) {
+            // NOLINTNEXTLINE(hicpp-no-assembler)
+            asm volatile("" ::SIMD_REG "m"(val));
+        } else {
+            // NOLINTNEXTLINE(hicpp-no-assembler)
+            asm volatile("" ::"g,m"(val));
+        }
     }
 
     template<typename T>
     void
     do_not_optimize(T &val) {
-#if defined(__clang__)
-        // NOLINTNEXTLINE(hicpp-no-assembler)
-        asm volatile("" : "+r,m"(val) : : "memory");
-#else
-        // NOLINTNEXTLINE(hicpp-no-assembler)
-        asm volatile("" : "+m,r"(val) : : "memory");
-#endif
+        if constexpr (sizeof(T) >= 16 || std::is_floating_point_v<T>) {
+            // NOLINTNEXTLINE(hicpp-no-assembler)
+            asm volatile("" : "+" SIMD_REG "g,m"(val));
+        } else {
+            // NOLINTNEXTLINE(hicpp-no-assembler)
+            asm volatile("" : "+g,m"(val));
+        }
     }
 
+#undef SIMD_REG
 #else
 #pragma optimize("", off)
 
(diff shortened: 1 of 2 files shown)


mattkretz:
> The previous `do_not_optimize(..)` (which btw. is used in other popular micro-benchmarking test-suites

Yes, I know. And it bothers me. I've also seen a talk where the the presenter says one needs the memory clobber. But the explanation was "trust me". That's not enough on this topic which I have studied really extensively. Diplomarbeit, publications, PhD thesis - I've invested months of work into doing precise micro-benchmarking. If someone tells me I'm doing it wrong, and the supposed solution produces assembler that makes no sense to me, then I need a really good explanation...

>     * The old:
> 
> ```c++
> template<typename T>
> void do_not_optimize(T &val) {
>    asm volatile("" : "+m,r"(val) : : "memory");
> }
> ```
> 
> that indicates to read from and write to the memory location pointed to by the input parameter `val`.

No. It says 2 things:

1. It tells the compiler that this assembler code (i.e. the code `""`) reads from `val` *and* writes to it. It can do so either via memory (`"m"`: "A memory operand is allowed, with any kind of address that the machine supports in general.") or register (`"r"`: "A register operand is allowed provided that it is in a general register.")

2. It indicates read and write to any memory address the process has access to: "The `"memory"` clobber tells the compiler that the assembly code performs memory reads or writes to items other than those listed in the input and output operands (for example, accessing the memory pointed to by one of the input parameters). To ensure memory contains correct values, GCC may need to flush specific register values to memory before executing the asm. Further, the compiler does not assume that any values read from memory before an asm remain unchanged after that asm; it reloads them as needed. Using the "memory" clobber effectively forms a read/write memory barrier for the compiler."

>, while
> 
>     * the new:
> 
> 
> ```c++
> template<typename T>
> void do_not_optimize(T &val) {
>     asm volatile("" : "+x,g,m"(val));
> }
> ```
> 
> does not specify which specific register the assembler should use with the "+x" operand.

`"x"` means "Any SSE register". And floating-point is nowadays stored in SSE registers. But it is x86-specific, which is why I had to add preprocessor logic to make the emscripten build pass again.

> I.e. could be a register and/or memory location while the previous forced it to be a memory location which in the end the function needs to export to. So this seems to be an optimisation that favours how the constexpr test is setup. Need to think how to modify the test that at least during the first and last iteration the reading/writing to RAM is enforced, otherwise this may be an unfair optimisation w.r.t. the other runtime tests that guarantee that.

FWIW, I believe the `do_not_optimize` functions were properly named for what they did. But nobody should ever kill the optimizer for benchmarking :shrug:. Instead you want more expressive control like `fake_read(x)`, `fake_modify(x)`, `force_store(x)`, `force_to_memory(x)`. And then, you'd do a `fake_modify` in `source` and a `force_store` in `sink`.

> Anyway, after having checked with @mormj, it turns out that the GR 3.X and 4.0 benchmark tests didn't use any of these `do_not_optimize(..)` guards or similar which makes the modified test a more realistic comparison to the GR tests.

My experience: All benchmarks that are not *understood* on the assembly level are wrong. (Note, measuring timings in real applications is typically not a problem.)

> The `src->(mult(2.0)->div(2.0)->add(-1))^10->sink` and the `(...)^10` tests in general still aren't close to the constexpr performance one would intuitively expect (i.e. equal or maybe factor 2 slower compared to the single pass execution).

Yes. I checked the asm. I have no complaints. :shrug: I still need to figure out why those numbers come out as they do.

> Maybe we could also have a joint look whether the runtime examples are close to the peak-performance they could be... as mentioned the performance isn't bad (factors faster than previous GR implementation) but I would like to make sure that we didn't reintroduce an avoidable bottleneck before fleshing out more of the graph API.

Yes. All my investigations before fixing the optimization barriers were useless. I can basically only start to understand the issues now.

PR #42, opened 2023-02-23
Simplify work() implementation

mattkretz:
The process_one functions return void, a value, or a tuple of values. Changing this to always return a tuple from process_one makes implementing process_one harder than acceptable. So take a middle-ground and add a invoke_process_one function that regularizes the return type by wrapping the size 0 and 1 cases into tuples. As a result, the work() implementation, calling invoke_process_one instead of process_one, is much simpler.

include/ChangeLog:

	* node.hpp (node::invoke_process_one): New. Calls self().process_one and ensures the return type is a tuple. (node::work): Call invoke_process_one instead of process_one, simplifying the logic, eliminating the constexpr-if branches.
	* utils.hpp: Remove dummy_t and invoke_void_wrapped since they are not used anymore.

### DIFF
--- a/include/node.hpp
+++ b/include/node.hpp
@@ -290,6 +290,20 @@ public:
         return success;
     }
 
+    template <typename... Ts>
+    constexpr auto
+    invoke_process_one(Ts&&... inputs)
+    {
+        if constexpr (traits::node::output_ports<Derived>::size == 0) {
+            self().process_one(std::forward<Ts>(inputs)...);
+            return std::tuple{};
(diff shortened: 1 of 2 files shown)


ivan-cukic (commented):
LGTM

RalphSteinhagen (approved):
Nice, I like the code reduction and makes it IMO cleaner and easier to read. In terms of compile-time overhead (and why clang is giving troubles in the CI) should be tackled in another PR. LGTM. :+1:

PR #43, opened 2023-02-24
fix clang to 15.0.6 and added clang 16

RalphSteinhagen:
clang 15.0.7 seems to have introduced a compile bug


wirew0rm (approved):
Looks good :+1:

PR #44, opened 2023-02-24
fix sink accounting, refactor optimization barriers and benchmark impl

mattkretz:


RalphSteinhagen (approved):
LGTM! Thanks for the simplification/rationalisations ...

PR #45, opened 2023-02-24
Require sources to implement inputs_status

mattkretz:
node::inputs_status reports whether, and how much data is available on the output ports of a node. However, for sources there is no way to generically derive any value. The generic node implementation simply says that 1 sample is available for sources.

With this change source nodes are required to implement inputs_status. A static_assert tries to be helpful.

Merged nodes that have no inputs (i.e. act as a source) also implement inputs_status in order to forward the query to their left node, which is the internal source node.

The benchmark source node implements inputs_status. It has as many samples available as were requested and not produced yet.

ChangeLog:

	* bench/bm_test_helper.hpp (source::inputs_status): New.
	* include/node.hpp (node::inputs_status): Remove the source node implementation, which returned 1. (merged_node::inputs_status): New.


bench/bm_test_helper.hpp
```diff
@@ -22,6 +22,19 @@ class source : public fg::node<source<T, min, count>> {
 
     source(std::size_t n_samples) : _n_samples_max(n_samples) {}
 
+    template<typename Self>
+    [[nodiscard]] constexpr auto
+    inputs_status(Self &self) noexcept {
+        const auto avail = _n_samples_max - n_samples_produced;
+
+        const struct {
```
ivan-cukic, line 30: I'd propose naming this struct and moving it to node.hpp - easier to understand for the user 'Source nodes must implement input_status_t input_status(...)' than what is currently written in the static_assert.

include/node.hpp
```diff
@@ -229,27 +229,24 @@ class node : protected std::tuple<Arguments...> {
...
-                        [&availableForPort] (auto&... input_port) {
-                            return meta::safe_min(availableForPort(input_port)...);
-                        },
-                        input_ports(&self));
-            } else {
-                return 1_UZ;
-            }
+            const auto availableForPort = [&at_least_one_input_has_data]<typename Port>(Port &port) noexcept {
+                const std::size_t available = port.reader().available();
+                if (available > 0LU) at_least_one_input_has_data = true;
```
ivan-cukic, line 239: Since this code is updated, it is a nice opportunity to replace `LU` with `_UZ`

ivan-cukic (approved):
A few nit picks

bench/bm_test_helper.hpp
RalphSteinhagen: Needed to think about this a bit more but I concur with @ivan-cukic.

The user-facing API should be as simple and as crisp as possible.

Outside the benchmark/unit tests merging of source-type node is rare and probably never happens.
This makes the comparison with the more common 'runtime' use case difficult. 
With the exception of benchmark/unit tests, most of the sources also do not know on start-up how many samples they are going to produce this can be anything from 0 to infinity.

If the structure is needed/helpful to minimise internal boiler-plate code, I'd suggest making it a node internal implementation that expert users may override if necessary (1-2% use-case IMO).

mattkretz: Yes, I was too quick not to name this struct. Will do.

I believe it is wrong for all sources to (almost) unconditionally, always say "I can give you only one sample, no more, *no less* even".

I also believe that merging sources with a gain, normalize, etc. doesn't sound too far fetched, no?

Wrt. sources being able to answer the question at all: I think they should be able to say "well I have 1K in my hardware input buffer ready to read, so ... 256 samples ready to go". They don't have to say how many samples they will produce in their lifetime. It's just that 32 can be optimized so much better than 1.

If all this doesn't convince you, then there should not be any benchmarks that merge the source node. Because if they exist, you're comparing apples with oranges wrt. SIMD. That's misleading at best.

I believe the `inputs_status` could be simplified for nodes that don't need as much control (especially nodes with a single output port). These could simply return a single `size_t`. What about an `available_samples` function that is only called if the node has a single output port? The could also be a `static_assert` to ensure no multi-output-port node implements the function.

mattkretz: Trying to understand the implications wrt. `graph::work`: Do I understand correctly, that, without this change, a source node's `work` function is called once *per single sample*... through an indirect function call (virtual function)? And since `graph::work` cannot take shortcuts in its loop it calls `work` on *all* nodes once per sample? (Unless, the source implements `work`, `process_bulk`, or `inputs_status`.)

mattkretz: Hmm, even `process_bulk` shouldn't help because the generic `node::work` function only requests a single sample from output port writers?

RalphSteinhagen: > Wrt. sources being able to answer the question at all: I think they should be able to say "well I have 1K in my hardware input buffer ready to read, so ... 256 samples ready to go". They don't have to say how many samples they will produce in their lifetime. It's just that 32 can be optimized so much better than 1.

For the dynamic/runtime case (inside `work()` this is queried and checked by the `size_t N = buffer.[reader, writer()].available()` expression.

If this is like `in case source block produces data, you can expect at least N samples' then there is the `std::size_t MIN_SAMPLES = std::dynamic_extent` NTTP on the port that maybe could be reused for that. 

We shouldn't forget that the merge-API is already helpful because it avoids writing intermediate results to a runtime buffer.

RalphSteinhagen:
unrelated to the above:

Until now most of the nodes/blocks that are used in the merge-API examples have either settings that are NTTPs and/or are unique/fixed. I am wondering how to handle merging of blocks where -- for example -- the multiplication factor is different for the merged blocks. :thinking:

mattkretz:
> Until now most of the nodes/blocks that are used in the merge-API examples have either settings that are NTTPs and/or are unique/fixed. I am wondering how to handle merging of blocks where -- for example -- the multiplication factor is different for the merged blocks.

Yes, let's discuss this in its own place/issue. I had some ideas while working on my original PoC.

bench/bm_test_helper.hpp
mattkretz: I still don't understand. `graph::work` calls `node_wrapper<T>::work` which calls `T::work`. The latter is `node<Derived>::work` unless `Derived` implements `work`. If the source does *not implement* `work`, then the generic `node::work` implementation queries `inputs_status` which unconditionally returns `{ .at_least_one_input_has_data = false, .available_values_count = 1 }` for source nodes. The `at_least_one_input_has_data` value is ignored because `available_values_count` is non-zero. `work` then calls `output_port.writer().get(1)` for each output port. If the source node implements `process_bulk`, then `process_bulk(writers...)` is called and `available_values_count` are published to the output ports. Otherwise, `work` loops via `for (i = 0; i < 1; ++i)` over `process_one` calls and writes the result to the ports.
So, unless source nodes implement `work`, they produce exactly one sample with each iteration in `graph::work`.

mattkretz: To summarize my understanding:
* A source node must either implement `work` or `inputs_status`.
* If a source node is to be merged at compile time it must implement `inputs_status`.

Since `inputs_status` is a very misleading function to call on source nodes, how about we introduce a different function? E.g. `size_t available_samples()`? However, how can the generic `node::work` implementation distinguish `work_return_t::INSUFFICIENT_INPUT_ITEMS` vs. `work_return_t::DONE`? `available_samples() == 0` could mean either one (where `INSUFFICIENT_INPUT_ITEMS` is a bad name again).

include/node.hpp
```diff
@@ -306,27 +301,57 @@ class node : protected std::tuple<Arguments...> {
...
+                constexpr std::size_t chunk_size = Derived::merged_work_chunk_size();
+                static_assert(
+                        chunk_size != std::dynamic_extent && chunk_size > 0,
+                        "At least one internal port must define a maximum number of samples or the non-member/hidden "
+                        "friend function `available_samples(const NodeType&)` must be defined.");
+                samples_to_process = chunk_size;
+            } else {
+                // derive value from output buffer size
+                samples_to_process
+                        = std::apply([&](const auto &...ports) { return std::min({ ports.writer().available()... }); },
```
RalphSteinhagen, line 329: I like this approach ... the limit should also take into account the MIN/MAX limits that the user imposed onto the ports.

include/node.hpp
```diff
@@ -439,6 +462,37 @@ class merged_node : public node<merged_node<Left, Right, OutId, InId>, meta::con
     Left  left;
     Right right;
 
+    // merged_work_chunk_size, that's what friends are for
```
RalphSteinhagen, line 465: https://youtu.be/TWa03rUR9pA

include/node.hpp
mattkretz: true, will change

mattkretz: Questions:
1. Should `min_buffer_size()` only be checked for the `is_source_node` case?
2. Should it be checked if the `available_samples` customization point is implemented? If yes, how will the "EOF" case, where only a few samples are left, work?
4. If it's also checked for not-source-nodes, and the number of samples is determined by `inputs_status`, what's the return code on failure? `work_return_t::INSUFFICIENT_INPUT_ITEMS`, right? Whereas on a source node it's `work_return_t::INSUFFICIENT_OUTPUT_ITEMS`?

RalphSteinhagen: Re 2:

I think the existing/proposed solution of `if constexpr (requires(const Derived &d) {..} ` plus the other to if-else cases is IMO good as is.

RalphSteinhagen (approved):
LGTM! Thanks for the changes!

PR #46, opened 2023-02-24
Ensure that merged nodes do not lose can_process_simd

mattkretz:
include/ChangeLog:

* node.hpp: Add static assertions for can_process_simd on merged nodes. (merged_node::process_one): Reenable requires clause to disambiguate with simd overload.

There may be another way to disambiguate so that error messages can be improved again. I have not thought much about it, though.

### DIFF
--- a/include/node.hpp
+++ b/include/node.hpp
@@ -478,14 +478,11 @@ public:
     }
 
     template<typename... Ts>
-    // In order to have nicer error messages, this is checked in the function body
-    // requires input_port_types::template are_equal<std::remove_cvref_t<Ts>...>
+    // Nicer error messages for the following would be good, but not at the expense of breaking
+    // can_process_simd.
+        requires(input_port_types::template are_equal<std::remove_cvref_t<Ts>...>)
     constexpr return_type
     process_one(Ts &&...inputs) {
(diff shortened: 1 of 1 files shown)


ivan-cukic (approved):

include/node.hpp
```diff
@@ -478,14 +478,11 @@ class merged_node : public node<merged_node<Left, Right, OutId, InId>, meta::con
     }
 
     template<typename... Ts>
-    // In order to have nicer error messages, this is checked in the function body
-    // requires input_port_types::template are_equal<std::remove_cvref_t<Ts>...>
+    // Nicer error messages for the following would be good, but not at the expense of breaking
+    // can_process_simd.
+        requires(input_port_types::template are_equal<std::remove_cvref_t<Ts>...>)
```
RalphSteinhagen, line 483: Does this imply that all input port types have to be equal? 

Haven't checked this before but this would be a stronger constraint since there are cases where int-float, float/double, or float/std::complex input port types are mixed.

mattkretz: No, it tests that each input port type is equal to the corresponding type in the pack `Ts`.

RalphSteinhagen: Thanks!

PR #48, opened 2023-02-28
CMake: remove duplicated code and use targets consistently

wirew0rm:
- remove global include directories and use targets consistently
- define compiler warnings in one place only (deduplicated into cmake/CompilerWarnings.cmake)


RalphSteinhagen (approved):
LGTM!

PR #49, opened 2023-03-01
Basic meson build support

mormj:
This PR adds basic meson build support for graph-prototype (not everything is in place yet that exists in the CMake)

To build:
```bash
cd graph-prototype
meson setup build_meson --buildtype=release
# meson setup build_meson --buildtype=release -denable_testing=false # to disable testing
cd build_meson
ninja
```

Issues and TODO
- [x] Compiling in debug mode (not specifying buildtype, or buildtype=debug) causes a compiler segfault
- [x] WASM support
- [x] Warnings configuration from CompilerWarnings.cmake
- [ ] The if blocks around Apple/MSVC, etc.

### DIFF
--- a/include/meson.build
+++ b/include/meson.build
@@ -1,0 +1,23 @@
+header_files = [
+    'vir/detail.h',
+    'vir/simd_bit.h',
+    'vir/simd_bitset.h',
+    'vir/simd_cast.h',
+    'vir/simd_float_ops.h',
+    'vir/simd_resize.h',
+    'vir/simd.h',
+    'buffer_skeleton.hpp',
+    'buffer.hpp',
+    'circular_buffer.hpp',
+    'claim_strategy.hpp',
+    'graph.hpp',
+    'node_traits.hpp',
+    'node.hpp',
+    'port_traits.hpp',
+    'port.hpp',
+    'sequence.hpp',
+    'typelist.hpp',
+    'utils.hpp',
+    'wait_strategy.hpp'
+]
+install_headers(header_files, subdir : 'graph-prototype')
--- a/include/meson.build
+++ b/include/meson.build
@@ -1,3 +1,6 @@
+
+graph_dep = declare_dependency(include_directories: '.', dependencies : [fmt_dep, reflcpp_dep])
+
 header_files = [
     'vir/detail.h',
     'vir/simd_bit.h',
@@ -20,4 +23,4 @@ header_files = [
     'utils.hpp',
     'wait_strategy.hpp'
 ]
-install_headers(header_files, subdir : 'graph-prototype')
+install_headers(header_files, subdir : 'graph-prototype', preserve_path: true)
--- a/include/meson.build
+++ b/include/meson.build
@@ -1,5 +1,5 @@
 
-graph_dep = declare_dependency(include_directories: '.', dependencies : [fmt_dep, reflcpp_dep])
+graph_dep = declare_dependency(include_directories: '.', dependencies : [fmt_dep, reflcpp_dep, pmt_dep])
 
 header_files = [
     'vir/detail.h',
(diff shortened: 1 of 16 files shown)


include/meson.build
```diff
@@ -0,0 +1,23 @@
...
+    'graph.hpp',
+    'node_traits.hpp',
+    'node.hpp',
+    'port_traits.hpp',
+    'port.hpp',
+    'sequence.hpp',
+    'typelist.hpp',
+    'utils.hpp',
+    'wait_strategy.hpp'
+]
```
wirew0rm, line 22: This just defines where the headers have to be installed, but not their dependencies, which have to be specified along with the include path everywhere it is used (in this project that's just the tests and benchmarks, but looking at gnuradio and the modularisation we talked about, there will be more dependencies between subprojects).
In CMake this is solved for our header-only modules by defining INTERFACE targets, which bundle the include directories and their dependencies into a single target.
It seems that the meson way of doing this would be to define an internal dependency with [`declare_dependency()`](https://mesonbuild.com/Reference-manual_functions.html#declare_dependency) here.

Something like:

``` meson
graph_dep = declare_dependencies(include_directories : ".", dependencies : [fmt_dep, reflcpp_dep])
```

then in the tests/benchmarks we just specify `deps : [graph_dep, ut_dep]`. This also allows to restructure the project into sub-dependencies (e.g. use an external version of vir_simd) or move parts around without things falling apart or needing changes everywhere.

wirew0rm (commented):
Nice, thanks for tackling this, looks good so far.

Some points from my side adding to yours:
- [ ] hava a github action for meson so it doesn't break directly
- [ ] create an internal dependency for graph? (see comment)

include/meson.build
```diff
@@ -0,0 +1,23 @@
...
+install_headers(header_files, subdir : 'graph-prototype')
```
eli-schwartz, line 23: So e.g. `vir/detail.h` should be installed to the system as `/usr/include/graph-prototype/detail.h`? This seems wrong -- at least vir/ should keep its structure.

Meson 0.63.0 added support for that with `preserve_path: true`.

test/meson.build
```diff
@@ -0,0 +1,22 @@
...
+  include_directories : include_directories, 
+  dependencies : [fmt_dep, ut_dep],
+  install: false)
+test('qa_buffer', buffer_test_exe)
+
+ut_test_names = [
+    'qa_dynamic_port',
+]
+deps = [fmt_dep, ut_dep, reflcpp_dep]
+cpp_args = '-fsanitize=address'
```
eli-schwartz, line 13: Can you just use meson's builtin support for `-Db_sanitize=address` here?

You don't *have* to apply it to the entire project, by the way. Applying it to a single target as `override_options: ['b_sanitize=address']` works.

mormj: Thanks for the feedback @eli-schwartz !

meson.build
```diff
@@ -0,0 +1,17 @@
...
+  version : '0.1',
+  default_options : ['warning_level=3', 'cpp_std=gnu++20'])
+
+fmt_dep = dependency('fmt', version:'8.1.1')
+ut_dep = dependency('boost.ut')
+
+cmake = import('cmake')
+libreflcpp = cmake.subproject('refl-cpp')
+reflcpp_dep = libreflcpp.dependency('refl-cpp')
+
```
wirew0rm, line 11: I think we are missing the pmtv dependency here: https://github.com/fair-acc/graph-prototype/actions/runs/4902739549/jobs/8772423723?pr=49#step:11:12 Probably it gets picked up from the system path on your system, but isn't available on the CI runner. Also it might not have been a dependency when this PR was started :)

mormj: Added it in - but to get this building with wasm cleanly requires https://github.com/gnuradio/pmt/pull/90 to have the python enable flag work

mormj: The wrap file had gotten messed up, but is fixed now - @wirew0rm if you don't mind kicking the CI run again

RalphSteinhagen: Kicked the CI. @mormj will add you to the organisation/repo if you do not mimd. FAIR enough?

PR #55, opened 2023-03-06
Size optimizations, code simplifications

mattkretz:
A collection of commits that seem to work alright.

Please note, that the new bm_buffer benchmark works for me as defined. However, if you increase the number of samples to process to be larger than the buffer size, then the benchmarks with more than one writer hang. Is the `circular_buffer<int32_t, std::dynamic_extent, ProducerType::Multi>` usage correct for multiple writers?

### DIFF
--- a/bench/bm_buffer.cpp
+++ b/bench/bm_buffer.cpp
@@ -1,0 +1,157 @@
+#include "benchmark.hpp"
+
+#include <barrier>
+#include <boost/ut.hpp>
+#include <chrono>
+#include <iostream>
+#include <thread>
+
+#include <fmt/format.h>
+
+#include <buffer.hpp> // new buffer header interface
+#include <buffer_skeleton.hpp>
+#include <circular_buffer.hpp>
+#include <utils.hpp>
+
+#include <iostream>
+
+using namespace gr;
+
+#if defined __has_include && not __EMSCRIPTEN__
+#if __has_include(<pthread.h>) && __has_include(<sched.h>)
+#include <errno.h>
+#include <pthread.h>
+#include <sched.h>
+
+void
+setCpuAffinity(const int cpuID) // N.B. pthread is not portable
+{
+    const auto nCPU = std::thread::hardware_concurrency();
+    // fmt::print("set CPU affinity to core {}\n", cpuID % nCPU);
+    cpu_set_t cpuset;
+    CPU_ZERO(&cpuset);
+    CPU_SET(cpuID % nCPU, &cpuset);
+    int rc = pthread_setaffinity_np(pthread_self(), sizeof(cpu_set_t), &cpuset);
+    if (rc != 0) {
+        constexpr auto fmt = "pthread_setaffinity_np({} of {}): {} - {}\n";
+        fmt::print(stderr, fmt, cpuID, nCPU, rc, strerror(rc));
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(20)); // to force re-scheduling
+}
+#else
+void
+setCpuAffinity(const int cpuID) {}
+#endif
+#else
+void
+setCpuAffinity(const int cpuID) {}
+#endif
+
+void
+testNewAPI(Buffer auto &buffer, const std::size_t vector_length, const std::size_t min_samples, const int nProducer,
+           const int nConsumer, const std::string_view name) {
+    fair::meta::precondition(nProducer > 0);
+    fair::meta::precondition(nConsumer > 0);
+
+    constexpr int n_repeat = 8;
+
+    std::barrier barrier(nProducer + nConsumer + 1);
+
+    // init producers
+    std::atomic<int> threadID = 0;
+    std::vector<std::thread> producers;
+    for (int i = 0; i < nProducer; i++) {
+        producers.emplace_back([&]() {
+            // set thread affinity
+            setCpuAffinity(threadID++);
+            barrier.arrive_and_wait();
+            for (int rep = 0; rep < n_repeat; ++rep) {
+                BufferWriter auto writer = buffer.new_writer();
+                std::size_t nSamplesProduced = 0;
+                barrier.arrive_and_wait();
+                while (nSamplesProduced <= (min_samples + nProducer - 1) / nProducer) {
+                    writer.publish([](auto &) {}, vector_length);
+                    nSamplesProduced += vector_length;
+                }
+                barrier.arrive_and_wait();
+            }
+        });
+    }
+
+    // init consumers
+    std::vector<std::thread> consumers;
+    for (int i = 0; i < nConsumer; i++) {
+        consumers.emplace_back([&]() {
+            setCpuAffinity(threadID++);
+            barrier.arrive_and_wait();
+            for (int rep = 0; rep < n_repeat; ++rep) {
+                BufferReader auto reader = buffer.new_reader();
+                std::size_t nSamplesConsumed = 0;
+                barrier.arrive_and_wait();
+                while (nSamplesConsumed < min_samples) {
+                    if (reader.available() < vector_length) {
+                        continue;
+                    }
+                    const auto &input = reader.get(vector_length);
+                    nSamplesConsumed += input.size();
+
+                    if (!reader.consume(input.size())) {
+                        throw std::runtime_error(fmt::format("could not consume {} samples", input.size()));
+                    }
+                }
+                barrier.arrive_and_wait();
+            }
+        });
+    }
+
+    // all producers and consumer are ready, waiting to give the sign
+    barrier.arrive_and_wait();
+    ::benchmark::benchmark<n_repeat>(fmt::format("{:>8}: {} producers -<{:^4}>-> {} consumers", name, nProducer,
+                                                 vector_length, nConsumer),
+                                     min_samples)
+            = [&]() {
+                  barrier.arrive_and_wait();
+                  barrier.arrive_and_wait();
+              };
+
+    // clean up
+    for (std::thread &thread : producers) thread.join();
+    for (std::thread &thread : consumers) thread.join();
+}
+
+inline const boost::ut::suite _buffer_tests = [] {
+    const uint64_t samples = 10'000'000; // minimum number of samples
+    enum class BufferStrategy
+    {
+      posix,
+      portable
+    };
+
+    for (BufferStrategy strategy : { BufferStrategy::posix, BufferStrategy::portable }) {
+        for (int veclen : {1, 1024}) {
+            for (int nP = 1; nP <= 4; nP *= 2) {
+                for (int nR = 1; nR <= 4; nR *= 2) {
+                    const std::size_t size = std::max(4096, veclen) * nR * 10;
+                    auto allocator = std::pmr::polymorphic_allocator<int32_t>();
+                    const bool is_posix = strategy == BufferStrategy::posix;
+                    auto invoke = [&](auto buffer) {
+                        testNewAPI(buffer, veclen, samples, nP, nR, is_posix ? "POSIX" : "portable");
+                    };
+                    if (nP == 1) {
+                        using BufferType = circular_buffer<int32_t, std::dynamic_extent, ProducerType::Single>;
+                        Buffer auto buffer = is_posix ? BufferType(size) : BufferType(size, allocator);
+                        invoke(buffer);
+                    } else {
+                        using BufferType = circular_buffer<int32_t, std::dynamic_extent, ProducerType::Multi>;
+                        Buffer auto buffer = is_posix ? BufferType(size) : BufferType(size, allocator);
+                        invoke(buffer);
+                    }
+                }
+            }
+        }
+    }
+};
+
+int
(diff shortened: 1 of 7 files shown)


mattkretz:
Adding `-pthread` to the compiler and linker flags didn't help emscripten. Any idea how to make it work?

bench/bm_buffer.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include <buffer.hpp> // new buffer header interface
+#include <buffer_skeleton.hpp>
+#include <circular_buffer.hpp>
+
+#include <iostream>
+
+using namespace gr;
+
+#if defined __has_include
+#if __has_include(<pthread.h>) && __has_include(<sched.h>)
```
RalphSteinhagen, line 20: maybe we need to include a guard here for WASM/emscripten since the `pthread_setaffinity_np(..)` and other affinity tools are not supported on the web-browsers. See for example [here](https://www.reddit.com/r/WebAssembly/comments/em18dk/emscripten_pthread_undefined_symbol_and_thread/#t1_fdmq5gr) and [here](https://github.com/emscripten-core/emscripten/issues/7365).

bench/bm_buffer.cpp
```diff
@@ -0,0 +1,153 @@
...
+    for (std::thread &thread : consumers) thread.join();
+}
+
+inline const boost::ut::suite _buffer_tests = [] {
+    const uint64_t samples = 100'000'000; // minimum number of samples
+
+    for (int nP = 1; nP <= 4; nP *= 2) {
+        for (int nR = 1; nR <= 4; nR *= 2) {
+            if (not (nP == 1 and nR == 1)) {
+                benchmark::results::add_separator();
```
RalphSteinhagen, line 127: I actually like the separator concept... we have to see that we get this benchmark upstream to the UT testing framework where the this separator is an reading aid in addition to the definition of suites... well done. :+1: 

image

bench/bm_buffer.cpp
```diff
@@ -0,0 +1,153 @@
...
+            }
+            for (int buffer_type = 4; buffer_type <= 5; ++buffer_type) {
+                for (int veclen = 16; veclen <= 4096; veclen *= 16) { // vector length (<-> size of bulk operation)
+                    const std::size_t size = std::max(4096, veclen) * nR * 10;
+                    auto allocator = std::pmr::polymorphic_allocator<int32_t>();
+                    auto invoke = [&](auto buffer) {
+                        testNewAPI(buffer, veclen, samples, nP, nR, buffer_type == 4 ? "POSIX" : "portable");
```
RalphSteinhagen, line 134: I like that both cases are benchmarked against each other the POSIX/mmap double-mapped buffer has been used in GR for ages/historic reason but the benchmark pretty much shows that the results are either equivalent or the portable/memcpy-based version even faster in some cases. :+1:

bench/bm_buffer.cpp
```diff
@@ -0,0 +1,153 @@
...
+inline const boost::ut::suite _buffer_tests = [] {
+    const uint64_t samples = 100'000'000; // minimum number of samples
+
+    for (int nP = 1; nP <= 4; nP *= 2) {
+        for (int nR = 1; nR <= 4; nR *= 2) {
+            if (not (nP == 1 and nR == 1)) {
+                benchmark::results::add_separator();
+            }
+            for (int buffer_type = 4; buffer_type <= 5; ++buffer_type) {
+                for (int veclen = 16; veclen <= 4096; veclen *= 16) { // vector length (<-> size of bulk operation)
```
RalphSteinhagen, line 130: I think we should limit/reduce the test to just one veclen (e.g. '1' and or '1024') since the test primarily only pushes the writer/reader atomic lock-guards to the limit and does not do real processing of the data. This is also seen by the fact that the results pretty much scale with factors of 16.

RalphSteinhagen: Maybe we should also add separators only when the number of writers change? :thinking: 

Would it make sense to align the test case names so that the 'writers' and 'readers' in the text align vertically? This would imply adding some fixed number formatting width and/or spaces.

RalphSteinhagen (changes requested):
@mattkretz thanks a lot for the many improvements, thorough check of the circular buffer infrastructure, and clean-ups! :+1:  
This is really useful. 

Hope this isn't two nit-picking, may I ask you to only to do the following changes:
 * add the guard for WASM/Emscripten to not attempt to set the thread affinity -- this is not available on browsers and it would be nice to have this PR also compile for WASM... 
 * reduce the buffer benchmark to only one vector size ... see the comment above. This would make the test more readable.
 
 This is more a question than a  suggestion: I saw that the commits are well split into different topics. At the same time one cannot bisect and intermediate commits do not compile (the end result does). Normally it is useful that each commit compiles to see at which one a potential bug or performance regression was introduced. Since this is not possible with this PR, would you mind if I squash the commits when merging? Otherwise, we would need a before/after at least for the buffer benchmark.
 
 Still, the larger part of this PR is excellent! Well done! :+1:

bench/bm_buffer.cpp
mattkretz: Different `veclen` basically reduces the number of calls until wrap-around in the buffer happens and when the writer might have to block because of slow readers. But in this artificial benchmark I agree that it's unlikely to make a difference.
Before I added the sweep over `veclen` I added separators only for different number of writers, yes. The separator is first and foremost a reading aid. So grouping ~3-7 lines was my goal.

bench/bm_buffer.cpp
mattkretz: Hmm, for `veclen == 16` the "POSIX" vs. "portable" results are the same order of magnitude, but for `veclen == 4096` there is a factor 10-20 throughput difference (your screenshot above, but that's also what I saw here).

bench/bm_buffer.cpp
mattkretz: Umm, let me take that back. `veclen` does make a difference, as noted above.

mattkretz:
>     * add the guard for WASM/Emscripten to not attempt to set the thread affinity -- this is not available on browsers and it would be nice to have this PR also compile for WASM...

:+1: found the necessary preprocessor condition and ready to push.

>     * reduce the buffer benchmark to only one vector size ... see the comment above. This would make the test more readable.

Which one? Since they're not equivalent...

> This is more a question than a suggestion: I saw that the commits are well split into different topics. At the same time one cannot bisect and intermediate commits do not compile (the end result does).

That would be a bug. I'll check and will fix.

> Normally it is useful that each commit compiles to see at which one a potential bug or performance regression was introduced. Since this is not possible with this PR, would you mind if I squash the commits when merging?

I'd rather not squash. That would lose potentially useful information on `git blame`.

bench/bm_buffer.cpp
RalphSteinhagen: Maybe an option is to change/reshuffle the order of the loops from

now:  'for 1 ... nProducer' -> 'for 1 ... nReader' -> 'for buffer_type' -> 'for veclen'

to:   'for buffer_type'  -> 'for veclen' -> 'for 1 ... nProducer' -> 'for 1 ... nReader'
and 
  * make the buffer_type not a number (<-> historic reasons) but an enum
  * make the veclen loop a loop over a sorted list starting with [1, 1024, ...]
the latter could be -- alongside 'nProducer' and 'nReader' -- eventually be defined by an 'argv[]' statement (not now, but one of the future PR, once we know how we want to handle cmd-line parameters).
What do you think?

RalphSteinhagen: N.B. 1, 1024, ...  would be complemented by power-of-two. The intermediate sizes < 1024 do not have real domain use cases but the '1' is sort of a stress test pushing the race condition for the atomics the most. 

I'd recommend keeping vector length sizes above 1024 and larger writer/reader combinations optional because these become quickly opinionated, depend on the available HW (the initial assumption was that most have PCs with at least 8 cores/threads), and use case (e.g. latency-critical application favour veclen<1024, throughput-related application usually veclen>>1024).

RalphSteinhagen:
> That would be a bug. I'll check and will fix.


> I'd rather not squash. That would lose potentially useful information on `git blame`.

I am OK with not squashing but didn't want to impose that you have to 'check and fix'. 

> Which one? Since they're not equivalent...

As highlighted in the idea above, maybe reshuffling the loops and limiting to veclen = [1, 1024] would IMO be enough.

bench/bm_buffer.cpp
mattkretz: Hmm, also interesting could be different `veclen` for producer / consumer. In one case the consumer is faster and must often wait for the producer, in the other case the producer hits the last consumer and must block.
I'll reorder the loops in any case.

mattkretz:
> As highlighted in the idea above, maybe reshuffling the loops and limiting to veclen = [1, 1024] would IMO be enough.

I recommend against `veclen = 1`: Instead of 1-2 Minutes, the benchmark then runs for ~30 Minutes. I'm testing with 8 now, but I suspect it's a very noticeable slowdown compared to 16.

RalphSteinhagen:
> I recommend against `veclen = 1`: Instead of 1-2 Minutes, the benchmark then runs for ~30 Minutes.

I think that reducing the number of samples to `10'000'000` should be sufficient. This alone brought the test time down to around 2 minutes:

image

Also, having 100 million samples and skipping 1024 in one batch w/o real processing essentially test the RAM/cache memory bandwidth as a ratio of invalidating the caches w.r.t. atomics. I think the primary purpose of this benchmark is to assess the overhead and potential regressions to the CAS loops for the N-Writer-M-Reader scenario. I added a simple check ('QA'  column) looking at the required digits for standard deviation vs. the mean duration and the lowered number of samples does not reduce the significance of the test.

I got an idea while tinkering with your branch regarding indicating the quality of the test using coloured encodings of the 'total time' and 'ops/s' according to the 'stddev/mean' ratio and requested significant digits. E.g. assuming significant digits `Ns`:
  * `sigma/mu <= 10% / std::max(1, Ns)` -> green (statistically significant)
  * `10% / std::max(1, Ns) < sigma/mu < 100%` -> yellow (poor statistics)
  * `sigma/mu > 100%`  -> red (bad). 

In addition, one could add additional iterations in case the stddev compared to mean is bad. This is done also in other testing/bm frameworks but I dropped this for simplicity because there are a lot of corner cases (e.g. forced abort for too long running tests) and additional API required. But that'll be maybe for later. 

I played a bit and also re-ordered the test loop to:
```cpp
// ...
const uint64_t samples = 10'000'000; // minimum number of samples

bool first_run = true;
for (auto veclen: {1, 1024}) { // vector length (<-> size of bulk operation)
    for (int buffer_type = 4; buffer_type <= 5; ++buffer_type) {
        if (first_run) {
            first_run = false;
        } else {
            benchmark::results::add_separator();
        }
        for (int nP = 1; nP <= 4; nP *= 2) {
            for (int nR = 1; nR <= 4; nR *= 2) {
                const std::size_t size = veclen * nR * 10;
                auto allocator = std::pmr::polymorphic_allocator<int32_t>();
                auto invoke = [&](auto buffer) {
                    testNewAPI(buffer, veclen, samples, nP, nR, buffer_type == 4 ? "POSIX   " : "portable");
                };
                if (nP == 1) {
                    using BufferType = circular_buffer<int32_t, std::dynamic_extent, ProducerType::Single>;
                    Buffer auto buffer = buffer_type == 4 ? BufferType(size) : BufferType(size, allocator);
                    invoke(buffer);
                } else {
                    using BufferType = circular_buffer<int32_t, std::dynamic_extent, ProducerType::Multi>;
                    Buffer auto buffer = buffer_type == 4 ? BufferType(size) : BufferType(size, allocator);
                    invoke(buffer);
                }
            }
        }
    }
}
// ...
```

In the long run, it might be useful to only run one scenario, for example, the most efficient/relevant use case, and to make the others optional using some cmd-line parameters. The primary intention of these benchmarks is to get the order of magnitude for the performance and to test regressions i.e. may eventually become part of the CI/CD process to test not only the correctness of the code (<-> regular unit-tests).

bench/bm_buffer.cpp
RalphSteinhagen: There are probably many more variations. I'd propose keeping this KISS and maybe adding benchmark tests for other corner tests later?

bench/bm_buffer.cpp
RalphSteinhagen: I would not interpret too much into my screenshot which was done in a hurry. Anything beyond 10 GS/s is anyway more L1/L2/L3 cache related and can become very quickly flaky on what the OS/processor is doing in parallel, temperature, pressure, cosmic background radiation (esp. for the 100 GS/s results), ...

bench/bm_buffer.cpp
mattkretz: image
Fine. I'll push this state in a different branch then.

RalphSteinhagen (approved):
Excellent PR! Thanks for incorporating the changes! :+1: 

Good to be merged from my pov unless you want to further add to this.

PR #56, opened 2023-03-15
RAII circular buffer writer split get-publish

RalphSteinhagen:
This addresses two things:
  * fix circular_buffer writer split get-partial-publish ... relevant only for single-publisher policy and in case more samples have been allocated than are being published + minor clean-ups sonarlint rightfully complained about
  * changed circular buffer writer split  get-publish interface to using RAII 
    * single- and multi-producer scenario:
       throws a terminal exception in case no data has been published <-> API/programming bug
    * multi-producer scenario:
       throws a terminal exception in case the amount of data reserved and published isn't identical <-> API/programming bug

Unfortunately, I couldn't make this a compile-time expression that checks for that since 'N claimed' is information that is only available during runtime.

Eventually, I'd like to retire the unsafe `buffer_writer::get(...)` in favour of the RAII API. Feedback would be welcome...

### DIFF
--- a/bench/bm_buffer.cpp
+++ b/bench/bm_buffer.cpp
@@ -48,8 +48,8 @@ setCpuAffinity(const int cpuID) {}
 #endif
 
 enum class WriteApi {
-    via_lambda,
-    via_split_get_publish, via_split_get_publish_RAII };
+    via_lambda, via_split_request_publish_RAII };
+
 
 template<WriteApi PublisherAPI = WriteApi::via_lambda, typename T = int>
 void testNewAPI(Buffer auto &buffer, const std::size_t vector_length, const std::size_t min_samples, const int nProducer,
@@ -76,11 +76,7 @@ void testNewAPI(Buffer auto &buffer, const std::size_t vector_length, const std:
                 while (nSamplesProduced <= (min_samples + nProducer - 1) / nProducer) {
                     if constexpr (PublisherAPI == WriteApi::via_lambda) {
                         writer.publish([](auto &) {}, vector_length);
-                    } else if constexpr (PublisherAPI == WriteApi::via_split_get_publish) {
-                        auto [data, token] = writer.get(vector_length);
-
-                        writer.publish(token, vector_length);
-                    }  else if constexpr (PublisherAPI == WriteApi::via_split_get_publish_RAII) {
+                    } else if constexpr (PublisherAPI == WriteApi::via_split_request_publish_RAII) {
                         auto data = writer.reserve_output_range(vector_length);
 
                         data.publish(vector_length);
@@ -143,7 +139,7 @@ inline const boost::ut::suite _buffer_tests = [] {
       portable
     };
 
-    for (WriteApi writerAPI : { WriteApi::via_lambda,  WriteApi::via_split_get_publish, WriteApi::via_split_get_publish_RAII }) {
+    for (WriteApi writerAPI : { WriteApi::via_lambda,  WriteApi::via_split_request_publish_RAII }) {
         for (BufferStrategy strategy : { /*BufferStrategy::posix,*/ BufferStrategy::portable }) {
             for (int veclen : { 1, 1024 }) {
                 if (not(strategy == BufferStrategy::posix and veclen == 1)) {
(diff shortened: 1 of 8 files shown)


wirew0rm:
image

include/circular_buffer.hpp
```diff
@@ -246,17 +246,72 @@ class circular_buffer
...
+        std::swap(_offset, other._offset);
+        std::swap(_published_data, other._published_data);
+        other._n_slots_to_claim = 0;
+        other._published_data = true;
+        return *this;
+    };
+    ~WriterToken() {
+        if constexpr (std::is_base_of_v<MultiThreadedStrategy<SIZE, WAIT_STRATEGY>, ClaimType>) {
+            if (_n_slots_to_claim) {
+                throw std::runtime_error(fmt::format("circular_buffer::multiple_writer::WriterToken() - did not publish {} samples", _n_slots_to_claim));
```
ivan-cukic, line 277: Throwing in a destructor is not a good idea

include/circular_buffer.hpp
```diff
@@ -444,9 +509,9 @@ class circular_buffer
 
     // implementation specific interface -- not part of public Buffer / production-code API
     [[nodiscard]] auto n_readers()       { return _shared_buffer_ptr->_read_indices->size(); }
-    [[nodiscard]] auto claim_strategy()  { return _shared_buffer_ptr->_claim_strategy; }
-    [[nodiscard]] auto wait_strategy()   { return _shared_buffer_ptr->_wait_strategy; }
-    [[nodiscard]] auto cursor_sequence() { return _shared_buffer_ptr->_cursor; }
+    [[nodiscard]] auto &claim_strategy()  { return _shared_buffer_ptr->_claim_strategy; }
```
ivan-cukic, line 512: `const auto&`?

include/circular_buffer.hpp
```diff
@@ -246,17 +246,72 @@ class circular_buffer
         size_type                   _size;
         ClaimType*                  _claim_strategy;
 
+    class WriterToken {
+        buffer_writer<U>& _parent;
```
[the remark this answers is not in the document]
RalphSteinhagen: So you mean I should copy the std::shared_ptr?

include/circular_buffer.hpp
RalphSteinhagen: So is not calling the publish function on this RAII method ... any other suggestions how to indicate this dramatic API usage failure? Derek suggested that he wouldn't mind if the house would burn down ... a bit too dramatic for me but we need something in between.

include/circular_buffer.hpp
RalphSteinhagen: :+1: done

include/circular_buffer.hpp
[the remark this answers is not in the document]
ivan-cukic: Raw pointer is fine for non-owining pointers - and the redirections will work as they ought to

include/circular_buffer.hpp
ivan-cukic: API usage failure could be `std::terminate` IMO - exceptions can be handled and ignored, bad API usage should not be easy to ignore.

include/circular_buffer.hpp
```diff
@@ -277,6 +332,16 @@ class circular_buffer
             }
         }
 
+        [[nodiscard]] constexpr auto get_alt(std::size_t n_slots_to_claim) noexcept -> std::pair<std::span<U>, WriterToken> {
```
ivan-cukic, line 335: Maybe a more communicative name than `get_alt`?

include/circular_buffer.hpp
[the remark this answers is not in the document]
RalphSteinhagen: changed to raw pointer...

include/circular_buffer.hpp
RalphSteinhagen: yes, this function will eventually replace the 'get(..)' function.

bench/bm_buffer.cpp
```diff
@@ -70,7 +74,19 @@ testNewAPI(Buffer auto &buffer, const std::size_t vector_length, const std::size
...
+                    if constexpr (PublisherAPI == WriteApi::via_lambda) {
+                        writer.publish([](auto &) {}, vector_length);
+                    } else if constexpr (PublisherAPI == WriteApi::via_split_get_publish) {
+                        auto [data, token] = writer.get(vector_length);
+
+                        writer.publish(token, vector_length);
+                    }  else if constexpr (PublisherAPI == WriteApi::via_split_get_publish_RAII) {
+                        auto [data, token] = writer.get_alt(vector_length);
+
+                        token.publish(vector_length);
```
mattkretz, line 86: I'd prefer a more descriptive name than `get` here. You're getting something very *specific*; not like `std::get(tuple)` which could be *anything*. So the name should be specific. How would you write it in prose? "Request to publish `vector_length` samples via `writer`. Write samples. Finalize/commit the publication."?
I also don't feel too happy about the `std::pair` return type. However, if the function name suggests that it returns two objects, maybe that'd make me happier. `writer.tell_me_where_can_I_put_my_samples_and_how_can_I_let_you_know_when_I_am_done` :stuck_out_tongue_winking_eye: 
```c++
auto bucket = writer.request_publication_bucket(vector_length);
for (auto& x : bucket) {
  x = 1;
}
bucket.publish(vector_length);
```
Can't we instead ask for one thing that *is a* `span` or contains a `span` or simply `is a` `contiguous_range`. And at the same it is the RAII object that ensures the request isn't forgotten?

wirew0rm:
image

```
commit d15f5cb0fa99c320a4ea211994309f9a0b680f6f (HEAD -> RAII_split_buffer_writer, origin/RAII_split_buffer_writer)
Author: rstein <r.steinhagen@gsi.de>
Date:   Wed Mar 15 09:44:51 2023 +0100
```

RalphSteinhagen:
bm_case1 comparisons of old vs. new split buffer writer API by @wirew0rm:
before:
image

after:
image

RalphSteinhagen:
If there are no further objections, I'd merge these changes are approved by @ivan-cukic and @mattkretz. 

To my surprise, in terms of performance, these RAII changes are making things not slower but rather faster (~ few percent). :open_mouth:

mattkretz (approved):

PR #57, opened 2023-03-15
cmake: fix compiler warnings

wirew0rm:
set_project_warnings has to be called on an interface target which can then be used as a link library for the compilation target to enable/disable warnings instead of calling it on the target.


RalphSteinhagen (approved):
LGTM! :+1:

PR #59, opened 2023-03-17
DataSet aggregate type

wirew0rm:
Add an DataSet aggregate datatype. Gnuradio's pmtv library is added via fetchCMake for flexible, domain agnostic representation of metadata.


RalphSteinhagen (approved):
Thanks for the update!

PR #64, opened 2023-04-26
basic node setting handling

RalphSteinhagen:
Implemented node settings as public node member fields (<-> PoCo) that can be interacted with 
  * directly (default from C++ point-of-view), or via 
  * key-value-map (`std::map<std::string, pmtv::pmt, std::less<>>`), that can be interacted via 'setting().set(..)' and 'setting().get(..)'.

Example block implementation:
```cpp
template<typename T>
struct TestBlock : public node<TestBlock<T>> {
    IN<T>       in;
    OUT<T>      out;
    T           scaling_factor = static_cast<T>(1);
    std::string  context;
    std::int32_t n_samples_max = -1;
    float        sample_rate        = 1000.0f;

    template<fair::meta::t_or_simd<T> V>
    [[nodiscard]] constexpr V
    process_one(const V &a) const noexcept {
        return a * scaling_factor;
    }
};
ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T), (fair::graph::setting_test::TestBlock<T>), in, out, scaling_factor, context, n_samples_max, sample_rate);
```

The settings management is both thread-safe as well as safe w.r.t. real-time/non-real-time usages. Settings are changed either directly during the initial node initialisation, or in the 'work()' function ('staged' -> 'active' settings) prior to calling the user's 'process_[one, bulk](...)' functions.

For example:
```cpp
// during initialisation:
auto &src = flow_graph.make_node<Source<float>>({ { "sample_rate", 49000.0f } });
// or:
src.setting().set({ { "sample_rate", 49000.0f } });

// [..]
auto optCtxValue = block1.settings().get("context")
auto pmtKeyValueMap = block1.settings().get({ "key1", "key2", "key3" })
expect(eq(src.sample_rate, block1.sample_rate)) << "block1 matching src sample_rate";

// [..]
auto forward = block1.settings().apply_staged_parameters(); // explicitly apply/transfer staged pmt settings to struct field
// N.B. forward contains the elements that should be forwarded to dependent children
block1.settings().update_active_parameters(); // synchronises struct fields to pmt 'active' map
```
N.B. the `get(..)` interface returns the values as stored in the 'active' settings map which may differ w.r.t. the real struct field value. To synchronise the function `update_active_parameters()` would need to be called. An automatic synchronised was omitted for performance reasons (requires locking, cannot be done while the node is processing data, etc....)

The fields that are automatically updated/forwarded are controlled via the following two sets:
```cpp
auto auto_update_set  = block1.settings().auto_update_parameters(); // add/remove keys as necessary
auto auto_forward_set = block1.settings().auto_forward_parameters();  // add/remove keys as necessary
```

Setting a parameter explicitly via `set(..)`  (usually by the user) automatically removes the field from the `auto_update_parameters()` set and thus prevents unintential/automatic updates later on.

The `auto_forward_parameters()` set is set during block initialisation for canonical tags, i.e. those that are defined in the `fair::graph::tag::DEFAULT_TAGS` tuple (e.g. 'sample_rate', 'signal_name', 'signal_unit', ..., 'trigger_name', 'trigger_time', ..., 'context').

This implementation is pluggable (i.e. user can provide their own setting handling logic) and has API provisions for optional transaction and timing/ctx-based settings management that is to be completed with a later PR.

### DIFF
--- a/include/settings.hpp
+++ b/include/settings.hpp
@@ -1,0 +1,428 @@
+#ifndef GRAPH_PROTOTYPE_SETTINGS_HPP
+#define GRAPH_PROTOTYPE_SETTINGS_HPP
+
+#include <atomic>
+#include <chrono>
+#include <concepts>
+#include <mutex>
+#include <optional>
+#include <reflection.hpp>
+#include <set>
+#include <tag.hpp>
+#include <variant>
+
+namespace fair::graph {
+
+struct SettingsCtx {
+    // using TimePoint = std::chrono::time_point<std::chrono::utc_clock>; // TODO: change once the C++20 support is ubiquitous
+    using TimePoint               = std::chrono::time_point<std::chrono::system_clock>;
+    std::optional<TimePoint> time = std::nullopt; /// UTC time-stamp from which the setting is valid
+    std::string              context;             /// user-defined multiplexing context for which the setting is valid
+};
+
+template<typename T, typename Node>
+concept Settings = requires(T t, Node& n, std::span<const std::string> parameter_keys, const std::string &parameter_key, const tag_t::map_type &parameters, SettingsCtx ctx) {
+    { init(n, parameters) } -> std::same_as<void>;
+    /**
+     * @brief returns if there are stages settings that haven't been applied yet.
+     */
+    { t.changed() } -> std::same_as<bool>;
+
+    /**
+     * @brief stages new key-value pairs that shall replace the block field-based settings.
+     * N.B. settings become only active after executing 'apply_staged_parameters()' (usually done early on in the 'node::work()' function)
+     * @return key-value pairs that could not be set
+     */
+    { t.set(parameters, ctx) } -> std::same_as<tag_t::map_type>;
+    { t.set(parameters) } -> std::same_as<tag_t::map_type>;
+
+    /**
+     * @brief updates parameters based on node input tags for those with keys stored in `auto_update_parameters()`
+     * Parameter changes to down-stream nodes is controlled via `auto_forward_parameters()`
+     */
+    { t.auto_update(parameters, ctx) } -> std::same_as<void>;
+    { t.auto_update(parameters) } -> std::same_as<void>;
+
+    /**
+     * @brief return all available node settings as key-value pairs
+     */
+    { t.get() } -> std::same_as<tag_t::map_type>;
+
+    /**
+     * @brief return key-pmt values map for multiple keys
+     */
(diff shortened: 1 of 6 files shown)


include/settings.hpp
```diff
@@ -0,0 +1,428 @@
...
+                });
+            }
+        }
+    }
+
+    [[nodiscard]] const tag_t::map_type &
+    staged_parameters() const noexcept {
+        std::lock_guard lg(_lock);
+        return _staged;
+    }
```
wirew0rm, line 325: By returning the reference to `_staged` the access to it outlives the scope of the lock guard. It's a const ref, so the current thread cannot change the it, but another thread could call `set(...)` while the current thread is reading from the map.

include/settings.hpp
```diff
@@ -0,0 +1,428 @@
...
+#include <tag.hpp>
+#include <variant>
+
+namespace fair::graph {
+
+struct SettingsCtx {
+    // using TimePoint = std::chrono::time_point<std::chrono::utc_clock>; // TODO: change once the C++20 support is ubiquitous
+    using TimePoint               = std::chrono::time_point<std::chrono::system_clock>;
+    std::optional<TimePoint> time = std::nullopt; /// UTC time-stamp from which the setting is valid
+    std::string              context;             /// user-defined multiplexing context for which the setting is valid
```
wirew0rm, line 20: I think we had a discussion yesterday about changing this to `tag_t`, just adding this here as a note, since the basic_setting implementation does not use it this can be revisited when we tackle the context aware setting implementation.

include/settings.hpp
```diff
@@ -0,0 +1,428 @@
...
+};
+
+template<typename T, typename Node>
+concept Setting = requires(T t, Node& n, std::span<const std::string> parameter_keys, const std::string &parameter_key, const tag_t::map_type &parameters, SettingsCtx ctx) {
```
wirew0rm, line 24:
```suggestion
concept Settings = requires(T t, Node& n, std::span<const std::string> parameter_keys, const std::string &parameter_key, const tag_t::map_type &parameters, SettingsCtx ctx) {
```
Small naming nit: plural to reflect that this is/manages a set of settings and not just a single one?

wirew0rm (changes requested):
This looks nice, especially (but not limited to) the detailed unit-test and documentation.

I found what I think could become a concurrency problem and since this is IMO not fixable without changing the API and `Setting` concept I think it should be fixed before merging, everything else looks good to merge.

wirew0rm (approved):
:+1:

PR #65, opened 2023-05-09
Add simple schedulers

wirew0rm:
This PR adds functionality needed for running the work functions of a configured flowgraph.

- [x] add scheduler API, move existing functionality out of flowgraph
  - [x] scheduler API
  - [x] make graph structure accessible to scheduler
  - [x] implement different schedulers
    - [x] simple scheduler (was work function of flowgraph until now)
    - [x] breadth first scheduler
      - [x] fix block return status handling
      - [x] fix triggered buffer assertion
- [ ] schedulers are nodes, but until now they don't have any input/output ports, so they can only be top level schedulers and not nested


include/scheduler.hpp
```diff
@@ -0,0 +1,125 @@
...
+                } else if (result == work_return_t::INSUFFICIENT_OUTPUT_ITEMS) {
+                    something_happened = true;
+                }
+            }
+            run = something_happened;
+        }
+
+        return work_return_t::DONE;
+    }
+private:
```
RalphSteinhagen, line 60: member fields on top, followed by public constructor ... IMO you may skip the `private:` declarator since fields after the class header definition are always private.

include/scheduler.hpp
```diff
@@ -0,0 +1,125 @@
+#ifndef GRAPH_PROTOTYPE_SCHEDULER_HPP
+#define GRAPH_PROTOTYPE_SCHEDULER_HPP
+#include <graph.hpp>
+#include <set>
+#include <queue>
+
+namespace fair::graph::scheduler {
```
RalphSteinhagen, line 7: Like it makes it more readable further below:
`gr::scheduler::simple(...)`
vs.
`gr::scheduler::breadth_first(...)`

:+1:

include/scheduler.hpp
```diff
@@ -0,0 +1,125 @@
...
+                    }
+                }
+            }
+            if (!anything_happened) {
+                return work_return_t::DONE;
+            }
+        }
+    }
+
+private:
```
RalphSteinhagen, line 118: member fields on top, followed by public constructor ... IMO you may skip the private: declarator since fields after the class header definition are always private.

include/scheduler.hpp
```diff
@@ -0,0 +1,125 @@
...
+    init_proof init;
+    using node_t = fair::graph::graph::node_model*;
+    std::map<node_t, std::set<node_t>> adjacency_list{};
+    std::set<node_t> source_nodes{};
+};
+}
+#endif // GRAPH_PROTOTYPE_SCHEDULER_HPP
```
RalphSteinhagen, line 125: missing new-line

include/scheduler.hpp
```diff
@@ -0,0 +1,125 @@
...
+                }
+            }
+            run = something_happened;
+        }
+
+        return work_return_t::DONE;
+    }
+private:
+    init_proof init;
+    fair::graph::graph &graph;
```
RalphSteinhagen, line 62: graph should be moved to and scheduler be owner of the graph, i.e. `fair::graph::graph &` -> `fair::graph::graph`.

include/scheduler.hpp
```diff
@@ -0,0 +1,125 @@
...
+};
+
+/**
+ * Breadth first traversal scheduler which traverses the graph starting from the source nodes in a breath first fashion
+ * detecting cycles and nodes which can be reached from several source nodes.
+ */
+class breadth_first : public node<breadth_first> {
+public:
+    explicit breadth_first(fair::graph::graph &graph) : init{fair::graph::scheduler::init(graph)} {
```
RalphSteinhagen, line 71: graph should be moved to and scheduler be owner of the graph, i.e. `fair::graph::graph &` -> `fair::graph::graph`.

pass graph as r-value only

include/scheduler.hpp
```diff
@@ -0,0 +1,125 @@
...
+    graph._connection_definitions.clear();
+    return result;
+}
+
+/**
+ * Trivial loop based scheduler, which iterates over all nodes in definition order in the graph until no node did any processing
+ */
+class simple : public node<simple>{
+public:
+    explicit simple(fair::graph::graph &graph)  : init{fair::graph::scheduler::init(graph)}, graph(graph) { }
```
RalphSteinhagen, line 32: pass graph as r-value only

include/scheduler.hpp
```diff
@@ -0,0 +1,125 @@
...
+        }
+        while (true) {
+            bool anything_happened = false;
+            std::queue<fair::graph::graph::node_model *> queue{};
+            std::set<fair::graph::graph::node_model *>   reached;
+            // add all source nodes to queue
+            for (fair::graph::graph::node_model *source_node : source_nodes) {
+                queue.push(source_node);
+                reached.insert(source_node);
+            }
```
RalphSteinhagen, line 96: for performance reason (this is a hot spot) and since this is static information the order computation should be computed in the constructor and re-used.

include/scheduler.hpp
```diff
@@ -0,0 +1,125 @@
...
+            // process all nodes, adding all unvisited child nodes to the queue
+            while (!queue.empty()) {
+                fair::graph::graph::node_model *node = queue.front();
+                queue.pop();
```
RalphSteinhagen, line 100: push/pop is expensive -> should be done once or very sparingly during the scheduler lifetime.

include/tag.hpp
```diff
@@ -57,7 +57,7 @@ struct alignas(hardware_constructive_interference_size) tag_t {
         return map.at(key);
```
RalphSteinhagen, line 57: you could add `[[nodiscard]]' for this non-const-at as well (a bit nit-picky ...sorry).

test/qa_scheduler.cpp
```diff
@@ -0,0 +1,222 @@
...
+template<>
+struct fmt::formatter<std::pair<std::string, std::size_t>> {
+  template <typename ParseContext>
+  constexpr auto parse(ParseContext& ctx) {
+    return ctx.begin();
+  }
+
+  template <typename FormatContext>
+  auto format(const std::pair<std::string, std::size_t> & fp, FormatContext& ctx) {
+    return format_to(ctx.out(), "{} \t{}", fp.first, fp.second);
```
RalphSteinhagen, line 31: please do not use `\ŧ' aligned formatting since this can fail/align more or less randomly... better use, for example, `{:20} {:10}`.

test/qa_scheduler.cpp
```diff
@@ -0,0 +1,222 @@
...
+                {"s2", 34464},
+                {"add", 34464},
+                {"out", 34464}
+        });
+    };
+};
+
+int
+main() { /* tests are statically executed */
+}
```
RalphSteinhagen, line 222: newline missing.

test/qa_scheduler.cpp
```diff
@@ -0,0 +1,222 @@
...
+    adder(trace_vector &trace, std::string_view name) : tracer{trace} {this->_name = name;}
+    template<fair::meta::t_or_simd<T> V>
+    [[nodiscard]] constexpr auto
+    process_one(V a, V b) const noexcept {
+        trace(tracer, this->name(), vir::stdx::is_simd_v<V> ? V::size() : 1);
+        return a + b;
+    }
+};
+
+/**
```
RalphSteinhagen, line 125: very nice test!

please add two very simple tests verifying the execution order for the following graph topologies for both the simple and BFS based scheduler:

 a) source -> node 1 -> node 2 -> sink result: both scheduler should produce the order src->1->2->sink
 b) split paths:
     src -> node 1 -> node 2 -> sink 1
     src -> node 3 -> node 4 -> sink 2
     
      expected results:
      simple: src->1->2->sink1->3->4->sink2
      BFS: src->1->3->2->4->sink1->sink2
      
That should conclude the testing for these basic types of scheduler ... more advanced scheduling should be done in further PRs.

RalphSteinhagen (changes requested):
This looks generally quite good and complete for the outlined scope! Well done. :+1: 

There are only a few minor comments (see above).

Please add another new `bm_scheduler` benchmark for the two topologies outlined in the qa_scheduler and for both schedulers variants we have so far.

PR #66, opened 2023-05-11
CI: execute UT unit tests

wirew0rm:
Runs all registered UT tests. Uses an emscripten enabled ut fork for now (draft PR pending). The tests are run through ctest and automatically use nodejs to run the tests in the emscripten build.

Things that I had to change or work around:
- ut: use fork with `attribute(__constructor__)` patched out because emscripten does not support it for functions with arguments, see https://github.com/boost-ext/ut/pull/565 the problem with upstreaming is that there are a lot of other functions which we do not use that are not trivial to support on emscripten
- ut + clang16: fall back to basic reporter instead of `reporter_junit` because clang16 causes segfaults related to cout redirection
- tag.hpp: remove some constexpr-ness, which is not supported in emscripten
- use ubuntu's nodejs v18, as the (pretty old, v15) nodejs included in the emsdk was not able to run the unittests
- disable address sanitizer based on compiler id in cmake. This should be invesigated in a followup, the failures are different depending on the compiler.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,9 +86,9 @@ jobs:
         git clone https://github.com/emscripten-core/emsdk.git
         cd emsdk
         # Download and install the latest SDK tools.
-        ./emsdk install latest
+        ./emsdk install releases-03ecb526947f6a3702a0d083083799fe410d3893-64bit
         # Make the "latest" SDK "active" for the current user. (writes .emscripten file)
-        ./emsdk activate latest
+        ./emsdk activate releases-03ecb526947f6a3702a0d083083799fe410d3893-64bit
         # Activate PATH and other environment variables in the current terminal
         source ./emsdk_env.sh
 
@@ -117,6 +117,11 @@ jobs:
         cd ../build
         emmake make
 
+    - name: execute tests
+      working-directory: ${{runner.workspace}}/build
+      shell: bash
+      run: ctest --output-on-failure
+
     - name: execute binary
       if: matrix.configurations.compiler != 'emscripten'
       working-directory: ${{runner.workspace}}/build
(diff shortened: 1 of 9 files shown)


wirew0rm:
Regarding the asan problems, I did some digging around before giving up and deciding to disable them:

Some of the existing tests trigger asan's leak detection. qa_dynamic_port and qa_settings to be specific. Also occurs on older commits, so it's not related to a recent change.

I narrowed it down to occur as soon as a node is instantiated in the test. This is easy to overlook in the IDE because at least CLion shows the tests as green, but there is a second tab which lists the sanitizer issues and the return code is 1, which cmake interprets as a test failure.

I did a bit of googling on the messages which led to false positives where asan triggers if ubsan finds an issue because the ubsan reporting calls some demangling function which is (potentially?) allocating and returns a char* without any ownership tracking. Obviously that is not what's happening for us, but there are some similarities and it's the closest i could find.
My current working theory is, that our reflection logic also calls this non ownership tracking de-mangling API, and so we leak one char[] per node.

```
1234567==417365==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 128 byte(s) in 5 object(s) allocated from:
    #0 0x7fc13a9612e5 in __interceptor_realloc.part.0 (/nix/store/wvm2hvqdbbsp1f11463mrw8nyv678ipm-gcc-12.2.0-lib/lib/libasan.so.8+0xbb2e5)
    #1 0x7fc13a73740a in d_growable_string_callback_adapter (/nix/store/wvm2hvqdbbsp1f11463mrw8nyv678ipm-gcc-12.2.0-lib/lib/libstdc++.so.6+0xb740a)

SUMMARY: AddressSanitizer: 128 byte(s) leaked in 5 allocation(s).
```

the number of leaked objects corresponds to the number of nodes

gcc thread about the sanitier/demable issue https://gcc.gnu.org/bugzilla//show_bug.cgi?id=80578

PR #67, opened 2023-05-11
Thread pool

wirew0rm:
Ported from opencmw including porting of tests from catch2 to ut
Renamed to fit into the project conventions and enabled emscripten
compatibility.

TODO:
- [x] emscripten support


wirew0rm:
Added the test proposed in https://github.com/fair-acc/opencmw-cpp/issues/243 to the threadpool.

test/qa_thread_affinity.cpp
```diff
@@ -0,0 +1,161 @@
+#include <boost/ut.hpp>
```
RalphSteinhagen, line 1: the qa_thread_affinity unit-test should be added to the CI/CMakeList.txt.

RalphSteinhagen (approved):
OK, thanks for migrating this from opencmw-cpp and making it compatible with WASM! :+1:

The qa_thread_affinity unit-test should be added to the CI/CMakeList.txt. 

Otherwise OK!

test/qa_thread_affinity.cpp
wirew0rm: thanks for catching that, should be fixed.

Issue #68, opened 2023-05-17
non-power-2 type size support for circular_buffer<T> (notably std:vector<>)


RalphSteinhagen:
There is a missing feature/bug in the `circular_buffer<T>` requiring that the `sizeof(T)` is power of twos which is an issue w.r.t. `sizeof(std::vector<>) == 24` bytes:

```cpp
error: template constraint failure for ‘template<class T, long unsigned int SIZE, gr::ProducerType producer_type, class WAIT_STRATEGY>  requires (WaitStrategy<WAIT_STRATEGY>) && has_single_bit(long unsigned int)(sizeof (T)) class gr::circular_buffer’
  130 | class vmultiply_const : public fg::node<vmultiply_const<T>,
```

The power of two constraint was necessary as the minimum memory page size allocated by mmap (POSIX-buffer version) is usually at least 4096 bytes which doesn't align well with 24 bytes which is the size of `std::vector<>`. N.B. the circular buffer relies internally on twice the buffer size where the first half is mirrored/copied to the second half (either with `std::copy` or via the shm-system and mmap) which allows bulk read/write access across the wrapping point and is much faster than implementing a modulo-operator based circular buffer.

I was thinking of either modifying it to:

- provide a padded vector wrapper (size 32 byte) -- which might not be as portable (POSIX, 32 vs. 64 bit, page size might vary across systems, ...), or
- adding a least common multiple computation that increases the minimum buffer size N to  fulfill 'N >= N\_requested`and`N >= least\_common\_multiple(size(T), 4096)\` -- which may cause larger buffer than necessary/desired, i.e. 24 bytes align with multiples of 4096 bytes only for N = 12288.

[Here](https://compiler-explorer.com/z/6zx46Gq71) is the compiler-explorer snippet and specific function:

```cpp
uint64_t least_common_multiple(uint64_t a, uint64_t b) {
    return a * b / std::gcd(a, b);
}
```

RalphSteinhagen:
went for: https://compiler-explorer.com/z/e9nPzbnx9
```cpp
template <typename T>
constexpr std::size_t align_with_page_size(const std::size_t min_size) noexcept {
    const std::size_t pageSize = static_cast<std::size_t>(getpagesize());
    const std::size_t elementSize = sizeof(T);
    // least common multiple (lcm) of elementSize and pageSize
    std::size_t lcmValue = (elementSize * pageSize) / std::gcd(elementSize, pageSize);

[...]

PR #71, opened 2023-05-19
fixes regression in circular_buffer<T> for non-power-of-two 'T'

RalphSteinhagen:
This addresses issue #68.

Added corresponding unit test and new least common multiple that fits both the size of the type, min memory page size, and minimum requested buffer.

for future reference/playing around: https://compiler-explorer.com/z/e9nPzbnx9
```cpp
template <typename T>
constexpr std::size_t align_with_page_size(const std::size_t min_size) noexcept {
    const std::size_t pageSize = static_cast<std::size_t>(getpagesize());
    const std::size_t elementSize = sizeof(T);
    // least common multiple (lcm) of elementSize and pageSize
    std::size_t lcmValue = (elementSize * pageSize) / std::gcd(elementSize, pageSize);

    // adjust lcmValue to be larger than min_size
    while (lcmValue < min_size) {
        lcmValue += lcmValue;
    }
    return lcmValue;
}
```

### DIFF
--- a/include/circular_buffer.hpp
+++ b/include/circular_buffer.hpp
@@ -152,7 +152,6 @@ public:
     }
 
     template<typename T>
-    requires (std::has_single_bit(sizeof(T)))
     static inline std::pmr::polymorphic_allocator<T> allocator()
     {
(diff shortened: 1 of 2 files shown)


ivan-cukic (approved):
LGTM

PR #72, opened 2023-05-23
std::vector<T> settings- & pmt-map constructor initialiser support

RalphSteinhagen:
* extended settings support to also include vectors of arithmetics
 * extended default node constructor to support initialiser lists and pmt (should make the setup/Python integration more generic)
 * this eliminated the need for init(node<T>, pmt-map)
 * changed SettingsCtx context to pmt-map
 * fixed the optional init(old_map, new_map) interface
 
Some examples from the [qa_settings](https://github.com/fair-acc/graph-prototype/blob/789a5633913ee0604fc0e7ebb707d079ce9741b8/test/qa_settings.cpp#L286-L297) unit-tests:
https://github.com/fair-acc/graph-prototype/blob/789a5633913ee0604fc0e7ebb707d079ce9741b8/test/qa_settings.cpp#L104-L132
//[..]
https://github.com/fair-acc/graph-prototype/blob/789a5633913ee0604fc0e7ebb707d079ce9741b8/test/qa_settings.cpp#L258-L267
https://github.com/fair-acc/graph-prototype/blob/789a5633913ee0604fc0e7ebb707d079ce9741b8/test/qa_settings.cpp#L277-L284
https://github.com/fair-acc/graph-prototype/blob/789a5633913ee0604fc0e7ebb707d079ce9741b8/test/qa_settings.cpp#L286-L298
yields the following test output:
```
settings changed - update_count: 1
    key value changed: 'vector_setting` = [3, 2, 1] -> [42, 2, 3]
```

This is/was a prerequisite for the next PR regarding the `history_buffer<T>` and FIR/IIR support.

### DIFF
--- a/include/graph.hpp
+++ b/include/graph.hpp
@@ -1,11 +1,11 @@
 #ifndef GNURADIO_GRAPH_HPP
 #define GNURADIO_GRAPH_HPP
 
-#include "circular_buffer.hpp"
 #include "buffer.hpp"
-#include "typelist.hpp"
-#include "port.hpp"
+#include "circular_buffer.hpp"
 #include "node.hpp"
+#include "port.hpp"
+#include "typelist.hpp"
 
 #include <algorithm>
 #include <complex>
@@ -73,7 +73,8 @@ class dynamic_port {
                 = 0;
 
         [[nodiscard]] virtual connection_result_t
-        connect(dynamic_port &dst_port) = 0;
+        connect(dynamic_port &dst_port)
+                = 0;
 
         // internal runtime polymorphism access
         [[nodiscard]] virtual bool
@@ -172,8 +173,7 @@ class dynamic_port {
         connect(dynamic_port &dst_port) override {
             if constexpr (T::IS_OUTPUT) {
                 auto src_buffer = _value.writer_handler_internal();
-                return dst_port.update_reader_internal(src_buffer) ? connection_result_t::SUCCESS
-                                                                   : connection_result_t::FAILED;
+                return dst_port.update_reader_internal(src_buffer) ? connection_result_t::SUCCESS : connection_result_t::FAILED;
             } else {
                 assert(!"This works only on input ports");
                 return connection_result_t::FAILED;
@@ -335,7 +335,6 @@ public:
 
 #endif
 
-
 class graph {
 private:
     class node_model {
@@ -350,8 +349,7 @@ private:
         work() = 0;
 
         virtual void *
-        raw()
-                = 0;
+        raw() = 0;
     };
 
     template<typename T>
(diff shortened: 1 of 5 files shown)


include/graph.hpp
```diff
@@ -603,19 +603,20 @@ class graph {
...
     make_node(const tag_t::map_type& initial_settings) {
         static_assert(std::is_same_v<Node, std::remove_reference_t<Node>>);
-        auto& new_node_ref = _nodes.emplace_back(std::make_unique<node_wrapper<Node>>());
-        auto raw_ref = static_cast<Node*>(new_node_ref->raw());
-        if (!initial_settings.empty()) {
-            static_cast<node<Node>*>(raw_ref)->settings().init(*raw_ref, initial_settings);
-        }
+        auto                 &new_node_ref = _nodes.emplace_back(std::make_unique<node_wrapper<Node>>());
+        auto                  raw_ref      = static_cast<Node *>(new_node_ref->raw());
+        [[maybe_unused]] auto _1           = raw_ref->settings().set(initial_settings);
```
ivan-cukic, line 618: Maybe `std::ignore = ...` instead of` [[maybe_unused]] auto ...`?

include/node.hpp
```diff
@@ -229,12 +229,16 @@ class node : protected std::tuple<Arguments...> {
...
 public:
-    constexpr node() noexcept : node(self()) {}
+    node() noexcept : node({}) {}
 
-    constexpr node(Derived &derived) noexcept
+    node(std::initializer_list<std::pair<const std::string, pmtv::pmt>> init_parameter) noexcept
         : _tags_at_input(traits::node::input_port_types<Derived>::size())
         , _tags_at_output(traits::node::output_port_types<Derived>::size())
-        , _settings(std::make_unique<basic_settings<Derived>>(derived)) {}
+        , _settings(std::make_unique<basic_settings<Derived>>(*static_cast<Derived *>(this))) {
```
ivan-cukic, line 237: passing `this` which is a pointer to a non-constructed-yet object can lead to issues. I'd like to see a comment explaining why it is ok here, or something to be refactored so that this is not needed

include/graph.hpp
RalphSteinhagen: thanks @ivan-cukic for pointing out the use of `std::ignore` :+1:

include/node.hpp
RalphSteinhagen: safe delegated use of this (i.e. not used during construction)

The `this` is needed when accessing/transferring the staged and active settings to/from the struct which is and can only be done after construction (unfortunately).

ivan-cukic (approved):

PR #73, opened 2023-05-23
history_buffer<T> and basic FIR and IIR filter

RalphSteinhagen:
History buffer syntax:
```cpp
gr::history_buffer<int, 5> buffer;
buffer.push_back(1);  // buffer: [1] (size: 1, capacity: 5)
buffer.push_back(2);  // buffer: [1, 2] (size: 2, capacity: 5)
buffer.push_back(3);  // buffer: [1, 2, 3] (size: 3, capacity: 5)
buffer.push_back(4);  // buffer: [1, 2, 3, 4] (size: 4, capacity: 5)
buffer.push_back(5);  // buffer: [1, 2, 3, 4, 5] (size: 5, capacity: 5)
buffer.push_back(6);  // buffer: [2, 3, 4, 5, 6] (size: 5, capacity: 5)

 // indexing:
 buffer[0];     // value: 6 - unchecked access of last/actual sample
 buffer[-1];    // value: 5 - unchecked access of previous sample
// ...
buffer.at(0);  // value: 6 - checked access of last/actual sample
buffer.at(-1); // value: 5 - checked access of last/actual sample
// ...
buffer.get_span(0, 3);  // span: [4, 5, 6]
buffer.get_span(1, 3);  // span: [3, 4, 5]
buffer.get_span(0);     // span: [2, 3, 4, 5, 6]
buffer.get_span(1);     // span: [2, 3, 4, 5]
```

example use (here [FIR](https://github.com/fair-acc/graph-prototype/blob/a396bda164c7415da625fd48b073de87fc45a3d1/test/blocklib/core/filter/time_domain_filter.hpp#L12-L39
)):

```cpp
template<typename T> 
     requires std::floating_point<T> 
 struct fir_filter : node<fir_filter<T>> { 
     IN<T>             in; 
     OUT<T>            out; 
     std::vector<T>    b{}; // feedforward coefficients 
     history_buffer<T> inputHistory{ 32 }; 
  
     void 
     init(const tag_t::map_type & /*old_settings*/, const tag_t::map_type &new_settings) noexcept { 
         if (new_settings.contains("b") && b.size() >= inputHistory.capacity()) { 
             inputHistory = history_buffer<T>(std::bit_ceil(b.size())); 
         } 
     } 
  
     constexpr T 
     process_one(T input) noexcept { 
         inputHistory.push_back(input); 
         return std::inner_product(b.begin(), b.end(), inputHistory.rbegin(), static_cast<T>(0)); 
     } 
 }; 
 ```

### DIFF
--- a/bench/bm_history_buffer.cpp
+++ b/bench/bm_history_buffer.cpp
@@ -1,0 +1,150 @@
+#include "benchmark.hpp"
+
+#include <boost/ut.hpp>
+#include <chrono>
+#include <iostream>
+#include <thread>
+
+#include <fmt/format.h>
+
+#include <circular_buffer.hpp>
+#include <history_buffer.hpp>
+#include <utils.hpp>
+
+inline const boost::ut::suite _buffer_tests = [] {
+    constexpr uint32_t n_repetitions = 10;
+    constexpr uint32_t samples       = 10'000'000; // minimum number of samples
+    using namespace benchmark;
+    using namespace gr;
+
+    {
+        circular_buffer<int, std::dynamic_extent, ProducerType::Multi> buffer(32);
+        auto                                                           writer                   = buffer.new_writer();
+        auto                                                           reader                   = buffer.new_reader();
+
+        "circular_buffer<int>(32) - multiple producer"_benchmark.repeat<n_repetitions>(samples) = [&writer, &reader] {
+            static int counter = 0;
+            for (int i = 0; i < samples; i++) {
+                writer.publish([&](auto &vec) { vec[0] = counter; }, 1LU);
+                if (const auto data = reader.get(1)[0] != counter) {
+                    throw std::runtime_error(fmt::format("write {} read {} mismatch", counter, data));
+                }
+                [[maybe_unused]] const auto ret = reader.consume(1);
+                counter++;
+            }
+        };
+    }
+    {
+        circular_buffer<int> buffer(32);
+        auto                 writer                                                                      = buffer.new_writer();
+        auto                 reader                                                                      = buffer.new_reader();
+
+        "circular_buffer<int>(32) - single producer via lambda"_benchmark.repeat<n_repetitions>(samples) = [&writer, &reader] {
+            static int counter = 0;
+            for (int i = 0; i < samples; i++) {
+                writer.publish([&](auto &vec) { vec[0] = counter; }, 1LU);
+                if (const auto data = reader.get(1)[0] != counter) {
+                    throw std::runtime_error(fmt::format("write {} read {} mismatch", counter, data));
+                }
+                [[maybe_unused]] const auto ret = reader.consume(1);
+                counter++;
+            }
+        };
+    }
+    {
+        circular_buffer<int> buffer(32);
+        auto                 writer                                                                       = buffer.new_writer();
+        auto                 reader                                                                       = buffer.new_reader();
+
+        "circular_buffer<int>(32) - single producer via reserve"_benchmark.repeat<n_repetitions>(samples) = [&writer, &reader] {
+            static int counter = 0;
+            for (int i = 0; i < samples; i++) {
+                auto write_data = writer.reserve_output_range(1LU);
+                write_data[0]   = counter;
+                write_data.publish(1);
+                if (const auto data = reader.get(1)[0] != counter) {
+                    throw std::runtime_error(fmt::format("write {} read {} mismatch", counter, data));
+                }
+                [[maybe_unused]] const auto ret = reader.consume(1);
+                counter++;
+            }
+        };
+    }
+    /*
+     * left intentionally some space to improve the circular_buffer<T> implementation here
+     */
+    {
+        history_buffer<int> buffer(32);
+
+        "history_buffer<int>(32)"_benchmark.repeat<n_repetitions>(samples) = [&buffer] {
+            static int counter = 0;
+            for (int i = 0; i < samples; i++) {
+                buffer.push_back(counter);
+                if (const auto data = buffer[0] != counter) {
+                    throw std::runtime_error(fmt::format("write {} read {} mismatch", counter, data));
+                }
+                counter++;
+            }
+        };
+    }
+    {
+        history_buffer<int> buffer(32);
+
+        "history_buffer<int, 32>"_benchmark.repeat<n_repetitions>(samples) = [&buffer] {
+            static int counter = 0;
+            for (int i = 0; i < samples; i++) {
+                buffer.push_back(counter);
+                if (const auto data = buffer[0] != counter) {
+                    throw std::runtime_error(fmt::format("write {} read {} mismatch", counter, data));
+                }
+                counter++;
+            }
+        };
+    }
+
+    {
+        circular_buffer<int> buffer(32);
+        auto                 writer                                                     = buffer.new_writer();
+        auto                 reader                                                     = buffer.new_reader();
+
+        "circular_buffer<int>(32) - no checks"_benchmark.repeat<n_repetitions>(samples) = [&writer, &reader] {
+            static int counter    = 0;
+            auto       write_data = writer.reserve_output_range(1LU);
+            write_data[0]         = counter;
+            for (int i = 0; i < samples; i++) {
+                write_data.publish(1);
+                [[maybe_unused]] const auto data = reader.get(1)[0];
+                [[maybe_unused]] const auto ret  = reader.consume(1);
+                counter++;
+            }
+        };
+    }
+    {
+        history_buffer<int, 32> buffer;
+
+        "history_buffer<int, 32>  - no checks"_benchmark.repeat<n_repetitions>(samples) = [&buffer] {
+            static int counter = 0;
+            for (int i = 0; i < samples; i++) {
+                buffer.push_back(counter);
+                [[maybe_unused]] const auto data = buffer[0];
+                counter++;
+            }
+        };
+    }
+    {
+        history_buffer<int> buffer(32);
+
+        "history_buffer<int>(32)  - no checks"_benchmark.repeat<n_repetitions>(samples) = [&buffer] {
+            static int counter = 0;
+            for (int i = 0; i < samples; i++) {
+                buffer.push_back(counter);
+                [[maybe_unused]] const auto data = buffer[0];
+                counter++;
+            }
+        };
+    }
+};
+
+int
+main() { /* not needed by the UT framework */
+}
--- a/include/history_buffer.hpp
+++ b/include/history_buffer.hpp
@@ -1,0 +1,224 @@
+#ifndef GRAPH_PROTOTYPE_HISTORY_BUFFER_HPP
+#define GRAPH_PROTOTYPE_HISTORY_BUFFER_HPP
+
+#include <algorithm>
+#include <array>
+#include <cstddef>
+#include <memory>
+#include <span>
+#include <stdexcept>
+#include <type_traits>
+#include <vector>
+
+#include <fmt/format.h>
+
+namespace gr {
+
+/**
+ * @brief A simple circular history buffer that supports contiguous ranged access across the wrap-around point.
+ *
+ * This buffer is similar to the `circular_buffer<T>` and uses a double-mapped memory approach.
+ * It is optimised for single-threaded use and does not provide thread-safety or multi-writer/multi-reader support.
+ *
+ * Example usage:
+ * gr::history_buffer<int, 5> buffer;
+ * buffer.push_back(1);  // buffer: [1] (size: 1, capacity: 5)
+ * buffer.push_back(2);  // buffer: [1, 2] (size: 2, capacity: 5)
+ * buffer.push_back(3);  // buffer: [1, 2, 3] (size: 3, capacity: 5)
+ * buffer.push_back(4);  // buffer: [1, 2, 3, 4] (size: 4, capacity: 5)
+ * buffer.push_back(5);  // buffer: [1, 2, 3, 4, 5] (size: 5, capacity: 5)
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: bench/bm_history_buffer.cpp [hunk 1]
```diff
@@ -0,0 +1,150 @@
...
+    constexpr uint32_t samples       = 10'000'000; // minimum number of samples
+    using namespace benchmark;
+    using namespace gr;
+
+    {
+        circular_buffer<int, std::dynamic_extent, ProducerType::Multi> buffer(32);
+        auto                                                           writer                   = buffer.new_writer();
+        auto                                                           reader                   = buffer.new_reader();
+
+        "circular_buffer<int>(32) - multiple producer"_benchmark.repeat<n_repetitions>(samples) = [&writer, &reader] {
```
mormj, line 25: How is this test a multiple producer - should the example have multiple writers?

test/blocklib/core/filter/time_domain_filter.hpp
```diff
@@ -0,0 +1,114 @@
...
+    process_one(T input) noexcept {
+        inputHistory.push_back(input);
+        return std::inner_product(b.begin(), b.end(), inputHistory.rbegin(), static_cast<T>(0));
+    }
+};
+
+enum class IIRForm {
+    DF_I,  /// direct form I: preferred for fixed-point arithmetics (e.g. no overflow)
+    DF_II, /// direct form II: preferred for floating-point arithmetics (less operations)
+    DF_I_TRANSPOSED,
```
mormj, line 44: What are the transposed forms preferred for?

include/history_buffer.hpp
```diff
@@ -0,0 +1,223 @@
...
+
+#include <fmt/format.h>
+
+namespace gr {
+
+/**
+ * @brief A simple circular history buffer that supports contiguous ranged access across the wrap-around point.
+ *
+ * This buffer is similar to the `circular_buffer<T>` and uses a double-mapped memory approach.
+ * It is optimised for single-threaded use and does not provide thread-safety or multi-writer/multi-reader support.
```
mormj, line 21: Will there be a way to throw an exception if multiple readers are connected to a history buffer, or make this transparent to the user?

test/blocklib/core/filter/time_domain_filter.hpp
```diff
@@ -0,0 +1,114 @@
...
+    init(const tag_t::map_type & /*old_settings*/, const tag_t::map_type &new_settings) noexcept {
+        const auto new_size = std::max(a.size(), b.size());
+        if ((new_settings.contains("b") || new_settings.contains("a")) && (new_size >= inputHistory.capacity() || new_size >= inputHistory.capacity())) {
+            inputHistory  = history_buffer<T>(std::bit_ceil(new_size));
+            outputHistory = history_buffer<T>(std::bit_ceil(new_size));
+        }
+    }
+
+    [[nodiscard]] T
+    process_one(T input) noexcept {
```
mormj, line 75: these look good!

mormj (approved):
These look good.  At some point we will need to decide how to structure the non-block DSP routines as a "separate" library/folder.  Also, with the history buffer, it will be necessary to prevent invalid usage (multi-consumer), and might also be a good idea to document how this compares with GNU Radio set_history() - as that is a perpetually confusing topic

include/history_buffer.hpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    /**
+     * @brief Adds an element to the end expiring the oldest element beyond the buffer's capacities.
+     */
+    constexpr void
+    push_back(const T &value) noexcept {
+        _buffer[_write_position]             = value;
+        _buffer[_write_position + _capacity] = _buffer[_write_position];
+        _write_position                      = (_write_position + 1) % _capacity;
```
mattkretz, line 84: `% _capacity` can be rather expensive. If there's no good solution to use a bit-mask, a branch should be cheaper. The compiler is unlikely to figure out that it can assume `_write_position < _capacity` — though you could tell it:
```c++
if (_write_position >= _capacity) std::unreachable();
```
I expect the following to be cheaper:
```c++
++_write_position;
if (_write_position == _capacity) {
  _write_position = 0;
}
```

include/history_buffer.hpp
```diff
@@ -0,0 +1,223 @@
...
+        _size                                = std::min(_size + 1, _capacity);
```
mattkretz, line 85: This should get compiled into a branch not a `cmov`. So actually using an `if` might help the compiler here. (an `[[unlikely]]` on the branch probably wouldn't hurt either)

mattkretz (changes requested):
I just took a preliminary look at `history_buffer`. I like the approach, but I have to think a bit more about some of the details.

I believe there should be a unit test for `history_buffer`.

RalphSteinhagen:
> These look good. At some point we will need to decide how to structure the non-block DSP routines as a "separate" library/folder. Also, with the history buffer, it will be necessary to prevent invalid usage (multi-consumer), and might also be a good idea to document how this compares with GNU Radio set_history() - as that is a perpetually confusing topic

:+1: The placement of the blocks is temporary. Will be moved with the next project structure update ... at least we are in a shape right now that we can start populating the library with meaningful user-level blocks. There are just two missing feature remaining and then we are good to go:
  * code documentation -> export for use in UI
  * `node<T>` template `T` specialisation constraints, i.e. limiting the types `T` for which the implemented block is valid (if necessary at all):
  
  The `history_buffer` is single-producer/single-consumer and non-multi-threaded use. Do we need protection for these constraints? If yes, how?

mattkretz:
> The `history_buffer` is single-producer/single-consumer and non-multi-threaded use. Do we need protection for these constraints? If yes, how?

This is a constraint that you can't check at compile-time. Documentation is all you can do.

bench/bm_history_buffer.cpp
RalphSteinhagen: The multi-producer case is tested already in another unit-test and benchmark. Here it's just testing the multi-producer strategy which should -- in theory -- be a bit more computational intensive. Actually, it turns out -- maybe this is a simplified testing artefact -- that the circular_buffer is faster than the history_buffer. Need to investigate why/how the compiler can optimise one better than the other (the history_buffer is basically simpler and should be thus faster)... :thinking:

test/blocklib/core/filter/time_domain_filter.hpp
RalphSteinhagen: Similar to the I and II but inversed. I added them more for academic/thoroughness and to give @mattkretz alternative implementations to see which one is better from a SIMD point of view.

include/history_buffer.hpp
RalphSteinhagen: ... not sure. Need to investigate whether this is possible w/o paying a penalty for the nominal use-case.

RalphSteinhagen:
> I just took a preliminary look at `history_buffer`. I like the approach, but I have to think a bit more about some of the details.
> 
> I believe there should be a unit test for `history_buffer`.

There is a unit-test in qa_buffer that includes the history_buffer.

include/history_buffer.hpp
RalphSteinhagen: Before:
```
┌───────────────────────benchmark:────────────────────────┬──────┬─#N──┬─CTX-SW─┬───CPU cache misses────┬──CPU branch misses───┬─<CPU-I>─┬──min───┬──mean──┬─stddev─┬─median─┬──max───┬─total time─┬─ops/s─┐
│ circular_buffer<int>(32) - multiple producer            │ PASS │ 10  │    71  │ 1.75M / 5.00M = 35.0% │ 62.2k / 801M =  0.0% │   69.0  │ 261 ms │ 277 ms │  13 ms │ 276 ms │ 306 ms │       3  s │ 36.1M │
│ circular_buffer<int>(32) - single producer via lambda   │ PASS │ 10  │    16  │  618k / 1.45M = 42.6% │ 42.8k / 701M =  0.0% │   57.0  │ 190 ms │ 199 ms │   8 ms │ 197 ms │ 217 ms │       2  s │ 50.3M │
│ circular_buffer<int>(32) - single producer via reserve  │ PASS │ 10  │     3  │  408k / 1.14M = 35.6% │ 43.5k / 601M =  0.0% │   52.0  │ 177 ms │ 204 ms │  24 ms │ 202 ms │ 235 ms │       2  s │ 49.1M │
│ history_buffer<int>(32)                                 │ PASS │ 10  │    18  │   290k / 645k = 45.0% │ 17.5k / 200M =  0.0% │   20.0  │ 170 ms │ 175 ms │   2 ms │ 174 ms │ 180 ms │       2  s │ 57.2M │
│ history_buffer<int, 32>                                 │ PASS │ 10  │    49  │   337k / 764k = 44.1% │ 22.6k / 201M =  0.0% │   20.0  │ 181 ms │ 186 ms │   4 ms │ 184 ms │ 197 ms │       2  s │ 53.9M │
│ circular_buffer<int>(32) - no checks                    │ PASS │ 10  │     1  │  45.1k / 102k = 44.2% │ 2.75k / 300M =  0.0% │   17.0  │  23 ms │  24 ms │ 383 us │  24 ms │  24 ms │     236 ms │  423M │
│ history_buffer<int, 32>  - no checks                    │ PASS │ 10  │     3  │   109k / 407k = 26.9% │ 10.7k / 100M =  0.0% │   14.0  │ 104 ms │ 114 ms │   7 ms │ 115 ms │ 123 ms │       1  s │ 87.9M │
│ history_buffer<int>(32)  - no checks                    │ PASS │ 10  │     2  │  77.1k / 297k = 26.0% │ 9.15k / 100M =  0.0% │   14.0  │ 101 ms │ 103 ms │   1 ms │ 103 ms │ 105 ms │       1  s │ 97.2M │
└─────────────────────────────────────────────────────────┴──────┴─────┴────────┴───────────────────────┴──────────────────────┴─────────┴────────┴────────┴────────┴────────┴────────┴────────────┴───────┘
```
with the suggested changes:
```
┌───────────────────────benchmark:────────────────────────┬──────┬─#N──┬─CTX-SW─┬───CPU cache misses────┬──CPU branch misses───┬─<CPU-I>─┬──min───┬──mean──┬─stddev─┬─median─┬──max───┬─total time─┬─ops/s─┐
│ circular_buffer<int>(32) - multiple producer            │ PASS │ 10  │     2  │  564k / 2.00M = 28.3% │ 44.9k / 801M =  0.0% │   69.0  │ 225 ms │ 232 ms │   3 ms │ 234 ms │ 236 ms │       2  s │ 43.0M │
│ circular_buffer<int>(32) - single producer via lambda   │ PASS │ 10  │    28  │   332k / 979k = 33.9% │ 44.1k / 701M =  0.0% │   57.0  │ 147 ms │ 162 ms │   7 ms │ 166 ms │ 171 ms │       2  s │ 61.6M │
│ circular_buffer<int>(32) - single producer via reserve  │ PASS │ 10  │     6  │  32.2k / 230k = 14.0% │ 35.1k / 600M =  0.0% │   52.0  │ 136 ms │ 142 ms │   8 ms │ 138 ms │ 162 ms │       1  s │ 70.7M │
│ history_buffer<int>(32)                                 │ PASS │ 10  │     1  │ 11.0k / 75.6k = 14.5% │ 7.34k / 300M =  0.0% │   20.1  │  65 ms │  66 ms │ 412 us │  66 ms │  67 ms │     659 ms │  152M │
│ history_buffer<int, 32>                                 │ PASS │ 10  │     1  │  24.3k / 103k = 23.6% │ 9.61k / 300M =  0.0% │   20.1  │  66 ms │  67 ms │   2 ms │  66 ms │  71 ms │     673 ms │  149M │
│ circular_buffer<int>(32) - no checks                    │ PASS │ 10  │     1  │ 9.91k / 52.0k = 19.0% │ 2.81k / 300M =  0.0% │   17.0  │  16 ms │  18 ms │   3 ms │  17 ms │  25 ms │     184 ms │  543M │
│ history_buffer<int, 32>  - no checks                    │ PASS │ 10  │     0  │ 4.84k / 24.1k = 20.0% │ 17.7k / 200M =  0.0% │   14.0  │  11 ms │  11 ms │  56 us │  11 ms │  12 ms │     114 ms │  878M │
│ history_buffer<int>(32)  - no checks                    │ PASS │ 10  │     0  │  893  / 10.0k =  8.9% │  943  / 200M =  0.0% │   14.0  │  11 ms │  11 ms │   2 us │  11 ms │  11 ms │     114 ms │  880M │
└─────────────────────────────────────────────────────────┴──────┴─────┴────────┴───────────────────────┴──────────────────────┴─────────┴────────┴────────┴────────┴────────┴────────┴────────────┴───────┘

```
:+1: @mattkretz well spotted ... your review is worth every penny. I owe you an :ice_cream: and :coffee: !

RalphSteinhagen:
@mattkretz incorporated your changes and needed to force-push to have them in the first (history_buffer) commit.

include/history_buffer.hpp
```diff
@@ -0,0 +1,224 @@
...
+
+    /**
+     * @brief Adds an element to the end expiring the oldest element beyond the buffer's capacities.
+     */
+    constexpr void
+    push_back(const T &value) noexcept {
+        _buffer[_write_position]             = value;
+        _buffer[_write_position + _capacity] = _buffer[_write_position];
+        ++_write_position;
+        [[unlikely]] if (_write_position == _capacity) { _write_position = 0; }
```
mattkretz, line 85: This compiles? `if (_write_position == _capacity) [[unlikely]] {` is the attribute placement I know.

mattkretz (approved):
You could still replace the `_size = std::min(_size + 1, _capacity);` with a branch... (unless you benchmarked/disassembled it to prove me wrong)

PR #78, opened 2023-06-02
scheduler: add thread pool backed parallel implementation to "simple" and "breadth_first" schedulers

### LINKED ISSUE #76: graph: multithreaded schedulers
Provide schedulers for graph-prototype that use the thread pool to call the work functions on multiple threads. Generic blocks should run on a fixed size threadpool, while blocking blocks have to be scheduled to a dedicated IO pool which dynamically grows to allow an arbitrary number of blocking nodes.

First implementation will assume that all tasks are non-blocking and will be scheduled on an `CPU_BOUND` type scheduler. The supported parallelized schedulers will be breadth-first and simple schedulers as a proof of concept.

Later on, the graph and scheduler APIs will be extended to support more complex schedulers.

optional: also port the existing scheduler from the gr4-dev branch to be able to compare performance

wirew0rm:
Adds basic multi-threaded implementations for the simple scheduling algorithms (simple & breadth-first).

fixes #76

### DIFF
--- a/bench/bm_case1.cpp
+++ b/bench/bm_case1.cpp
@@ -471,7 +471,7 @@ invoke_work(auto &sched) {
     using namespace benchmark;
     test::n_samples_produced = 0LU;
     test::n_samples_consumed = 0LU;
-    sched.work();
+    sched.run_and_wait();
     expect(eq(test::n_samples_produced, N_SAMPLES)) << "did not produce enough output samples";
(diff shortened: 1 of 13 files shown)


RalphSteinhagen (approved):
Thanks for doing the in-person review... was IMO more efficient since this wasn't just checking for errors but also designing and evolving the overall scheduler design.

This looks good from my point of view and is an excellent basis for further extensions! :+1:

PR #80, opened 2023-06-06
data_sink with preliminary streaming callback/polling

frankosterfeld:


RalphSteinhagen (approved):
@frankosterfeld looks good for what is being implemented. There are some simplification possible (see above) but the general direction looks good! :+1: 

I think the registry might be important as a next step since, for example, the UI won't have a direct reference on the data sink block.

include/data_sink.hpp
```diff
@@ -0,0 +1,155 @@
...
+        std::lock_guard lg(pending_listeners.mutex);
+        pending_listeners.list.push_back({
+            .mode = acquisition_mode::Continuous,
+            .buffer = gr::circular_buffer<T>(0),
+            .callback = std::move(callback)
+        });
+        pending_listeners.dirty = true;
+    }
+
+    [[nodiscard]] work_return_t work() {
```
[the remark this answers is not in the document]
frankosterfeld: > Rather than implementing the full `work()` function it might be easier to just implement the `process_bulk(std::span<const T> in) {...}` function. This way you can re-use a lot of the underlying tag and error reporting mechanics.

I looked at process_bulk, but my problem was that I want don't necessarily want to process all the available data, and I didn't see a way to report that.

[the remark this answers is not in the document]
wirew0rm: I added that usecase to the work function refactoring issue #81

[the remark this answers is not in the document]
frankosterfeld: Thinking about it, I wonder if there's a good reason not to process everythin, dropping/blocking more aggressively.

include/data_sink.hpp
```diff
@@ -72,6 +76,14 @@ class data_sink : public node<data_sink<T>> {
     std::vector<listener> listeners;
 
 public:
+    // TODO sink should register itself on construction, but name is set afterwards via
+    // set_name, which we have no hook into. Maybe the registration should be done by the
+    // graph creating/destroying the sink instead?
```
RalphSteinhagen, line 81: I added another PR https://github.com/fair-acc/graph-prototype/pull/84 that should provide each node with a unique name... maybe this helps?

include/node.hpp
```diff
@@ -330,13 +330,20 @@ class node : protected std::tuple<Arguments...> {
...
-
-                if (tag_stream_head_distance > 0 && availableSamples > tag_stream_head_distance) {
+                if (tag_stream_head_distance == 0) {
+                    tagsToProcess = 1;
+                    if (availableTags > 1) {
+                        tag_stream_head_distance = tagData[1].index - port.streamReader().position();
+                        assert(tag_stream_head_distance >= 0 && "negative tag vs. stream distance (2nd tag)");
+                    }
+                }
+                if (tag_stream_head_distance > 0 && availableSamples > static_cast<std::size_t>(tag_stream_head_distance)) {
```
frankosterfeld, line 346: Note that this code doesn't handle the case yet where there are more than one tag for a sample in the input stream (I think that can happen here, merging seems to happen later?).
Then it would have to get all available streams and iterate/increment tagsToProcess accordingly.

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+        return { std::string{ name }, {} };
+    }
+};
+
+class data_sink_registry {
+    std::mutex            _mutex;
+    std::vector<std::any> _sinks;
+
+public:
+    // TODO this shouldn't be a singleton but associated with the flow graph (?)
```
RalphSteinhagen, line 68: Yes, I concur. The registry should eventually be owned by the graph and injected to the nodes that need this. We may need to generalise this before doing so.... Maybe in the end simply iterating over a flattened graph-node hierarchy would be sufficient. :thinking: 

@wirew0rm @ivan-cukic what do you think?

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+                return std::any_cast<data_sink<T> *>(v) == sink;
+            } catch (...) {
+                return false;
+            }
+        });
+    }
+
+    template<typename T>
+    std::shared_ptr<typename data_sink<T>::poller>
+    get_streaming_poller(const data_sink_query &query, blocking_mode block = blocking_mode::NonBlocking) {
```
RalphSteinhagen, line 97: default should be 'blocking'

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+        std::lock_guard lg{ _mutex };
+        auto            sink = find_sink<T>(query);
+        return sink ? sink->get_streaming_poller(block) : nullptr;
+    }
+
+    template<typename T, TriggerPredicate P>
+    std::shared_ptr<typename data_sink<T>::dataset_poller>
+    get_trigger_poller(const data_sink_query &query, P p, std::size_t pre_samples, std::size_t post_samples, blocking_mode block = blocking_mode::NonBlocking) {
```
RalphSteinhagen, line 105: default should be 'blocking'

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+        std::lock_guard lg{ _mutex };
+        auto            sink = find_sink<T>(query);
+        return sink ? sink->get_trigger_poller(std::forward<P>(p), pre_samples, post_samples, block) : nullptr;
+    }
+
+    template<typename T, TriggerObserverFactory F>
+    std::shared_ptr<typename data_sink<T>::dataset_poller>
+    get_multiplexed_poller(const data_sink_query &query, F triggerObserverFactory, std::size_t maximum_window_size, blocking_mode block = blocking_mode::NonBlocking) {
```
RalphSteinhagen, line 113: default should be 'blocking'

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+        std::lock_guard lg{ _mutex };
+        auto            sink = find_sink<T>(query);
+        return sink ? sink->get_multiplexed_poller(std::forward<F>(triggerObserverFactory), maximum_window_size, block) : nullptr;
+    }
+
+    template<typename T, TriggerPredicate P>
+    std::shared_ptr<typename data_sink<T>::dataset_poller>
+    get_snapshot_poller(const data_sink_query &query, P p, std::chrono::nanoseconds delay, blocking_mode block = blocking_mode::NonBlocking) {
```
RalphSteinhagen, line 121: default should be 'blocking'

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+    { p(tag) } -> std::convertible_to<bool>;
+};
+
+template<typename T>
+concept TriggerObserver = requires(T o, tag_t tag) {
+    { o(tag) } -> std::convertible_to<trigger_observer_state>;
+};
+
+template<typename T>
+concept TriggerObserverFactory = requires(T f) {
```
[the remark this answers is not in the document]
RalphSteinhagen: Maybe std::invocable with a `TriggerObserver` would be easier w.r.t. C++ STL?

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+
+    template<typename T, TriggerPredicate P>
+    std::shared_ptr<typename data_sink<T>::dataset_poller>
+    get_snapshot_poller(const data_sink_query &query, P p, std::chrono::nanoseconds delay, blocking_mode block = blocking_mode::NonBlocking) {
+        std::lock_guard lg{ _mutex };
+        auto            sink = find_sink<T>(query);
+        return sink ? sink->get_snapshot_poller(std::forward<P>(p), delay, block) : nullptr;
+    }
+
+    template<typename T, typename Callback>
```
RalphSteinhagen, line 127: `Callback` should be a constraint -> here, a concept would be helpful for users to communicate which signature is needed.

Same for the occurrences below.

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+ * and either retrieve a poller handler that allows asynchronous data from a different thread,
+ * or register a callback that is invoked by the sink if the user-conditions are met.
+ *
+ * <pre>
+ * @code
+ *         ╔═══════════════╗
+ *    in0 ━╢   data sink   ║                      ┌──── caller ────┐
+ * (err0) ━╢ (opt. error)  ║                      │                │
+ *         ║               ║  retrieve poller or  │ (custom non-GR │
+ *        ━╢ :signal_name  ║←--------------------→│  user code...) │
```
RalphSteinhagen, line 240: tiny change: replace unused `╢` with `║` 

as discussed: we decided that each sink has only one input at a time. this simplifies the internal design and potential error handling mechanism in case input signals have incompatible sampling rates.

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+class data_sink : public node<data_sink<T>> {
+    struct abstract_listener;
+
+    static constexpr std::size_t                   _listener_buffer_size = 65536;
+    std::deque<std::unique_ptr<abstract_listener>> _listeners;
+    std::mutex                                     _listener_mutex;
+    gr::history_buffer<T>                          _history = gr::history_buffer<T>(1);
+
+public:
+    Annotated<float, "sample rate", Doc<"signal sample rate">, Unit<"Hz">>           sample_rate = 10000.f;
```
RalphSteinhagen, line 271: maybe a default of `1` might be better?!? This would correspond to a horizontal axis of 1 tick == 1 sample.
Similar for signal_unit which should maybe default to `a.u.` or `Ts`?

signal_min and signal_max should be at the min/max rails of the type `T`

signal_name default value should be either `unknown_siganal` or maybe through a static (atomic?) global counter `i` something like `unknown_signal#<i>`?

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+    Annotated<std::string, "signal name", Visible>                                   signal_name;
+    Annotated<std::string, "signal unit", Visible, Doc<"signal's physical SI unit">> signal_unit;
+    Annotated<T, "signal min", Doc<"signal physical min. (e.g. DAQ) limit">>         signal_min;
+    Annotated<T, "signal max", Doc<"signal physical max. (e.g. DAQ) limit">>         signal_max;
+
+    IN<T, std::dynamic_extent, _listener_buffer_size>                                in;
+
+    struct poller {
+        gr::circular_buffer<T>            buffer       = gr::circular_buffer<T>(_listener_buffer_size);
+        decltype(buffer.new_reader())     reader       = buffer.new_reader();
```
RalphSteinhagen, line 281: maybe re-use the existing infrastructure from `port<T>` we have there already definitions like:
 ```cpp
    using ReaderType    = decltype(std::declval<BufferType>().new_reader());
    using WriterType    = decltype(std::declval<BufferType>().new_writer());
    using IoType        = std::conditional_t<IS_INPUT, ReaderType, WriterType>;
    using TagReaderType = decltype(std::declval<TagBufferType>().new_reader());
    using TagWriterType = decltype(std::declval<TagBufferType>().new_writer());
    using TagIoType     = std::conditional_t<IS_INPUT, TagReaderType, TagWriterType>;

private:
    std::string  _name         = static_cast<std::string>(PortName);
    std::int16_t _priority     = 0; // → dependents of a higher-prio port should be scheduled first (Q: make this by order of ports?)
    std::size_t  _min_samples  = (MIN_SAMPLES == std::dynamic_extent ? 1 : MIN_SAMPLES);
    std::size_t  _max_samples  = MAX_SAMPLES;
    bool         _connected    = false;

    IoType       _ioHandler    = new_io_handler();
    TagIoType    _tagIoHandler = new_tag_io_handler();
 ```

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+    struct dataset_poller {
+        gr::circular_buffer<DataSet<T>> buffer     = gr::circular_buffer<DataSet<T>>(_listener_buffer_size);
+        decltype(buffer.new_reader())   reader     = buffer.new_reader();
+        decltype(buffer.new_writer())   writer     = buffer.new_writer();
+
+        std::atomic<bool>               finished   = false;
+        std::atomic<std::size_t>        drop_count = 0;
+
+        [[nodiscard]] bool
+        process_bulk(std::invocable<std::span<DataSet<T>>> auto fnc) {
```
RalphSteinhagen, line 328: In most other cases `process_bulk(...)` and `process_one(..)` are usually XOR implementations. Which should maybe focus on one. 

Since this will be initially (mostly) needed for UI purposes, maybe @giucam and/or @ivan-cukic could comment?
@ivan-cukic

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+            const auto to_write = std::min(in_data.size(), _history.capacity());
+            _history.push_back_bulk(in_data.last(to_write));
+        }
+
+        return work_return_t::OK;
+    }
+
+private:
+    void
+    ensure_history_size(std::size_t new_size) {
```
RalphSteinhagen, line 460: N.B. OK for now but we should add a strong 'TODO' to define a strategy how to 
  a) either grow/shrink buffer (otherwise we may end with some out-of-memory exploits),
  b) or: have the `history_buffer<T>` be limited and associated with the poller/callback client (ie. buffer being freed when caller goes out-of-scope).
  
  In any case, this needs to be managed because this will be prone to exploits by external users of this API (and/or via the corresponding OpenCMW worker).

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+                    } else {
+                        poller->drop_count++;
+                    }
+                }
+            }
+        }
+
+        void
+        set_sample_rate(float r) override {
+            sample_delay = std::round(std::chrono::duration_cast<std::chrono::duration<float>>(time_delay).count() * r);
```
RalphSteinhagen, line 857: Naming: 'set_sample_rate' when setting a 'sampling_delay'?

'r' = ??? -> might be useful to define this as a std::chrono::duration type.

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+            if (auto p = polling_handler.lock()) {
+                p->finished = true;
+            }
+        }
+    };
+};
+
+} // namespace fair::graph
+
+ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T), (fair::graph::data_sink<T>), in, sample_rate);
```
RalphSteinhagen, line 898: Missing fields that need to be registered:
  * signal_name
  * signal_unit
  * signal_min
  * signal_max
  
  These would also need to be propagated to the poller/callback function and populating the DataSet<T> meta information

test/qa_data_sink.cpp
```diff
@@ -0,0 +1,688 @@
...
+    }
+
+    static inline bool
+    changed(int x, std::optional<int> other) {
+        return !same(x, other);
+    }
+
+    trigger_observer_state
+    operator()(const tag_t &tag) {
+        const auto ty = tag.get("Y");
```
RalphSteinhagen, line 115: For readability sake:
  * 'Y' -> `Year`
  * `M' -> `Month`
  * `D` -> `Day`
  
  This helps new users/readers to get into the gist of things.

RalphSteinhagen (changes requested):
For my part, this looks generally good.  :+1: 

There are some code and concept duplications and some parts where we probably may profit from some more explicit concept template constraints. The minor changes and comments that should be addressed from my point-of-view are outlined below.

Since this has a bigger impact on the overall user-API and UI-usage, I'd recommend that
@giucam, @ivan-cukic, and @wirew0rm should also have a look at this.

The remaining focus should be IMO on API simplification and minimising code-redundancies but I think the present solution should work.

include/data_sink.hpp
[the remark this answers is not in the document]
frankosterfeld: I added the factory as we said that we need to track multiple instances of the (stateful) observer for some case (overlapping windows?)

include/data_sink.hpp
frankosterfeld: This should also be solved before we move the block to blocklib/, as the registry might have to live with the graph? Alternatively, the graph could provide an interface to query nodes, then the user could create the registry (should get another name then), which would query that interface with data_sink instances.

include/data_sink.hpp
frankosterfeld: it's calculating the delay from the rate (`r`). But yes, naming could be improved.

include/data_sink.hpp
frankosterfeld: From our discussions I got that most users would use non-blocking? Can change this of course.

test/qa_data_sink.cpp
```diff
@@ -0,0 +1,688 @@
...
+        return fmt::format_to(ctx.out(), "{}", tag.index);
+    }
+};
+
+namespace fair::graph::data_sink_test {
+
+static constexpr std::int32_t n_samples = 200000;
+
+template<typename T>
+struct Source : public node<Source<T>> {
```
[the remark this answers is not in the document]
frankosterfeld: Difference is that this one returns n_samples_produced (i.e. std::iota) as sample value with the tests then testing against that to see that they receive the right samples, while the one in tag_monitors returns 1/0. Should I extend your source to take an optional generator function?

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+    Ignore        ///< Ignore tag
+};
+
+// TODO is the scope where want these?
+struct null_type {};
+
+// Until clang-format can handle concepts
+// clang-format off
+template<typename T>
+concept TriggerPredicate = requires(const T p, tag_t tag) {
```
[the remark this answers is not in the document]
RalphSteinhagen: What we discussed for `trigger_observer_state` was to have a `MATCHING`, `NOT_MATCHING`, and `IGNORE` case. The difference w.r.t. to the predicate would be primarily only the 'IGNORE` case.

include/data_sink.hpp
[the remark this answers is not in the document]
RalphSteinhagen: We discussed that -- except for the trigger with it's pre-/post-samples -- there wouldn't be any overlapping ranges for the time being. Let's keep the code KISS and add complexity later when it's actually needed.

include/data_sink.hpp
RalphSteinhagen: For the time being this needs to be only documented. We'll make that move once the graph-scheduler interface advanced a bit more.

include/data_sink.hpp
RalphSteinhagen: Nope... most users want/need a data delivery guarantee... Handling dropped samples is and advanced use-case for those who know what they are doing.

test/qa_data_sink.cpp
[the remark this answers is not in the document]
RalphSteinhagen: KISS ... hiding the n_samples_produced in the return value is not transparent and IMO a too strong coupling/misused functionality between blocks. Also, the downstream nodes shouldn't need to know how many samples have been produced.

include/data_sink.hpp
```diff
@@ -0,0 +1,930 @@
...
+
+template<typename T>
+class data_sink;
+
+struct data_sink_query {
+    std::optional<std::string> sink_name;
+    std::optional<std::string> signal_name;
+
+    static data_sink_query
+    with_signal_name(std::string_view name) {
```
ivan-cukic, line 83: Naming nit-pick - I'd remove `with_`, or replace with `has_` - for me, `with_` implies a non-static member function that behaves like builder pattern - returns a new instance with a certain field updated.

include/data_sink.hpp
ivan-cukic: If implementations were more complex than this, I'd say go for one, but this is trivial, so I'm fine either way - with a single function and with both process_one and process_bulk

include/data_sink.hpp
ivan-cukic: Can you add to the TODO that mutexes/locking should be rethought once it becomes a part of the graph?

include/data_sink.hpp
```diff
@@ -0,0 +1,930 @@
...
+
+enum class trigger_observer_state {
+    Start,        ///< Start a new dataset
+    Stop,         ///< Finish dataset
+    StopAndStart, ///< Finish pending dataset, start a new one
+    Ignore        ///< Ignore tag
+};
+
+// TODO is the scope where want these?
+struct null_type {};
```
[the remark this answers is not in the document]
frankosterfeld: Moved to utils

include/data_sink.hpp
```diff
@@ -0,0 +1,900 @@
...
+
+// Until clang-format can handle concepts
+// clang-format off
+template<typename T>
+concept TriggerPredicate = requires(const T p, tag_t tag) {
+    { p(tag) } -> std::convertible_to<bool>;
+};
+
+template<typename T>
+concept TriggerObserver = requires(T o, tag_t tag) {
```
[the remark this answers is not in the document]
frankosterfeld: Done

include/data_sink.hpp
frankosterfeld: Done

include/data_sink.hpp
frankosterfeld: Done

include/data_sink.hpp
frankosterfeld: Hmm, that'd introduce 5 more `using` declarations that are then used exactly once, saving the `decltype()`. I find it more cryptic in the end, to find out what the type actually is.

include/data_sink.hpp
frankosterfeld: no opinions here either. If single function, would you prefer process_one or process_bulk?

include/data_sink.hpp
frankosterfeld: Ascii art fix Done

include/data_sink.hpp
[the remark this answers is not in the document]
frankosterfeld: Done, removed the factory

include/data_sink.hpp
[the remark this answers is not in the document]
frankosterfeld: Ignore would be equal to NotMatching here

include/data_sink.hpp
frankosterfeld: Done

include/data_sink.hpp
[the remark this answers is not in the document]
frankosterfeld: Unified as suggested

include/data_sink.hpp
frankosterfeld: Add to reflection, not populated/propagated yet

test/qa_data_sink.cpp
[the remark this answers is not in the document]
frankosterfeld: Well, here it's for testing, to make sure the right samples/sample value ranges have been be passed to the poller/callback. That's like 50% of what the tests are testing for.

include/data_sink.hpp
RalphSteinhagen: The idea is not to use the `using` but the whole `port<T>` declaration... there is a lot of stream/tag boiler plating that is duplicated here.

include/data_sink.hpp
RalphSteinhagen: `process_bulk(..)` because the `process_one(...)` as defined here is not SIMD-fied, nor loop-unrolled, nor inlined.

include/data_sink.hpp
ivan-cukic: Do you mean to add a port to poller, or...?

include/data_sink.hpp
frankosterfeld: Ok, will do so. Here process_one() processes whole dataset though, not individual samples, not sure we can help the user with SIMD much?

frankosterfeld: process_bulk is process() now, process_one removed

include/data_sink.hpp
frankosterfeld: Added the defaults, as well as reading the values from settings and tags (settings have preference), and forwarding the info to the datasets.

include/data_sink.hpp
frankosterfeld: Now read from settings or incoming tags (settings have preference), and populating the datasets.

test/qa_data_sink.cpp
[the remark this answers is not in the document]
frankosterfeld: Will be unified by @wirew0rm later.

RalphSteinhagen (approved):
As discussed this looks good! :+1: 

@frankosterfeld thanks for the update and particularly for your patience through this lengthy PR review.

PR #84, opened 2023-06-07
unique node 'id' and 'name'

RalphSteinhagen:
This adds a unique technical identifier for each node created during run-time. The intentional use is to be able to disambiguate between blocks that have been created through cloning and/or for which users set the same `name()` but, for example, require different settings.

### DIFF
--- a/include/node.hpp
+++ b/include/node.hpp
@@ -193,11 +193,15 @@ output_ports(Self *self) noexcept {
  */
 template<typename Derived, typename... Arguments>
 class node : protected std::tuple<Arguments...> {
+    static std::atomic<std::size_t> _unique_id_counter;
+
 public:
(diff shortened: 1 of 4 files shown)


include/node.hpp
```diff
@@ -855,6 +868,9 @@ class merged_node : public node<merged_node<Left, Right, OutId, InId>, meta::con
     }
 };
 
+template<source_node Left, sink_node Right, std::size_t OutId, std::size_t InId>
+inline std::atomic<std::size_t> merged_node<Left, Right, OutId, InId>::_unique_id_counter{ 0_UZ };
```
ivan-cukic, line 872: Maybe `std::atomic_size_t` - whatever you think is more readable

wirew0rm (approved):
Looks good :+1:

PR #79, opened 2023-06-02
Remove conversion's warnings.

drslebedev:
Remove conversion's warnings and make explicit cast where appropriate.
`std::make_signed_t<std::size>` is used for most of the "index" types and `std::size_t` is used for "size" types.
Initially is was planned to remove warnings only for `circular_buffer` class, but since it has many dependencies, other classes were also changed if needed.

**Benchmark bm_buffer results**
The results slightly differ fromm run to run. Here are the 2 randomly selected example outputs.

**Before updates**
```text
┌─────────────────────────benchmark:─────────────────────────┬──────┬─#N─┬─CTX-SW─┬───CPU cache misses────┬───CPU branch misses───┬─<CPU-I>─┬──min───┬──mean──┬─stddev─┬─median─┬───max───┬─total time─┬─ops/s─┐
│                                                            │ SKIP │    │        │                       │                       │         │        │        │        │        │         │            │       │
│ portable: 1 producers -< 1  >-> 1 consumers                │ PASS │ 8  │     8  │ 17.2k / 32.2k = 53.4% │ 3.14k / 29.9k = 10.5% │    2.1m │ 154 ms │ 167 ms │  10 ms │ 171 ms │  185 ms │       1  s │ 59.9M │
│ portable: 1 producers -< 1  >-> 2 consumers                │ PASS │ 8  │     8  │ 15.8k / 29.9k = 52.9% │ 2.99k / 29.3k = 10.2% │    2.1m │ 205 ms │ 222 ms │  14 ms │ 230 ms │  243 ms │       2  s │ 45.0M │
│ portable: 1 producers -< 1  >-> 4 consumers                │ PASS │ 8  │     8  │ 17.0k / 32.5k = 52.4% │ 3.24k / 31.6k = 10.2% │    2.3m │ 302 ms │ 372 ms │  38 ms │ 385 ms │  422 ms │       3  s │ 26.9M │
│ portable: 2 producers -< 1  >-> 1 consumers                │ PASS │ 8  │     8  │ 16.3k / 30.8k = 52.9% │ 3.02k / 29.2k = 10.3% │    2.1m │ 580 ms │ 588 ms │   4 ms │ 591 ms │  593 ms │       5  s │ 17.0M │
[...]

PR #85, opened 2023-06-12
Remove remaining conversion's warnings.

drslebedev:
Continue removing remaining conversion's warnings after #79.

### DIFF
--- a/test/qa_buffer.cpp
+++ b/test/qa_buffer.cpp
@@ -197,7 +197,7 @@ const boost::ut::suite UserApiExamples = [] {
             expect(eq(localReader.available(), std::size_t{ 0 }));
 
             auto lambda = [](auto w) { // test writer generating consecutive samples
-                static int offset = 1;
+                static std::size_t offset = 1;
                 std::iota(w.begin(), w.end(), offset);
                 offset += w.size();
             };
@@ -217,7 +217,7 @@ const boost::ut::suite UserApiExamples = [] {
         // populate with some more data
         for (std::size_t i = 0; i < 3; i++) {
             const auto demoWriter = [](auto w) {
-                static int offset = 1;
+                static std::size_t offset = 1;
                 std::iota(w.begin(), w.end(), offset);
                 offset += w.size();
             };
@@ -260,8 +260,8 @@ const boost::ut::suite CircularBufferTests = [] {
                 BufferReader auto reader = buffer.new_reader();
                 expect(nothrow([&reader] { expect(eq(reader.buffer().n_readers(), std::size_t{ 1 })); })); // created one reader
 
-                int  offset = 1;
-                auto lambda = [&offset](auto w) {
+                std::size_t offset = 1;
+                auto        lambda = [&offset](auto w) {
                     std::iota(w.begin(), w.end(), offset);
                     offset += w.size();
                 };
@@ -443,7 +443,7 @@ const boost::ut::suite StreamTagConcept = [] {
 
         for (int i = 0; i < 3; i++) { // write-only worker (source) mock-up
             auto lambda = [&tagWriter](auto w, std::int64_t writePosition) {
-                static int offset = 1;
+                static std::size_t offset = 1;
                 std::iota(w.begin(), w.end(), offset);
                 offset += w.size();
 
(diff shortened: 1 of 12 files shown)


test/qa_buffer.cpp
```diff
@@ -199,7 +199,7 @@ const boost::ut::suite UserApiExamples = [] {
             auto lambda = [](auto w) { // test writer generating consecutive samples
                 static int offset = 1;
                 std::iota(w.begin(), w.end(), offset);
-                offset += w.size();
+                offset += static_cast<int>(w.size());
```
RalphSteinhagen, line 202: N.B. `auto w` is actually `int32_t w` -> cast should probably be `static_cast<int32_t>(w.size());`

test/qa_buffer.cpp
```diff
@@ -219,7 +219,7 @@ const boost::ut::suite UserApiExamples = [] {
             const auto demoWriter = [](auto w) {
                 static int offset = 1;
                 std::iota(w.begin(), w.end(), offset);
-                offset += w.size();
+                offset += static_cast<int>(w.size());
```
RalphSteinhagen, line 222: same as a above ... maybe change the `auto w` to `int32_t w` to be more explicit/expressive for future readers?!?

same for the other occurrences further below.

test/qa_dynamic_port.cpp
```diff
@@ -125,17 +129,17 @@ const boost::ut::suite PortApiTests = [] {
         auto lambda = [&offset](auto &w) {
```
RalphSteinhagen, line 129: same here ... auto == int32_t

test/qa_scheduler.cpp
```diff
@@ -20,17 +20,17 @@ trace(trace_vector &traceVector, std::string_view id) {
 }
 
 // define some example graph nodes
-template<typename T, std::int64_t N>
+template<typename T, std::make_signed_t<std::size_t> N>
```
RalphSteinhagen, line 23: should be std::size_t

test/qa_scheduler.cpp
```diff
@@ -20,17 +20,17 @@ trace(trace_vector &traceVector, std::string_view id) {
...
 class count_source : public fg::node<count_source<T, N>, fg::OUT<T, 0, std::numeric_limits<std::size_t>::max(), "out">> {
-    trace_vector &tracer;
-    std::int64_t  count = 0;
+    trace_vector                   &tracer;
+    std::make_signed_t<std::size_t> count = 0;
```
RalphSteinhagen, line 26: should be std::size_t

test/qa_settings.cpp
```diff
@@ -85,7 +85,7 @@ struct Source : public node<Source<T>> {
     void
     init(const tag_t::map_type &old_settings, const tag_t::map_type &new_settings) {
         // optional init function that is called after construction and whenever settings change
-        fair::graph::publish_tag(out, { { "n_samples_max", n_samples_max } }, n_tag_offset);
+        fair::graph::publish_tag(out, { { "n_samples_max", n_samples_max } }, static_cast<std::size_t>(n_tag_offset));
```
RalphSteinhagen, line 88: n_tag_offset should probably be std::size_t

test/qa_settings.cpp
```diff
@@ -166,7 +166,7 @@ struct Sink : public node<Sink<T>> {
         */
 
         if constexpr (fair::meta::any_simd<V>) {
-            n_samples_consumed += V::size();
+            n_samples_consumed += static_cast<std::int32_t>(V::size());
```
RalphSteinhagen, line 169: n_samples_consumed should probably be std::size_t

RalphSteinhagen (changes requested):
Thanks for the changes... many conversion warnings are gone. 

While the castings remove the compiler warning, this also shadows an underlying design deficiency that should be fixed at the source. For some changes, I think it would be better to avoid `static_cast<>` by changing the variable type in the first place.

I noted some of the places where I feel this would improve the overall design so that we do not have to touch the code twice.

N.B. this PR correctly fixes the `-Wconversion` warnings but we should eventually also look into the other ones. We should all start looking a bit deeper/closer to those as well.

test/qa_buffer.cpp
drslebedev: I think here everything is correct. 'w' is 'std::span<int>', but I cast 'w.size()' to be the same type as offset, which is 'int'. One of the possibility is to declare 'offset' as 'std::size_t'.

RalphSteinhagen: Maybe 'std::size_t' would be easier and lowering the cognitive complexity ...

RalphSteinhagen (approved):
@drslebedev very good ... thanks for the refactoring, checking and eliminating the warnings! 

Job well done! :+1:

PR #87, opened 2023-06-14
enable settings mgmt via 'graph'

RalphSteinhagen:
* removed template dependency of settings_base ... needed for type-erasure and run-time polymorphism
* exposed unique_name() & settings(), through run-time type-erased interface ...  needed to allow for type-erased modification of block setting parameter.

### DIFF
--- a/include/graph.hpp
+++ b/include/graph.hpp
@@ -388,8 +388,23 @@ public:
 
     virtual ~node_model() = default;
 
+    /**
+     * @brief user defined name
+     */
     virtual std::string_view
     name() const
+            = 0;
+
+    /**
+     * @brief process-wide unique name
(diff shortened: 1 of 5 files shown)


include/graph.hpp
```diff
@@ -475,74 +490,99 @@ class node_wrapper : public node_model {
...
+    }
 
-    std::vector<edge> _edges;
+    [[nodiscard]] constexpr int32_t
+    weight() const noexcept {
+        return _weight;
+    }
+
+    [[nodiscard]] constexpr bool
+    connected() const noexcept {
```
ivan-cukic, line 576: maybe `is_connected`

ivan-cukic (approved):

PR #90, opened 2023-06-15
Windows build support

Agrael1:
The project now builds under clang-cl with windows
Updated fmt dependency to latest version


include/graph.hpp
```diff
@@ -851,7 +851,7 @@ operator<<(std::ostream &os, const port_domain_t &value) {
     return os << static_cast<int>(value);
 }
 
-#ifndef __EMSCRIPTEN__
+#if !defined(__EMSCRIPTEN__) && !(defined(_MSC_VER) && defined(__clang__)) //clang-cl doesn't have source location yet
```
wirew0rm, line 854: The comment only refers to the changed second part of the statement, so it should be either in the commit or specify which part it refers to so that the comment is understandable outside of the diff view.

RalphSteinhagen:
@Agrael1 closing this for now since there are 
  * no associated unit tests 
  * nor CI/CD integration 
to verify that the changes are compatible now and/or for future PRs.

Please feel free to reopen as soon as these issues are resolved.

PR #92, opened 2023-06-19
Fix compiler warnings and incorporate clang-tidy suggestions

wirew0rm:
This commit tries to reduce the number of compiler warnings to zero and performs changes suggested by clang-tidy.

reduction of the number of lines in the respective build steps:
- gcc: 2800 -> 400: all will be fixed by https://github.com/gnuradio/pmt/pull/100, the only one remaining is the node constructor with pmt based settings initialisation. [error message](https://github.com/fair-acc/graph-prototype/actions/runs/5314768958/jobs/9622354105?pr=92#step:10:149)
  I had to disable the `maybe-unused` waring in gcc because it triggers a massive amount of false positives in the standard library in combination with asan. There is an open gcc issue that I linked from the line in the cmake where I disabled it.
- clang: 1800 -> 112: caused by 1 [simd related warning](https://github.com/fair-acc/graph-prototype/actions/runs/5314768958/jobs/9622354392?pr=92#step:10:67)
- emscripten:  4000 -> 2600: mostly caused by size_t related conversion errors in pmt, ut and local code.


RalphSteinhagen (approved):
Very good! :+1:

PR #93, opened 2023-06-21
node.hpp: Fix process_bulk case in constexpr

frankosterfeld:
Need else in if constexpr, otherwise the code below still requires the implementation to have process_one, even if never called. (GCC 12.3.0)

### DIFF
--- a/include/node.hpp
+++ b/include/node.hpp
@@ -631,50 +631,50 @@ public:
             const bool success = consume_readers(self(), samples_to_process);
             forward_tags();
             return success ? ret : work_return_t::ERROR;
-        }
-
-        using input_simd_types  = meta::simdize<typename input_types::template apply<std::tuple>>;
-        using output_simd_types = meta::simdize<typename output_types::template apply<std::tuple>>;
-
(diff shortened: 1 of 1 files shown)


include/node.hpp
```diff
@@ -631,50 +631,50 @@ class node : protected std::tuple<Arguments...> {
...
-                meta::tuple_for_each([i](auto &output_range, const auto &result) { result.copy_to(output_range.data() + i, stdx::element_aligned); }, writers_tuple, results);
-            }
-            simd_epilogue(width, [&](auto w) {
-                if (i + w <= samples_to_process) {
-                    const auto results = simdize_tuple_load_and_apply(w, input_spans, i, [&](auto &&...input_simds) { return invoke_process_one_simd(w, input_simds...); });
-                    meta::tuple_for_each([i](auto &output_range, auto &result) { result.copy_to(output_range.data() + i, stdx::element_aligned); }, writers_tuple, results);
-                    i += w;
-                }
-            });
         } else {
```
frankosterfeld, line 634: Make sure to look at this with "Hide whitespace"...

PR #95, opened 2023-06-21
Function for creating a graph from a given GRC file

ivan-cukic:


test/common_nodes.hpp
```diff
@@ -0,0 +1,185 @@
+#ifndef TEST_COMMON_NODES
```
RalphSteinhagen, line 1: to note @wirew0rm: this should be moved with the next refactoring to the regular blocklib.

For the time being, this is IMO better placed in `./test/blocklib/core/unit-test` ... this helps finding those block implementations that are of global/core interest.

test/qa_plugins_test.cpp
```diff
@@ -147,17 +147,14 @@ const boost::ut::suite BasicPluginNodesConnectionTests = [] {
     };
 
     "Graph"_test = [] {
-        fg::graph                 flow_graph;
+        fg::graph flow_graph;
 
-        auto                      node_source_load     = context().loader.instantiate(names::fixed_source, "double");
-        auto                     &node_source          = flow_graph.add_node(std::move(node_source_load));
+        auto     &node_source     = context().loader.instantiate_in_graph(flow_graph, names::fixed_source, "double");
+        auto     &node_multiply_1 = flow_graph.make_node<builtin_multiply<double>>(2.0);
```
RalphSteinhagen, line 153: We should use only either the default constructor and/or the one accepting a `prarameter_map`...

include/grc.hpp
RalphSteinhagen, line 1: this needs a better name ...

RalphSteinhagen (changes requested):
This looks good and seems to fulfill most of the immediate needs. :+1: 

Some minor things that need to be addressed before being merged:
 * naming and location of some of the source files, common blocks etc. (see comments below)
 * we should drop the user-defined constructor and limit the signature to either
    * a default constructor (no parameters)
    * a constructor accepting the `parameter_map
 * the graph#1->save-to-grc (string) -> load-from-grc (string) -> graph#2 + compare both
 * document the issues to be addressed in the follow-up PR/issues (brief/concise description would be enough)

test/app_grc.cpp
```diff
@@ -0,0 +1,43 @@
+#include <scheduler.hpp>
```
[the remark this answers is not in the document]
ivan-cukic: Yes. In general, I like having both - a test application and a separate unit-test suite

test/common_nodes.hpp
ivan-cukic: I'm not sure the same nodes are of interest to the global public and to tests. In tests, we will probably need strange nodes that nobody should use or write.

test/qa_plugins_test.cpp
ivan-cukic: TBH, I prefer value-based constructors to `parameter_map` ones. For plugin instantiation, only the `parameter_map`-based one can work.

include/grc.hpp
```diff
@@ -0,0 +1,120 @@
+#include <charconv>
+
+#include <yaml-cpp/yaml.h>
+
+#include <graph.hpp>
+#include <plugin_loader.hpp>
+
+namespace fair::graph {
+
+    template <typename ...ts> struct pt;
```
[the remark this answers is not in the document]
RalphSteinhagen: ... shadows of the past. ;-)

include/grc.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+    std::map<std::string, node_model *> created_nodes;
+
+    YAML::Node                          tree   = YAML::Load(yaml_source);
+    auto                                blocks = tree["blocks"];
+    for (const auto &grc_block : blocks) {
+        auto name              = grc_block["name"].as<std::string>();
+        auto id                = grc_block["id"].as<std::string>();
+
+        auto &current_node = loader.instantiate_in_graph(flow_graph, id, "double");
```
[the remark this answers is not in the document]
RalphSteinhagen: :+1:  if you could create/reference an issue here, I'd consider this resolved.

test/app_grc.cpp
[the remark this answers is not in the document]
RalphSteinhagen: For GNU Radio, we should minimise the 'attack-surface' and duplicate tests/examples to minimise the maintenance overhead.  OK. let's document this (-> issue, reference here).

test/common_nodes.hpp
RalphSteinhagen: still blocklib subdir is a better place than the qa root directory....

test/qa_plugins_test.cpp
RalphSteinhagen: OK, then just have the `parameter_map`-based one only and mimick the default through an empty/default map?!?

include/grc.hpp
RalphSteinhagen: and include lock-guards.

include/grc.hpp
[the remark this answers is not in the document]
ivan-cukic: https://github.com/fair-acc/graph-prototype/issues/96

test/qa_plugins_test.cpp
RalphSteinhagen: Great! :+1:

include/grc.hpp
RalphSteinhagen: :+1:

test/app_grc.cpp
```diff
@@ -0,0 +1,45 @@
...
+    using namespace fair::graph;
+    test_context      context(std::move(paths));
+    register_builtin_nodes(&context.registry);
+
+    std::ifstream     input(TESTS_SOURCE_PATH "/grc/test.grc");
+    std::stringstream buffer;
+    buffer << input.rdbuf();
+
+    auto graph = fg::load_grc(context.loader, buffer.str());
+    std::cout << fg::save_grc(graph) << std::endl;
```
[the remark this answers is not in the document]
RalphSteinhagen: nicely done!

include/graph.hpp
```diff
@@ -819,6 +820,8 @@ class graph {
     template<typename Source, typename Sink>
     connection_result_t
     dynamic_connect(Source &source, std::size_t source_index, Sink &sink, std::size_t sink_index) {
+        static_assert(!std::is_pointer_v<std::remove_cvref_t<Source>>);
```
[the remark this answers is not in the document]
RalphSteinhagen: putting these into a `requires {}` clause is also a good option! :+1:

RalphSteinhagen (approved):
Thanks a lot for the changes! This looks good!

PR #98, opened 2023-06-22
upgrade TagSource to user-level production code examples

RalphSteinhagen:


wirew0rm (approved):
Looks good, I hope we can reduce the amount of things a user has to keep in mind for writing a block later on, but I think writing up everything that is required in an example is a good step toward that goal.

PR #99, opened 2023-06-26
Transfer and improve Transactions

vimpostor:
I am opening this PR already, so that the API can be reviewed.

Some things had to be changed from the opencmw implementation to make it more generic. Specifically `TimingCtx` requires a predicate both for matching and for parsing, we can't hardcode the FAIR routines anymore.
And instead of hardcoding members for BPCID, SID, BPID and GID we allow to use a generic identifier map (type is `pmtv::map_t`, which is a `std::map`from `std::string` to `pmt`).

The final thing that is still WIP is the port of `CtxSetting` and the reason why it is not easy to port, is that it needs to default-construct `TimingCtx` instances, which is a bit tricky to get right with the lambdas needed for `TimingCtx`.
The easy way out of this would be to use `std::function` instead, but I want to avoid it due to performance reasons.

At the bottom of `test/qa_settings.cpp` there is an example of the problem, where the compilation fails. The example in the `"SettingBase constructors"_test` works however, because the lambda types are matching. But if we use a different lambda, the type system will complain.

I tried to provide templated default predicates for the default-values in `TimingCtx` like this (and some variations of it) so that a default predicate is created specifically for every different templated type, but wasn't able to make the type system happy with it:

```cpp
template<typename T>
auto defaultParse = [](auto) {};
template<typename T>
auto defaultMatch = [](auto, auto) { return true; };
```

Maybe someone has an idea how to tackle this problem for `CtxSetting`.

The buffer also needed to be ported from opencmw::RingBuffer to the graph-prototype circular buffer.
That's about it for all the major changes, although there are also a lot of other smaller changes when porting all the different components, for example the cache line size is not hardcoded to 64 anymore and instead we try to use `std::hardware_destructive_interference_size` when available.

Closes https://github.com/fair-acc/opencmw-cpp/issues/191


RalphSteinhagen:
> The easy way out of this would be to use std::function instead, but I want to avoid it due to performance reasons.

Will have a look at this. The performance isn't that critical as I initially thought since this code would be only invoked if a tag arrives at the node which is (worst-case) only once every 100 ms. Maybe the use of `std::function` is OK in this case.

Anyway, having a good modular solution now is better than a perfect at an undefined moment in the future. :+1:

include/utils.hpp
```diff
@@ -395,6 +396,22 @@ static_assert(std::is_same_v<std::vector<int>, type_transform<std::vector, int>>
...
 static_assert(std::is_same_v<void, type_transform<std::vector, void>>);
 
+#ifdef __cpp_lib_hardware_interference_size
+static constexpr const std::size_t kCacheLine = std::hardware_destructive_interference_size;
+#else
+static constexpr const std::size_t kCacheLine = 64;
+#endif
+
+constexpr bool
+isPower2(std::unsigned_integral auto value) {
```
RalphSteinhagen, line 406: It could be -- in theory -- expressed as `std::has_single_bit(value)`. We should check whether this is still needed. Most of the buffers can also work with non-powers-of-twos...

include/transactions.hpp
```diff
@@ -0,0 +1,497 @@
...
+ * @tparam U is the internally stored setting type that may include additional meta data.
+ * @tparam TransactionToken unique identifier with which to store/commit transactions.
+ * @tparam N_HISTORY the maximum number of setting history
+ * @tparam TimeDiff the std::chrono::duration time-base for the time-outs
+ * @tparam timeOut maximum time given in units of TimeDiff after which a setting automatically expires if unused. (default: -1 -> disabled)
+ * @tparam timeOutTransactions maximum time given in units of TimeDiff after which a transaction automatically expires if not being committed. (default: -1 -> disabled)
+ */
+template<std::movable T, std::movable U = T, std::equality_comparable TransactionToken = std::string, std::size_t N_HISTORY = 1024, typename TimeDiff = std::chrono::seconds, int timeOut = -1,
+         int timeOutTransactions = -1>
+    requires(fair::meta::is_power2_v<N_HISTORY> && N_HISTORY > 8)
```
RalphSteinhagen, line 114: Is this still functionally needed or just a performance optimisation?

include/utils.hpp
```diff
@@ -395,6 +396,22 @@ static_assert(std::is_same_v<std::vector<int>, type_transform<std::vector, int>>
...
+    return !(value == 0) && !(value & (value - 1));
+}
+
+template<std::unsigned_integral auto N>
+static constexpr bool is_power2_v = isPower2(N);
+
+#define FWD(x) std::forward<decltype(x)>(x)
```
RalphSteinhagen, line 413: While I initially liked/introduce this short-hand... the question is whether this is clean from a STL-lib point-of-view. Maybe it's better to explicitly state the forwarding and types where it is needed rather than using this auto-forward.

Also this may not be obvious to new C++ developers how this works... :thinking:

include/reader_writer_lock.hpp
```diff
@@ -0,0 +1,110 @@
...
+ * in situations where the thread is not allowed to block.
+ *
+ * The lock is implemented using atomic CAS-loops on a counter, which is
+ * incremented (/decremented) when a thread acquires a read (/write) lock, and
+ * decremented (/incremented) when the thread releases the read (/write) lock.
+ *
+ * N.B. The lock is unlocked when the counter reaches 0.
+ */
+class ReaderWriterLock {
+    alignas(64) mutable std::atomic<std::int64_t> _activeReaderCount{ 0 };
```
RalphSteinhagen, line 22: re-use the destructive interference or `fair::meta::kCacheLine` short-hand definition.

RalphSteinhagen (changes requested):
Many parts look familiar but I am a bit confused w.r.t. the usage...

The general idea was/is that these transaction-based settings are an alternative drop-in replacement for `basic_settings<node>` and that is compatible with the following interface:

https://github.com/fair-acc/graph-prototype/blob/8be673430f5fd17b561a8ab09f404aed2726f029/include/settings.hpp#L16-L77

I.e. from the user perspective, the user would `set(..., ctx)` a given `property_map` for a given block and `SettingsCtx ctx` and when a tag passes along where the tag matches the parameters in `ctx` and time, that the block settings would be swapped w.r.t. the matching ones from the transaction-based settings.

Maybe I am missing something. 

Regarding performance -- compared to the previous implementation in OpenCMW -- we changed the settings-change philosophy a bit and only swap-in settings on matching tags which should make the worst-case scenario be something like one settings property_map->struct-reflection-data swap-in every 100 ms. I guess using `std::function` should be sufficient for that. 

Some unnamed/unconstraint template parameters make the code a bit harder to interpret. Maybe it's worthwhile to add some concept-based constraints on these.

The general code structure and qa_tests are thorough! :+1: 

Did I mention that I like unit-tests.... ;-)

include/utils.hpp
vimpostor: Indeed, I have changed it too `std::has_single_bit()` now

include/transactions.hpp
vimpostor: I transfered this requirement from the opencmw implementation, I guess its purpose was just performance.

include/utils.hpp
vimpostor: ACK, I have removed the auto-forwarding

include/reader_writer_lock.hpp
vimpostor: Good catch!

vimpostor:
@RalphSteinhagen Hi, I refactored the implementation to use a normal collection instead of the circular buffer and removed the redundant locking. This simplified the implementation a lot and will hopefully also help with the emscripten issues.

I also adapted the interface so that the multiplexed settings can now properly be used as a drop-in replacement for the basic settings. However, in order to actually change the settings of a node, I had to add a `set_settings()` function in the Node class, but other than that the interface stays the same.

For auto_updating and auto_forwarding, it isn't really clear to which of the multiplexed settings this applies, so I mostly reused the code from the basic settings. Let me know if you had something else in mind here.

The Transactions still use a user-provided matching predicate to match the settings, but there is some special handling without the predicate if there is an exact match available.

include/transactions.hpp
```diff
@@ -0,0 +1,443 @@
...
+
+#include "utils.hpp"
+
+namespace fair::graph {
+
+static auto nullMatchPred = [](auto, auto) { return false; };
+
+template<typename Node>
+class ctx_settings : public settings_base {
+    using MatchPredicate                                = std::function<bool(const property_map &, const property_map &)>;
```
RalphSteinhagen, line 31: As discussed during the meeting today: the existing OpenCMW implementation also supported partial matching, i.e. in case the perfect match hasn't been found the next best one in the (user-defined) hierarchy shall be returned, as illustrated, for example, [here](https://github.com/fair-acc/opencmw-cpp/pull/156). Specifically:

```cpp
opencmw::CtxSetting<int, std::string, 16, std::chrono::seconds, 3600 * 24, 10> settings; // history length of 24 hours, time-out of 10 seconds

assert(settings.commit(TimingCtx(1, 1, 1, 1), 101)); // store 101 for 'TimingCtx(cid=1, sid=1, pid=1, gid=1)'
assert(settings.commit(TimingCtx(1, 1, 1), 102));    // store 102 for 'TimingCtx(cid=1, sid=1, pid=1, gid=-1)'
assert(settings.commit(TimingCtx(1, 1), 103));       // store 103 for 'TimingCtx(cid=1, sid=1, pid=-1, gid=-1)'
assert(settings.commit(TimingCtx(1), 104));          // store 104 for 'TimingCtx(cid=1, sid=-1, pid=-1, gid=-1)'

// retrieve settings
assert(settings.get(TimingCtx(1)) == 104);
assert(settings.get(TimingCtx(1, 1, 1, 1)) == 101);
assert(settings.get(TimingCtx(1, 1, 1, 1)).timingCtx == TimingCtx(1, 1, 1, 1));
assert(settings.get(TimingCtx(1, 1, 1, 1)) == TimingCtx(1, 1, 1, 1)); // short-hand notation
assert(settings.get(TimingCtx(1, 1, 1, 2)) == 102); // no setting for 'gid=2' -> fall back to 'gid=-1'
assert(settings.get(TimingCtx(1, 1, 1, 2)) == TimingCtx(1, 1, 1));
assert(settings.get(TimingCtx(1, 1, 2, 2)) == 103); // no setting for 'pid=2' and 'gid=2' -> fall back to 'pid=gid=-1'
assert(settings.get(TimingCtx(1, 1, 2, 2)) == TimingCtx(1, 1));

// stage new settings
auto [committed, commitTimeStamp] = settings.commit(TimingCtx(1), 55); // immediately store 55 for 'TimingCtx(cid=1)'
assert(not settings.stage(TimingCtx(1), 555, "tokenNo1"));
assert(not settings.stage(TimingCtx(1), 556, "tokenNo2"));

assert(settings.get(TimingCtx(1)) == 55); // staged settings not yet commited
assert(settings.get(TimingCtx(2)) != TimingCtx(2));
assert(settings.commit("tokenNo1"));
assert(settings.get(TimingCtx(1)) == 555);
assert(settings.commit("tokenNo2"));
assert(settings.get(TimingCtx(1)) == 556);
```

If you could also add this as corresponding unit tests. 

As already discussed the simple true/false predicate might not be enough to implement this fall-back logic, maybe we could use a signature like

```cpp
  std::optional<bool>(const property_map &, const property_map &, std::size_t attempt = 0);
  // or complementary
 std::optional<bool>(const property_map &, const property_map &, std::size_t attempt = 0);
```
Where the attempt indicates the search iteration, e.g. 
 * `attempt==0` -> first try, perform a full match of e.g. `(bpcid, cid,sid, pid)`
 * `attempt==1` -> second try, perform a partial match of e.g. `(bpcid, cid,sid)`
 * `attempt==2` -> third try, perform a partial match of e.g. `(bpcid, cid)`
 * `attempt==3` -> fourth try, perform a partial match of e.g. `(bpcid)`
 * `attempt==4` -> fifth try, give up and return `std::nullopt`
 
Hope this is not to hackish but this should at least fit the FAIR specific use-case and hopefully be generic enough to extend to other applications. One could also query the test function for all cases to have only a single pass through all test cases and classify them according to a 'match score', e.g. '0<->perfect match`, '1<->one aspect did not match`. The latter would have the issue of not being explicit which hierarchical search parameter did not match... :thinking: 

Your feedback/improvement proposals would be welcome.

Many thanks in advance.

test/qa_settings.cpp
```diff
@@ -524,6 +530,82 @@ const boost::ut::suite AnnotationTests = [] {
...
+    "CtxSettings Drop-In Settings replacement"_test = [&] {
+        // the multiplexed Settings can be used as a drop-in replacement for "normal" Settings
+        graph flow_graph;
+        auto &block = flow_graph.make_node<TestBlock<float>>({ { "name", "TestName" }, { "scaling_factor", 2.f } });
+        auto  s     = std::make_unique<ctx_settings<std::remove_reference<decltype(block)>::type>>(block, matchPred);
+        block.set_settings(s);
+        auto ctx0   = SettingsCtx(std::chrono::system_clock::now());
+        std::ignore = block.settings().set({ { "name", "TestNameAlt" }, { "scaling_factor", 42.f } }, ctx0);
+        expect(eq(std::get<float>(*block.settings().get("scaling_factor")), 42.f));
+    };
```
RalphSteinhagen, line 606: As discussed above, could you also add the FAIR specific use-case and matcher that would implement the following scenario implemented [here](https://github.com/fair-acc/opencmw-cpp/pull/156), notably:

```cpp
opencmw::CtxSetting<int, std::string, 16, std::chrono::seconds, 3600 * 24, 10> settings; // history length of 24 hours, time-out of 10 seconds

assert(settings.commit(TimingCtx(1, 1, 1, 1), 101)); // store 101 for 'TimingCtx(cid=1, sid=1, pid=1, gid=1)'
assert(settings.commit(TimingCtx(1, 1, 1), 102));    // store 102 for 'TimingCtx(cid=1, sid=1, pid=1, gid=-1)'
assert(settings.commit(TimingCtx(1, 1), 103));       // store 103 for 'TimingCtx(cid=1, sid=1, pid=-1, gid=-1)'
assert(settings.commit(TimingCtx(1), 104));          // store 104 for 'TimingCtx(cid=1, sid=-1, pid=-1, gid=-1)'

// retrieve settings
assert(settings.get(TimingCtx(1)) == 104);
assert(settings.get(TimingCtx(1, 1, 1, 1)) == 101);
assert(settings.get(TimingCtx(1, 1, 1, 1)).timingCtx == TimingCtx(1, 1, 1, 1));
assert(settings.get(TimingCtx(1, 1, 1, 1)) == TimingCtx(1, 1, 1, 1)); // short-hand notation
assert(settings.get(TimingCtx(1, 1, 1, 2)) == 102); // no setting for 'gid=2' -> fall back to 'gid=-1'
assert(settings.get(TimingCtx(1, 1, 1, 2)) == TimingCtx(1, 1, 1));
assert(settings.get(TimingCtx(1, 1, 2, 2)) == 103); // no setting for 'pid=2' and 'gid=2' -> fall back to 'pid=gid=-1'
assert(settings.get(TimingCtx(1, 1, 2, 2)) == TimingCtx(1, 1));

// stage new settings
auto [committed, commitTimeStamp] = settings.commit(TimingCtx(1), 55); // immediately store 55 for 'TimingCtx(cid=1)'
assert(not settings.stage(TimingCtx(1), 555, "tokenNo1"));
assert(not settings.stage(TimingCtx(1), 556, "tokenNo2"));

assert(settings.get(TimingCtx(1)) == 55); // staged settings not yet commited
assert(settings.get(TimingCtx(2)) != TimingCtx(2));
assert(settings.commit("tokenNo1"));
assert(settings.get(TimingCtx(1)) == 555);
assert(settings.commit("tokenNo2"));
assert(settings.get(TimingCtx(1)) == 556);
```

Many thanks in advance.

RalphSteinhagen (changes requested):
@vimpostor, great work on the PR. :+1: 

* The choice of not utilizing `circular_buffer<T>` is sound, especially given that the block-API ensures no real-time/threading data races on the settings object.

* The method of injecting an alternative settings implementation via `set_settings()` is fine. However, it would align better with our upcoming naming conventions for GR 4.0 if it's renamed to camelCase, making it `setSettings(...)`. In GR 4.0, we'll be adhering to the C++ Core Guideline naming convention which dictates UpperCamelCase for user-specific (i.e., non C++ STL) classes and lowerCamelCase for `methodsFunctions(...)` and `double fieldMembers;`.

* Regarding auto-forwarding, the default behavior appears to be appropriate, which is to forward the canonical settings (inclusive of 'context') as defined [here](https://github.com/fair-acc/graph-prototype/blob/ed71b4cde6e4116c25da0ff99c4842f1f1fdc349/include/tag.hpp#L195). If this behavior is inherited from `basic_settings`, then it should be suitable.

* I've listed a few minor comments; kindly refer above/below for them. The primary request is to incorporate hierarchical search/match functionality. This means if a perfect match isn't found, the system should return the next best option, as highlighted [here](https://github.com/fair-acc/opencmw-cpp/pull/156). This might necessitate some API alterations to the matcher function. Although the solution I've sketched might appear a bit hackish, it should cater to the specific FAIR use case and likely many others.

Feel free to suggest any alternative solutions if you envision a better approach.

Beyond the scope of this PR, here's a suggestion for future consideration (possibly create an issue with an SP estimate):
* Integrate transactions within the larger block->scheduler framework, particularly a two-phase commit. In this scenario, settings are staged (with a 'commit token') and validated, but only added to the selection once the predefined 'commit token' is activated through the settings API or indirectly via a `tag_t` on one of the streaming ports or a `pmt_t` message on one of the message ports.

vimpostor:
@RalphSteinhagen Many thanks for the detailed review. I implemented the matching over multiple rounds as requested. My first thought was also that it feels a bit "workaroundy", but after implementing it, it honestly feels like the best solution.
This is also a great solution for usecases, where no hierarchical matching is needed - so they can just ignore the `attempt` argument completely.

> As discussed above, could you also add the FAIR specific use-case and matcher that would implement the following scenario implemented https://github.com/fair-acc/opencmw-cpp/pull/156

The test has been ported and added, it works flawlessy with the new `attempt` API.

RalphSteinhagen (approved):
@vimpostor excellent! Thanks for making these quick changes! :+1:

PR #100, opened 2023-06-27
clean-up some more compiler-warnings & enabled `-Werror`

RalphSteinhagen:
The only remaining warning that needed to be masked were:

  * in graph_yampl_importer.hpp:
```cpp
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wold-style-cast"
#include <yaml-cpp/yaml.h>
#pragma GCC diagnostic pop
```
  -> to be followed up with the yampl-cpp community

  * And in tag_monitors.cpp:
```cpp
for (const auto &[key, value] : map) {
    // workaround for:
    // fmt/core.h:1674:10: warning: possibly dangling reference to a temporary [-Wdangling-reference]
    // 1674 |   auto&& arg = arg_mapper<Context>().map(FMT_FORWARD(val));
    //      |          ^~~
    #pragma GCC diagnostic push
    #pragma GCC diagnostic ignored "-Wdangling-reference"
    fmt::print(" {:>5}: {} ", key, value);
    #pragma GCC diagnostic pop
}

Thanks a lot for the tremendous preparatory work by @drslebedev and @wirew0rm in the previous commits inside and outside the graph-prototype.

### DIFF
--- a/cmake/CompilerWarnings.cmake
+++ b/cmake/CompilerWarnings.cmake
@@ -32,6 +32,7 @@ function(set_project_warnings project_name)
   )
 
   set(CLANG_WARNINGS
+      # -Werror # avoid warnings since they are often indicative of immature API and/or potential sources of bugs # TODO: enable
       -Wall
       -Wextra # reasonable and standard
       -Wshadow # warn the user if a variable declaration shadows one from a parent context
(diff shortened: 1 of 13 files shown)


RalphSteinhagen:
Need to fix differences between gcc13 (tested), gcc12 (doesn't know -Wdangling-reference, ...), and clang.

Should be minor...

RalphSteinhagen:
Tested how far we got w.r.t. eliminating warning so that we eventually could enable `-Werror` and ... we are sooo close: https://github.com/fair-acc/graph-prototype/actions/runs/5392519487

There are a few remaining issues in pmtv and SIMD (their maintainers @jsallay and @mattkretz are informed).

This could be merged as is since it reduces some more warnings.

PR #102, opened 2023-06-28
refactored node name, meta_info as reflectable properties

RalphSteinhagen:
* added fmt definition for Annotated<>
* added implicit conversion operator from/to std::string for Annotated
* properties that cannot be matched to fields (settings().set(..)) end up in the node's meta_information map
* to note: `using property_map = std::map<std::string, pmtv::pmt, detail::transparent_less>;` changed to `pmtv::map_t` (needs follow-up in pmt w.r.t. key comparator)
* `node::meta_information` now stores the field/property annotation meta info (e.g. usable in UI) such as 
   * `<field_name>::description`, 
   * `<field_name>::documentation`,
   * `<field_name>::unit`,
   * `<field_name>::visible`, and 
   * other future annotation meta info.
* changed  `_node->init(/* old settings */ oldSettings, /* new settings */ staged);` to `_node->settings_changed(/* old settings */ oldSettings, /* new settings */ staged);`

### DIFF
--- a/include/annotated.hpp
+++ b/include/annotated.hpp
@@ -131,10 +131,21 @@ struct Annotated {
         }
     }
 
+    template<typename U>
+    Annotated &
+    operator=(const U &sv) noexcept requires std::is_same_v<T, std::string> && std::is_same_v<U, std::string_view> {
+        value = std::string(sv); // Convert from std::string_view to std::string and assign
+        return *this;
+    }
+
+    operator std::string_view() const noexcept requires std::is_same_v<T, std::string> {
+        return std::string_view(value); // Convert from std::string to std::string_view
+    }
+
     // meta-information
     static constexpr std::string_view
     description() noexcept {
-        return std::string_view{ description_ };
+        return std::string_view{description_};
     }
 
     static constexpr std::string_view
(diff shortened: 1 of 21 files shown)


include/tag.hpp
```diff
@@ -121,6 +114,32 @@ ENABLE_REFLECTION(fair::graph::tag_t, index, map);
...
+        if (auto nested_map = std::get_if<pmtv::map_t>(&value)) {
+            // If it's a nested map
+            if (auto it = dest.find(key); it != dest.end()) {
+                // If the key exists in the destination map
+                auto dest_nested_map = std::get_if<pmtv::map_t>(&(it->second));
+                if (dest_nested_map) {
+                    // Merge the nested maps recursively
+                    merge_maps(*nested_map, *dest_nested_map);
+                } else {
+                    // Key exists but not a map, replace it
```
ivan-cukic, line 129: Unsure this is the expected behavior of a function called merge. Can you document the reason for the behavior and why it is fine to forget the old value?

include/annotated.hpp
```diff
@@ -131,10 +131,21 @@ struct Annotated {
         }
     }
 
+    template<typename U>
+    Annotated &
+    operator=(const U &sv) noexcept requires std::is_same_v<T, std::string> && std::is_same_v<U, std::string_view> {
```
ivan-cukic, line 136: `&& std::is_same_v<U, std::string_view>`

I'd replace this with a `is_constructible_v<T, U>` and make `sv` perfectly-forwarded (maybe someone wants to use std::string instead of std::string_view, or something else that converts to string)

include/settings.hpp
```diff
@@ -234,10 +275,10 @@ class basic_settings : public settings_base {
...
             std::lock_guard lg(_lock);
             for (const auto &[localKey, localValue] : parameters) {
-                const auto &key    = localKey;
-                const auto &value  = localValue;
-                bool        is_set = false;
-                for_each(refl::reflect(*_node).members, [&, this](auto member) {
+                const auto &key                 = localKey;
+                const auto &value               = localValue;
+                bool        is_set              = false;
+                auto        iterate_over_member = [&, this](auto member) {
```
ivan-cukic, line 281: Not a fan of the naming `iterate_over_member` - the names should reflect what the function does - reading `for_each(..., iterate_over_member)` doesn't say much

include/tag.hpp
wirew0rm: `update_map` ? Would also work better with the `dest_` parameter name, imho.

RalphSteinhagen: done... adopted `update_map`

include/settings.hpp
RalphSteinhagen: me neither... just refactored this out because this lambda needs to be called twice: once for the base class and once for the user block implementation.

RalphSteinhagen: @ivan-cukic feel free to make a proposal how and/or change these names directly.

ivan-cukic (approved):

PR #103, opened 2023-06-29
scheduler: have run_and_wait reset a stopped scheduler

wirew0rm:
Fixes benchmarks that run an existing scheduler repeatedly.


RalphSteinhagen (approved):
LGTM!

PR #130, opened 2023-07-17
fixed Heisenbug in the circular_buffer<T>

RalphSteinhagen:
... that rarely occurs while remapping the 'second_copy_addr' that may fail according to the docs:

"If another mapping already exists there, the kernel picks a new address that may or may not depend on the hint."


wirew0rm (approved):
Looks good, good step to make the code more stable.
Low coverage is to be expected as long as we don't have a clear strategy for testing error paths which a lot of the changes consist of.

PR #132, opened 2023-07-19
CI: update restyled to clang-format 18

wirew0rm:
update restyled to use clang-format 18. Fixed on a specific image so there will be no surprise reformatting if clang-format gets it's next release.

Also disable creating pull requests and instead only push the branch and set the status check to decrease the amount of PRs especially for Draft PRs.

Since clang-format is not yet enabled here, we could also wait with merging this until we have finalized the clang-format rules.

Depends on restyled upstream clang-format support: https://github.com/restyled-io/restylers/pull/735

### DIFF
--- a/.restyled.yaml
+++ b/.restyled.yaml
@@ -1,51 +1,9 @@
-# Default Restyler Configuration
-#
-# All keys are optional and default as shown. All list values can also be given
-# a single value, to indicate a single-element list.
-#
-####
+# Default Restyler Configuration: https://github.com/restyled-io/restyler/blob/main/config/default.yaml
+pull_requests: false # Open Restyle PRs?
+request_review: none
+labels: []
 
-# Do anything at all?
-enabled: true
-
-# Patterns to exclude from all Restylers
-#
-# By default, we ignore directories that are often checked-in but rarely
-# represent project code. This glob is slightly complicated to match paths
-# within directories of these names appearing at any depth.
-#
-# This behavior can be disabled in your project with:
-#
-#   exclude: []
-#
-exclude:
-  - "**/node_modules/**/*"
-  - "**/vendor/**/*"
-
-# Push the style fixes directly to the original PR
-#
-# This setting implies pull_requests: false for origin PRs, and has no effect on
-# forked PRs (since we can't push to those).
-#
-auto: false
-
-# Download remote files before restyling
-#
-# Example:
-#
-#   remote_files:
-#     - url: https://raw.github.com/.../hlint.yaml
-#       path: .hlint.yaml
-#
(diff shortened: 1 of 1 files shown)


RalphSteinhagen:
@wirew0rm @ivan-cukic @mormj @willcode @marcusmueller We should nail down the clang-formatting rules sooner rather than later.

I initially hoped to have a community-driven, majority-voting process but this may take too much time before GrCon23 and getting this prototype as the new GR 4.0 draft in. If this is a bridge too far, we could postpone this until this repo has been migrated to the gnuradio organisation.

Maybe we could do a 2-step mini-workshop going through the clang-formatting rules? 
  1. agree on the overall goals and non-goals
  2. going through the clang-format rule one-by-one and see which and whether some of these are enforced.
  
Similarly for the sonar[lint,cube] QA rules (but in another round, since the default is already quite reasonable IMO).
  
At least for my part, I do not want to discuss code formatting, spelling, or nit-picking things in PR reviews while losing track of the bigger picture aspects like architecture design, performance etc.

Your input/feedback on this would be much appreciated.

marcusmueller:
Re: mini-workshop: I'd be available for that. 

Re: The goal/non-goal discussion: I think GR's reasons for having a .clang-format file that is applied for every PR is pretty much the same as yours. There's just enough substantial things one can discuss about code, delegating the "mechanical" aspect of code style to a machine takes

1. work load off the reviewer, because they don't have to do that;
2. mental load off the author, as they can let a machine format their code in a compatible way;
3. personal friction off the project, as "computer says no, and we agreed to listen to the computer on that" is just one less thing to argue about.

Matter of fact, GR went through the pain of doing a whole-tree reformat, which IMHO greatly improved maintainability for years now, at the short-term cost of "invalidating" all forks/merge requests. I can recommend doing that should we settle on a code style that requires changing existing code.

RalphSteinhagen:
@wirew0rm and @drslebedev we should move this to clang-format18 now and settled on a common format. 

As discussed earlier in https://github.com/fair-acc/graph-prototype/issues/311 we should:

> [ ] improved clang-format definition, needs draft for further discussions and evaluation => **ACTION:** Semen
>  *focus on keeping intentional line breaks/long-lines & being vertically compact otherwise*

Let's keep the ball rolling and finish this so that we can release a GR 4.0 beta0 version

wirew0rm:
Updated to clang-format-18, which has a pending PR for support in restyled https://github.com/restyled-io/restylers/pull/735 which usually are merged pretty quickly.

Actual clang-format configuration is a different issue and since there is no agreed upon configuration right now this has no effects since restyled is disabled on this repository.

PR #133, opened 2023-07-19
Sonarcloud fixes

wirew0rm:
Fixes some of the code issues that sonarcube complains about.
Focuses on std::forward/std::move issues.

### DIFF
--- a/test/qa_settings.cpp
+++ b/test/qa_settings.cpp
@@ -15,6 +15,21 @@ template<>
 auto boost::ut::cfg<boost::ut::override> = boost::ut::runner<boost::ut::reporter<>>{};
 #endif
 
+template <typename T>
+struct fmt::formatter<std::complex<T>>{
+    template<typename ParseContext>
+    auto
+    parse(ParseContext &ctx) {
+        return std::begin(ctx);
+    }
+
+    template<typename FormatContext>
+    auto
+    format(const std::complex<T> value, FormatContext &ctx) const {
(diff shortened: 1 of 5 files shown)


test/qa_settings.cpp
```diff
@@ -15,6 +15,21 @@ template<>
 auto boost::ut::cfg<boost::ut::override> = boost::ut::runner<boost::ut::reporter<>>{};
 #endif
 
+template <typename T>
```
wirew0rm, line 18: Do we want this somewhere more general in the project? Fmt does not want to include a default one because it is simple to define in user-code and there was a proposal to add it to std::format, which was dropped from c++23.
If yes, where would we put it?

RalphSteinhagen: On the one hand, I agree that this is very common and that it is tempting to have a global definition, on the other hand, since this library will be used by many (who may have conflicting definitions) and this is only (for now) used in a unit-test context, I'd tend to not include this yet.

Also, this would IMHO expand the maintenance surface/library API. It's easy to add but unless this becomes critical to the core functionality, I'd leave it out.

We may eventually add an optional convenience library that contains all of these nick nags and that user may include in their top-level application but as a core feature this seems to be a bit heavy and also dangerous (due to potential conflict with other definitions).

RalphSteinhagen (approved):
Looks nice and good job hunting down all these issues and potential bugs ... :+1: 

We should add some metric/mojo factor for all who improve and simplify the code w/o adding additional maintenance burden. :-)

PR #134, opened 2023-07-20
Bug fixing in node::forward_tags method.

drslebedev:
Span size is 1 and index must always be 0.


RalphSteinhagen (approved):
Thanks for the quick-fix! LGTM :+1:

wirew0rm (approved):
:+1:

PR #136, opened 2023-07-20
Fix compiler error for clang15/16: call to format_to is ambiguous

drslebedev:


wirew0rm (approved):
:+1:  looks good thanks for fixing!

PR #137, opened 2023-07-21
CMake: Improve build type defaults

mattkretz:
* Since we have -Og, -O0 (the default) is never a good idea.
* Building the project without a build type (the cmake default) is "usually not desirable" (cmake docs), therefore default to Release with assertions enabled.
* Document build types.
* Warn if the user somehow managed to build without optimization.


mattkretz:
Clang repos are down, that's why Clang is failing.

PR #138, opened 2023-07-21
Modularisation to prepare gr4.0 integration

wirew0rm:
Restructure to a more gnuradio compatible directory structure.

TODO in this PR:
 - [x] reenable block unit tests
 - [x] move algorithm to separate mini library
 - [x] disable tests and examples by default if we are not the top level cmake project to simplify fetch_content use
 - [x] namespaces

TODO in later PR:
 - [ ] meson (without emscripten)
 - [ ] fix single header(s) generation
 - [ ] extract common nodes from unittests to block library
 - [ ] generify cmake functions for benchmark/test registration
 - [ ] plugin system
 - [ ] namespace versioning
 - [ ] add install directives


RalphSteinhagen (approved):
Additionally renamed header include-guards to GNURADIO_... and removed the renamed/split qa_fft test from the coverage runner

PR #139, opened 2023-07-21
demo clock source

RalphSteinhagen:
### Implements a BlockingIO clock source that generates burst samples with an average constant sampling rate.

The meaty part for the end-user and targeted reference implementation for blocking IO is:
```cpp
template<typename T, typename ClockSourceType = std::chrono::system_clock, bool basicPeriodAlgorithm = true>
class ClockSource : public node<ClockSource<T, ClockSourceType>, BlockingIO, Doc<R""(
ClockSource Documentation -- add here
)"">> {
    std::chrono::time_point<ClockSourceType> nextTimePoint = ClockSourceType::now();

public:
    OUT<T>             out;
    //
    A<std::uint64_t, "n_samples_max", Visible, Doc<"0: unlimited">>              n_samples_max = 1024;
    std::uint64_t                                                                n_samples_produced{ 0 };
    A<std::uint64_t, "chunk_size", Visible, Doc<"number of samples per update">> chunk_size  = 100;

    void
    settings_changed(const property_map & /*old_settings*/, const property_map & /*new_settings*/) {
        nextTimePoint = ClockSourceType::now();
    }

    work_return_status_t
    process_bulk(PublishableSpan auto &output) noexcept { // <- called in a separate thread from the ioThreadPool
        if (n_samples_max > 0 && n_samples_produced >= n_samples_max) {
            output.publish(0_UZ); // <- N.B. PublishableSpan allows to publish an arbitrary number of samples -> is explicit (assertion)
            return work_return_status_t::DONE;
        }

        std::this_thread::sleep_until(nextTimePoint); // <- blocking here
        const std::size_t writableSamples = output.size();
        if (writableSamples < chunk_size) {
            output.publish(0_UZ);
            return work_return_status_t::INSUFFICIENT_OUTPUT_ITEMS;
        }

        const std::size_t remaining_samples  = n_samples_max - n_samples_produced;
        const std::size_t limit              = std::min(writableSamples, remaining_samples);
        const std::size_t n_available        = std::min(limit, chunk_size.value);
        // [...]
        output.publish(std::min(n_available, n_samples_max.value));
        n_samples_produced += samples_to_produce;


        const auto updatePeriod = std::chrono::microseconds(static_cast<long>(1e6f * static_cast<float>(chunk_size) / sample_rate));
        nextTimePoint += updatePeriod;
    }
};
```
Added a graph-specific atomic 'progress' counter that can be listened to that is updated whenever a blocking block unblocked and processed data. This can be used, for example, by the scheduler to sleep -- if no other block can process data -- and be woken up when one of the blocking blocks produced some data.

###  Added `node::available_[input,output]_samples(Container&)` and `DONE` workaround
   ... re-checking whether there are no pending unprocessed samples for blocks implementing the BlockingIO policy. The issue is that the present 'DONE' concept assumes a global state that does not hold up for all multithreading corner cases.

To be discussed: a better solution could be to propagate the `DONE` state to dependent blocks (i.e. via tags) and only declare `DONE` until all blocks ordered from source to sink can communicate done. ... to be further evaluated.


RalphSteinhagen:
closing now in order to re-open

PR #140, opened 2023-07-23
demo clock and source handling BlockingIO

RalphSteinhagen:
### Implements a BlockingIO clock source that generates burst samples with an average constant sampling rate.

The meaty part for the end-user and targeted reference implementation for blocking IO is:
```cpp
template<typename T, typename ClockSourceType = std::chrono::system_clock, bool basicPeriodAlgorithm = true>
class ClockSource : public node<ClockSource<T, ClockSourceType>, BlockingIO, Doc<R""(
ClockSource Documentation -- add here
)"">> {
    std::chrono::time_point<ClockSourceType> nextTimePoint = ClockSourceType::now();

public:
    OUT<T>             out;
    //
    A<std::uint64_t, "n_samples_max", Visible, Doc<"0: unlimited">>              n_samples_max = 1024;
    std::uint64_t                                                                n_samples_produced{ 0 };
    A<std::uint64_t, "chunk_size", Visible, Doc<"number of samples per update">> chunk_size  = 100;

    void
    settings_changed(const property_map & /*old_settings*/, const property_map & /*new_settings*/) {
        nextTimePoint = ClockSourceType::now();
    }

    work_return_status_t
    process_bulk(PublishableSpan auto &output) noexcept { // <- called in a separate thread from the ioThreadPool
        if (n_samples_max > 0 && n_samples_produced >= n_samples_max) {
            output.publish(0_UZ); // <- N.B. PublishableSpan allows to publish an arbitrary number of samples -> is explicit (assertion)
            return work_return_status_t::DONE;
        }

        std::this_thread::sleep_until(nextTimePoint); // <- blocking here
        const std::size_t writableSamples = output.size();
        if (writableSamples < chunk_size) {
            output.publish(0_UZ);
            return work_return_status_t::INSUFFICIENT_OUTPUT_ITEMS;
        }

        const std::size_t remaining_samples  = n_samples_max - n_samples_produced;
        const std::size_t limit              = std::min(writableSamples, remaining_samples);
        const std::size_t n_available        = std::min(limit, chunk_size.value);
        // [...]
        output.publish(std::min(n_available, n_samples_max.value));
        n_samples_produced += samples_to_produce;


        const auto updatePeriod = std::chrono::microseconds(static_cast<long>(1e6f * static_cast<float>(chunk_size) / sample_rate));
        nextTimePoint += updatePeriod;
    }
};
```
Added a graph-specific atomic 'progress' counter that can be listened to that is updated whenever a blocking block unblocked and processed data. This can be used, for example, by the scheduler to sleep -- if no other block can process data -- and be woken up when one of the blocking blocks produced some data.

###  Added `node::available_[input,output]_samples(Container&)` and `DONE` workaround
   ... re-checking whether there are no pending unprocessed samples for blocks implementing the BlockingIO policy. The issue is that the present 'DONE' concept assumes a global state that does not hold up for all multithreading corner cases.

To be discussed: a better solution could be to propagate the `DONE` state to dependent blocks (i.e. via tags) and only declare `DONE` until all blocks ordered from source to sink can communicate done. ... to be further evaluated.


RalphSteinhagen:
Emscripten blocks on the CI ... need to further investigae.

drslebedev:
The idea is that basically only source nodes "knows" that no more data should be expected.
What if we consider a variant when only source(s) can send `DONE` tag. This tag(s) should be propagated through all nodes and when node receives this tag it returns `DONE`. One does not need workarounds for blocking nodes or potentially some other edge cases. What are the potential disadvantages of such an approach?

include/node.hpp
```diff
@@ -733,8 +820,28 @@ class node : protected std::tuple<Arguments...> {
...
+                            ioThreadShallRun.exchange(false, std::memory_order_relaxed);
+                        }
+                    }
+                    ioThreadRunning.fetch_sub(1);
+                    ioThreadRunning.notify_all();
+                });
+            }
+            const work_return_status_t lastStatus                 = ioLastWorkStatus.exchange(work_return_status_t::OK, std::memory_order_relaxed);
+            const auto &[accumulatedRequestedWork, performedWork] = ioWorkDone.getAndReset();
+            return { accumulatedRequestedWork, performedWork, performedWork > 0 ? work_return_status_t::OK : lastStatus };
```
drslebedev, line 844: I localize now the problem with `emsripten`
For the moment I could fix the problem by changing this lne to 
`return { accumulatedRequestedWork, performedWork, lastStatus };`.
But I need some time to understand why `performedWork` is always > 0.

drslebedev: One solution is to implement `available_samples` method for `ClockSource` class:
```cpp
    constexpr std::make_signed_t<std::size_t>
    available_samples(const ClockSource & /*self*/) noexcept {
        const auto ret = static_cast<std::make_signed_t<std::size_t>>(n_samples_max - n_samples_produced);
        return ret > 0 ? ret : -1; // '-1' -> DONE, produced enough samples
    }
```
With this change ll test passed on my computer.

drslebedev: Also one needs to be accurate with a `chunk_size` and add some additional checks:
```cpp
    constexpr std::make_signed_t<std::size_t>
    available_samples(const ClockSource & /*self*/) noexcept {
        const auto ret = static_cast<std::make_signed_t<std::size_t>>(n_samples_max - n_samples_produced);
        if (ret < chunk_size) return -1;
        return ret > 0 ? ret : -1; // '-1' -> DONE, produced enough samples
    }
```

and in `process_bulk` function:
```cpp
        if (writableSamples < chunk_size && n_samples_max - n_samples_produced < chunk_size) {
            output.publish(0_UZ);
            return work_return_status_t::DONE;
        }
```

But with these checks one can produce less samples and tests fail.

drslebedev: Here is a branch https://github.com/fair-acc/graph-prototype/tree/deadlock_emscripten

RalphSteinhagen: @drslebedev we want to get rid of the `available_samples(..)` functions for sources. I fond a workaround and pushed again. Should work now. 

Your past decimator/interpolator+stride PR seems to have some regression w.r.t. that the number of samples that are processed as  a batch by `process_bulk(...)` and `process_one(...)` isn't limited to the min of available input samples and the next sample with an attached tag. I commented the corresponding test. Maybe you could investigate and fix this after this PR is merged.

wirew0rm (approved):
Important functionality which is needed for the implementation of important block types like the ones for network, actual hardware, etc. There are some things that can probably be simplified later on but need other parts of the code changed first.

- Is there a way to circumvent the additional thread for (almost) all block types by using async APIs? For now it is included, but if we determine that everything works without it, we can later remove that part again, simplifying the work function.
- Is there a way to separate some of the logic and state out of the `node` type again in the future such that it doesn't impact simple blocks?
- In #148 and the message passing issue (tbd) we will have to revisit the way the nodes/blocks communicate their processing state. At the moment each block handles its own state and checks with any other block whether it should wait for future data. OTOH checking for available data already exploits the graph topology via the buffers between the nodes. Once this is unified, this code can probably also be simplified a lot.

While this is probably not going to stay in its current form, this state fulfills the immediate needs for implementing blocking nodes that we need rather sonner than later, is well tested and will allow us to gain experience implementing blocking code.

PR #151, opened 2023-08-17
Fix segfault when using the fft block

giucam:
The fft block is freeing fftw memory in its destructor which is shared among all of them. Use a refcount mechanism instead to only free it when there is no user left.

### DIFF
--- a/test/blocklib/core/fft/fft.hpp
+++ b/test/blocklib/core/fft/fft.hpp
@@ -27,9 +27,32 @@ struct fft : node<fft<T>> {
 public:
     static_assert(not LongDoubleType<T>, "long double is not supported");
 
+    struct fftw_base {
+    private:
+        static inline int        counter = 0;
+        static inline std::mutex mutex;
+
+    protected:
+        template<auto F>
+        static void
+        cleanup_impl() {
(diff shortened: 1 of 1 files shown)


drslebedev:
Thanks for your PR! Can you please give a use case example with memory sharing? In general this was not intended to share `fftw` internal memory between several bocks.

giucam:
It's entirely possible i'm doing something wrong. I encountered the segfault when i have a graph with an fft block and i create a new graph with a new fft block and then delete the old graph. The fftw functions don't take a context object, so it seems to me they must be using some static memory which is shared between blocks.

drslebedev:
@giucam Please look at my PR https://github.com/fair-acc/graph-prototype/pull/152. I added support for DataSet<float> and DataSet<double> for `port` and add example of usage FFT block with low graph.

drslebedev:
This PR can be closed. The issue addressed in this PR were solved in 
https://github.com/fair-acc/graph-prototype/pull/154 and https://github.com/fair-acc/graph-prototype/pull/152. 
The problem was mostly in a usage of `fftw_cleanup` function. 
New unit tests were also added for this case.

PR #153, opened 2023-08-22
Fix plugin test build

frankosterfeld:
Make sure the generated export header is found when building the plugin tests. Seen when using this repo via FetchContent_Declare().


wirew0rm (approved):
Looks good, thanks for investigating and fixing!

The CI error is unrelated to the change as confirmed locally and masked in #155.

PR #154, opened 2023-08-23
Thread safe fftw plan creation and destroy. Remove fftw_cleanup.

drslebedev:
This PR includes thread-safe plan creation and destruction. 
Also it remove fftw_cleanup.
This PR should fix problems when several FFT blocks are created and destroyed or run in parallel.


wirew0rm (approved):
Looks good.

#155 should fix the CI build failure by disabling ASAN temporarily for the plugin tests.

PR #155, opened 2023-08-23
cmake: disable asan for qa_plugin_test

wirew0rm:
temporarily disable asan for test which fails due to changes in the github action environment.

runner release that probably caused the failures: https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20230814.1

Only  mentioned related change is a micro version in cmake.

Also resolves all gcc warnings.


drslebedev (approved):
Thanks for the changes, they look good to me! PR can be merged.

PR #156, opened 2023-09-01
Several changes in FFT

drslebedev:
* Implement window (apodization) function.
* Meta information for DataSet.
* unique_ptr for in_data, out_data and plan.
* Compute ranges as part of copying to DataSet.
* Add integral type to FFT tests.
* fftw wisdom import/export via string.
* Additional unit tests.
* General improvements and bug fixing.
* Change code style according to opendigitizer style guide.

#### Note for reviewer: 
- `enum class window_function : int {...}` includes all supported window types.
- `window` type is declared as `int` in `fft` class to make it available to change it via settings mechanism.
- where it is needed `static_cast` is used to convert between `int` and `window_function` and vice versa.
- @mattkretz suggested another possible solution:
```cpp
struct EnumLike
{
  int value;

  template <Enum T>
  EnumLike(T x) : value(int(x)) {}
  
  template <Enum T>
  operator T() const { return T(value); }
};
```
But it requires to add `EnumLike` as `pmt` supported type.
See example here: https://godbolt.org/z/h1xjjxGn4

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,13 +17,13 @@ if (EMSCRIPTEN)
     target_compile_options(graph-prototype-options INTERFACE
             -fwasm-exceptions
             -pthread
-            )
+    )
     target_link_options(graph-prototype-options INTERFACE
             "SHELL:-s ALLOW_MEMORY_GROWTH=1"
             -fwasm-exceptions
             -pthread
             "SHELL:-s PTHREAD_POOL_SIZE=30"
-            )
+    )
 endif ()
 
 
@@ -92,7 +92,7 @@ ExternalProject_Add(fftw_ext
         BUILD_COMMAND ""
         INSTALL_COMMAND ""
         LOG_DOWNLOAD ON
-        )
+)
 
 include_directories(${FFTW_PREFIX}/install/include)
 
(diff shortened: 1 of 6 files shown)


RalphSteinhagen (approved):
This is OK!

Some of the open points to be tackled next
  * less reliance on 'fftw' (GPLv3 <-> licencing issues, we want to remain LGPLv3)
  * meson integration
  * optional computation of signal errors (via boolean flag -> open topic for GR architecture group)
  * special export/return for `std::vector<std::complex<T>>`
  * use of more modern C++20/23 paradigms (<-> maintain emscripten/libc++ compatibility), e.g.
   ```cpp
// [..]
case WindowFunction::Hamming: {
  const T a = pi2 / static_cast<T>(n);
  for (std::size_t i = 0; i < n; i++) res[i] = static_cast<T>(0.53836) - static_cast<T>(0.46164) * std::cos(a * static_cast<T>(i));
  return res;
}
// [..]    
   ```
   (which also misses the `{}` pair after the for loop) would be more compiler and SIMD-optimisable-friendly as
   ```cpp
case WindowFunction::Hamming: {
    const T a_factor = 2 * std::numbers::pi_v<T> * static_cast<T>(0.46164) / static_cast<T>(n);
    std::ranges::transform(std::views::iota(std::size_t(0), n), begin(res), [a_factor](std::size_t i) {
            return static_cast<T>(0.53836 - std::cos(a_factor * i));
    });
    return res;
}
```

PR #157, opened 2023-09-03
updated FIR and IIR filter doc

RalphSteinhagen:


wirew0rm (approved):
lgtm

we have to investigate why the emscripten test fails, but that's unrelated to this pull request

PR #158, opened 2023-09-04
Allow registering nodes with several type template parameters

ivan-cukic:


RalphSteinhagen (approved):
Generally, looks good enough. We need to tackle the final naming later. 

At this stage: do you want to stick to the `node_type_parameters` naming? 
I'd propose to use the CamelCase in this case since this is a more 'custom' definition and not STL-standard like (i.e. apply OpenCMW code-style rules).

The main two issues I see are:
  * The naming, because it doesn't really capture that these are the port types (and not 'node')
  * That there is another `SupportedTypes` with similar intention which IMO should be replaced/retired by this.

If you'd have the courage to refactor/throw away the old(er) bit now so that we are not accumulating debt here ... otherwise we tackle this minor thing during a later refactoring.

PR #163, opened 2023-09-13
Use std::ranges to calculate FFT window function.

drslebedev:


RalphSteinhagen (approved):
As discussed in person, LGTM! :+1:

PR #164, opened 2023-09-13
refactor settings_changed(old, new, fwd) & Decimate example block

RalphSteinhagen:
... to account for old- vs. new- vs. fwd settings.

Example:

```cpp
 template<typename T, bool Average = false>
struct Decimate : public node<Decimate<T, Average>, SupportedTypes<float, double>, PerformDecimationInterpolation, Doc<R""(
@brief reduces sample rate by given fraction controlled by denominator
)"">> {
    IN<T>                            in{};
    OUT<T>                           out{};
    A<float, "sample rate", Visible> sample_rate = 1.f;

    void
    settings_changed(const property_map & /*old_settings*/, property_map &new_settings, property_map &fwd_settings) noexcept {
        if (new_settings.contains(std::string(fair::graph::tag::SIGNAL_RATE.shortKey())) || new_settings.contains("denominator")) {
            const float fwdSampleRate = sample_rate / static_cast<float>(this->denominator);
            fwd_settings[std::string(fair::graph::tag::SIGNAL_RATE.shortKey())] = fwdSampleRate; // TODO: handle 'gr:sample_rate' vs 'sample_rate';
        }
    }

    constexpr work_return_status_t
    process_bulk(std::span<const T> input, std::span<T> output) noexcept {
        assert(this->numerator == std::size_t(1) && "block implements only basic decimation");
        assert(this->denominator != std::size_t(0) && "denominator must be non-zero");

        auto outputIt = output.begin();
        if constexpr (Average) {
            for (std::size_t start = 0; start < input.size(); start += this->denominator) {
                constexpr auto chunk_begin = input.begin() + start;
                constexpr auto chunk_end   = chunk_begin + std::min(this->denominator, std::distance(chunk_begin, input.end()));
                *outputIt++                = std::reduce(chunk_begin, chunk_end, T(0)) / static_cast<T>(this->denominator);
            }
        } else {
            for (std::size_t i = 0; i < input.size(); i += this->denominator) {
                *outputIt++ = input[i];
            }
        }

    }
};
```

Also fixed some tag-forwarding bugs related to `map::insert(...)`  which does not overwrite existing map entries and `map::insert_or_assign(...)` which does.

### DIFF
--- a/include/node.hpp
+++ b/include/node.hpp
@@ -433,7 +433,15 @@ public:
     init(std::shared_ptr<gr::Sequence> progress_, std::shared_ptr<fair::thread_pool::BasicThreadPool> ioThreadPool_) {
         progress     = std::move(progress_);
         ioThreadPool = std::move(ioThreadPool_);
-        std::ignore  = settings().apply_staged_parameters();
+        if (const auto forward_parameters = settings().apply_staged_parameters(); !forward_parameters.empty()) {
+            std::for_each(_tags_at_output.begin(), _tags_at_output.end(), [&forward_parameters](tag_t &tag) {
+                for (const auto &[key, value] : forward_parameters) {
+                    tag.map.insert_or_assign(key, value);
+                }
+            });
+            _output_tags_changed = true;
+        }
+
         // TODO: expand on this init function:
         //  * store initial setting -> needed for `reset()` call
         //  * ...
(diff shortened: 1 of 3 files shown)


wirew0rm (approved):
Looks good :+1:

PR #165, opened 2023-09-14
CI: Remove clang-15, add gcc-13.

drslebedev:
This PR contains the following changes in CI:
* Drop support of `clang-15`
* Add `gcc-13`


wirew0rm (approved):
Looks good!

PR #166, opened 2023-09-14
added store/reset default block setting functionality

RalphSteinhagen:
Based on a suggestion/request by @Jsallay added functionality to reset to and store default block parameter. 

The initial defaults are those settings during block initialisation, including the (optional) parameters that have been passed as constructor arguments, e.g. 
```cpp 
auto &block = flow_graph.make_node<TestBlock<float>>({ { "name", "TestName" }, { "scaling_factor", 2.f } });
```

The reset and store can be triggered via streaming tags `gr:reset_default` and ``gr:store_default` (N.B. naming open for change), or explicitly via the setting accessor methods:
```cpp
 block.settings().store_defaults();
 // ...
 block.settings().reset_defaults();
 // reverting to old settings
 // block.reset() is optionally called
```

Refactored, introduced some concepts for the optional `settings_changed()` and `reset()` block call back functions, and tried to simplify the code where possible (<-> low hanging fruits). 

The function signature for these optional block functions are checked using `static_assert(...)`s in the settings implementation that -- if violated -- yield something like:

image
or
image

Next on the list would be the parameter range checks...

### DIFF
--- a/include/node.hpp
+++ b/include/node.hpp
@@ -442,9 +442,8 @@ public:
             _output_tags_changed = true;
         }
 
-        // TODO: expand on this init function:
-        //  * store initial setting -> needed for `reset()` call
-        //  * ...
+        // store default settings -> can be recovered with 'reset_defaults()'
+        settings().store_defaults();
     }
 
     void
--- a/include/settings.hpp
+++ b/include/settings.hpp
@@ -21,6 +21,31 @@ struct SettingsCtx {
     property_map             context;             /// user-defined multiplexing context for which the setting is valid
 };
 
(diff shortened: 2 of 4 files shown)


wirew0rm (approved):
Looks good and can be merged once the conflicts are resolved by rebasing :+1: 

One small thing that could be interesting to add to the unit test and which i think is not covered currently are the `STORE_DEFAULTS` and `RESET_DEFAULTS` tags triggering the resets. But this would probably warrant a separate integration test to check all the different tag handling usecases that can be done in a separate PR.

PR #167, opened 2023-09-14
Enhance `Annotated<...>` with Range Limits and Generic Validation via `Limits<Min, Max, OptValidatorFunc>`

RalphSteinhagen:
This PR enhances the functionality of the `Annotated<...>` type by introducing range limits and generic validation checks. These are made possible via the optional `Limits<Min, Max, OptValidatorFunc>` type, where the minimum, maximum, and validator are specified as Non-Type Template Parameters (NTTPs).

### Examples

Here's how you can utilize the new feature:

1. **Basic Limit Specification**:
```cpp
Annotated<float, "sample rate", Limits<int64_t(0), std::numeric_limits<int64_t>::max()>> sample_rate = 1000.0f;
```

2. **Using a Separate Validator Function**:
```cpp
constexpr auto isPowerOfTwo = [](const int &val) { return val > 0 && (val & (val - 1)) == 0; };
Annotated<int, "power of two", Limits<0, 0, isPowerOfTwo>> needPowerOfTwo = 2;
```
  N.B. `MIN==MAX` disable the range check leaving the optional arbitrary validator function.

3. **Incorporating an Inline Lambda Validator**:
```cpp
Annotated<int, "power of two", Limits<0, 0, [](const int &val) { return val > 0 && (val & (val - 1)) == 0; }>> needPowerOfTwoAlt = 2;
expect(needPowerOfTwoAlt.validate(4));
expect(not needPowerOfTwoAlt.validate(5));
```

4. **Validation In Action**:
```cpp
block.settings().set({ { "sample_rate", -1.0f } });
block.settings().apply_staged_parameters(); // This currently prints a warning. TODO: Replace with a more informative PMT error message sent via the msgOut port.
// example error output:
// cannot set field fair::graph::setting_test::TestBlock<float>#24(fair::graph::setting_test::TestBlock<float>)::1 = -1 to sample_rate due to limit constraints [0, 9223372036854775807] validate func is not defined
```

### Future Work

The exact format of the PMT error message, which will be sent through the `MSG_OUT msgOut` port, is still under discussion. To be discussed during one of the upcoming GR Architecture WG meetings (@mormj et al.).

### DIFF
--- a/include/annotated.hpp
+++ b/include/annotated.hpp
@@ -3,6 +3,7 @@
 
 #include <string_view>
 #include <type_traits>
+#include <utility>
 #include <utils.hpp>
 
 namespace fair::graph {
@@ -93,6 +94,73 @@ static_assert(fair::meta::is_instantiation_of<DefaultSupportedTypes, SupportedTy
 static_assert(fair::meta::is_instantiation_of<SupportedTypes<float, double>, SupportedTypes>);
 
 /**
+ * @brief Represents limits and optional validation for an Annotated<..> type.
+ *
+ * The `Limits` structure defines lower and upper bounds for a value of type `T`.
+ * Additionally, it allows for an optional custom validation function to be provided.
+ * This function should take a value of type `T` and return a `bool`, indicating
+ * whether the value passes the custom validation or not.
+ *
+ * Example:
+ * ```
+ * Annotated<float, "example float", Visible, Limits<0.f, 1024.f>>             exampleVar1;
+ * // or:
+ * constexpr auto isPowerOfTwo = [](const int &val) { return val > 0 && (val & (val - 1)) == 0; };
+ * Annotated<float, "example float", Visible, Limits<0.f, 1024.f, isPowerOfTwo>> exampleVar2;
+ * // or:
+ * Annotated<float, "example float", Visible, Limits<0.f, 1024.f, [](const int &val) { return val > 0 && (val & (val - 1)) == 0; }>> exampleVar2;
+ * ```
+ */
+template<auto LowerLimit, decltype(LowerLimit) UpperLimit, auto Validator = nullptr>
+    requires(requires(decltype(Validator) f, decltype(LowerLimit) v) {
+        { f(v) } -> std::same_as<bool>;
+    } || Validator == nullptr)
+struct Limits {
+    using ValueType                                    = decltype(LowerLimit);
+    static constexpr ValueType           MinRange      = LowerLimit;
+    static constexpr ValueType           MaxRange      = UpperLimit;
+    static constexpr decltype(Validator) ValidatorFunc = Validator;
+
+    static constexpr bool
+    validate(const ValueType &value) noexcept {
+        if constexpr (LowerLimit == UpperLimit) { // ignore range checks
+            if constexpr (Validator != nullptr) {
+                try {
+                    return Validator(value);
+                } catch (...) {
+                    return false;
+                }
+            } else {
+                return true; // if no validator and limits are same, return true by default
+            }
+        }
+        if constexpr (Validator != nullptr) {
+            try {
+                return value >= LowerLimit && value <= UpperLimit && Validator(value);
+            } catch (...) {
+                return false;
+            }
+        } else {
+            return value >= LowerLimit && value <= UpperLimit;
+        }
+        return true;
+    }
+};
+
+template<typename T>
+struct is_limits : std::false_type {};
+
+template<auto LowerLimit, decltype(LowerLimit) UpperLimit, auto Validator>
+struct is_limits<Limits<LowerLimit, UpperLimit, Validator>> : std::true_type {};
+
+template<typename T>
+concept Limit    = is_limits<T>::value;
+
+using EmptyLimit = Limits<0, 0>; // nomen-est-omen
+
+static_assert(Limit<EmptyLimit>);
+
+/**
  * @brief Annotated is a template class that acts as a transparent wrapper around another type.
  * It allows adding additional meta-information to a type, such as documentation, unit, and visibility.
  * The meta-information is supplied as template parameters.
@@ -100,6 +168,7 @@ static_assert(fair::meta::is_instantiation_of<SupportedTypes<float, double>, Sup
 template<typename T, fair::meta::fixed_string description_ = "", typename... Arguments>
 struct Annotated {
     using value_type = T;
+    using LimitType  = typename fair::meta::typelist<Arguments...>::template find_or_default<is_limits, EmptyLimit>;
     T value;
 
     Annotated() = default;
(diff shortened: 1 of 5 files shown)


include/annotated.hpp
```diff
@@ -92,6 +93,74 @@ using DefaultSupportedTypes = SupportedTypes<>;
...
+    } || Validator == nullptr)
+struct Limits {
+    using ValueType = decltype(LowerLimit);
+    static constexpr ValueType MinRange = LowerLimit;
+    static constexpr ValueType MaxRange = UpperLimit;
+    static constexpr decltype(Validator) ValidatorFunc = Validator;
+
+    static constexpr bool
+    validate(const ValueType &value) noexcept {
+        if (LowerLimit == UpperLimit) {
```
ivan-cukic, line 126: If the limits are equal, they are not applied? Should the value be checked `value == LoweLimit`?

include/annotated.hpp
```diff
@@ -153,6 +223,23 @@ struct Annotated {
         return *this;
     }
 
+    template<typename U>
+        requires std::is_same_v<std::remove_cvref_t<U>, T>
+    constexpr bool
+    validate(U &&value_) {
```
ivan-cukic, line 229: I'd propose renaming to `validate_and_set` or similar.

include/node.hpp
```diff
@@ -598,7 +605,7 @@ struct node : protected std::tuple<Arguments...> {
 
     constexpr void
     forward_tags() noexcept {
-        if (!_output_tags_changed) {
+        if (!_output_tags_changed && !_input_tags_present) {
```
ivan-cukic, line 608: Wasn't this (and some other parts) a part of 166?

include/annotated.hpp
RalphSteinhagen: yes, the limits are ignored if lower equals the upper limit.

include/annotated.hpp
RalphSteinhagen: OK.

include/node.hpp
RalphSteinhagen: Yes, this is the third PR in a row. The built upon each other but are functionally not directly related.

include/annotated.hpp
RalphSteinhagen: The most common case is that parameters are checked for min/max but while doing the unit-tests I wanted to add also the possibility to have only a validator function.

RalphSteinhagen:
need to push some clang fixes (missing 'template' pre-statement).

include/annotated.hpp
wirew0rm: I agree that it's a bit suprising to have min == max accept all values instead of the one exact value, but on the other hand having a setting that only accepts a single value is kind of pointless.

The more natural choice imo would be something like
``` c++
template <typename T>
using EmptyLimits = Limits<std::numeric_limits<T>::min, std::numeric_limits<T>::max>;
```
though I'm not sure if this would cause other problems.

I'd be ok with the currently implemented choice, just has to be documented accordingly.

include/annotated.hpp
```diff
@@ -92,6 +93,74 @@ using DefaultSupportedTypes = SupportedTypes<>;
...
+                } catch (...) {
+                    return false;
+                }
+            } else {
+                if (value < LowerLimit || value > UpperLimit) return false;
+            }
+            return true;
+        }
+    }
+};
```
wirew0rm, line 149: Is there a reason for nesting the otherwise not related branches? The following should be identical and save some nesting and repetition.

``` c++
if (LowerLimit == UpperLimit) {
    if (value < LowerLimit || value > UpperLimit) return false;
}
if constexpr (Validator != nullptr) {
    try {
        if (!Validator(value)) return false;
    } catch (...) {
        return false;
    }
}
return true;
```

include/annotated.hpp
```diff
@@ -92,6 +93,74 @@ using DefaultSupportedTypes = SupportedTypes<>;
...
+            }
+        } else {
+            if constexpr (Validator != nullptr) {
+                try {
+                    if (value < LowerLimit || value > UpperLimit || !Validator(value)) return false;
+                } catch (...) {
+                    return false;
+                }
+            } else {
+                if (value < LowerLimit || value > UpperLimit) return false;
```
wirew0rm, line 144: Inclusive vs exclusive limits: I guess there's applications for both, I would have expected inclusive limits, e.g. I'd want to be able to set `gain = 0.0` while limiting it to [0.0, 1.0].

include/node.hpp
```diff
@@ -316,11 +316,11 @@ struct node : protected std::tuple<Arguments...> {
...
-    A<std::size_t, "numerator", Doc<"The top number of a fraction = numerator/denominator: decimation (fraction < 1), interpolation (fraction > 1), no effect (fraction = 1)">>      numerator   = 1_UZ;
-    A<std::size_t, "denominator", Doc<"The bottom number of a fraction = numerator/denominator: decimation (fraction < 1), interpolation (fraction > 1), no effect (fraction = 1)">> denominator = 1_UZ;
-    A<std::size_t, "stride", Doc<"Number of samples between two data processing: overlap (stride < N), skip (stride > N), undefined-default (stride = 0)">>                          stride      = 0_UZ;
-    std::size_t                                                                                                    stride_counter                                                                = 0_UZ;
-    const std::size_t                                                                                              unique_id   = _unique_id_counter++;
+    A<std::size_t, "numerator", Doc<"bottom number of input-to-output sample ratio: < 1 decimation, >1 interpolation, 1_ no effect">, Limits<1_UZ, std::size_t(-1)>>   numerator      = 1_UZ;
```
wirew0rm, line 319: Accidental change top->bottom?

wirew0rm (approved):
Looks good overall, I added some small suggestions/comments none of them major, either trivial to fix or not significant enough to block this, feel free to incorporate before merging.

include/annotated.hpp
RalphSteinhagen: Note the `constexpr-if' cascade ... your code above is runtime evaluated.

include/annotated.hpp
RalphSteinhagen: That's the behaviour right now. The function returns false if it does not meet ... e.g. LowerLimit == 0.0 let's value == 0.0 pass.

include/node.hpp
RalphSteinhagen: fixed.

include/annotated.hpp
RalphSteinhagen: ...pushed the wrong version ... sorry. fixed.

include/annotated.hpp
wirew0rm: ah yes, one inversion too much for me, you are of course right, sorry for the noise.

PR #168, opened 2023-09-18
Remove warnings, disable warnings from fftw3 library.

drslebedev:


CMakeLists.txt
```diff
@@ -65,12 +65,13 @@ target_include_directories(pmtv INTERFACE ${pmt_SOURCE_DIR}/include/)
 
 # FFTW3 is build 2 times for float and double precisions
 SET(FFTW_PREFIX ${CMAKE_CURRENT_BINARY_DIR}/fftw)
+SET(FFTW_CFLAGS "${CFLAGS} -w")
 if (EMSCRIPTEN)
     SET(FFTW_CONFIG cd ${FFTW_PREFIX}/src/ && emconfigure ./configure --quiet --disable-fortran --prefix=${FFTW_PREFIX}/install)
-    SET(FFTW_BUILD emmake make -j && emmake make install && emmake make clean)
+    SET(FFTW_BUILD emmake make -j CFLAGS=${FFTW_CFLAGS} && emmake make install && emmake make clean)
```
wirew0rm, line 71: Looking at the build output I think we might also want to pass `--silent` to make to suppress the echo of the compile commands it is running? Right now the CI build step mostly shows the output of make for fftw (and that even two times).

wirew0rm (approved):
Looks good, thanks for cleaning up the build. I added 2 suggestions, feel free to incorporate if you agree and merge this.

test/qa_fft.cpp
```diff
@@ -99,8 +99,7 @@ const boost::ut::suite fftTests = [] {
             std::ignore = fft1.settings().set({ { "outputInDb", t.outputInDb } });
             std::ignore = fft1.settings().apply_staged_parameters();
             const auto signal{ generateSinSample<T>(t.N, t.sampleRate, t.frequency, t.amplitude) };
-
-            fft1.inputHistory.push_back_bulk(signal.begin(), signal.end());
+            for (const auto s : signal) fft1.inputHistory.push_back(static_cast<typename fft<T>::InHistoryType>(s));
```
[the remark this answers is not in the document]
drslebedev: Thanks for pointing this out. I included braces.

CMakeLists.txt
drslebedev: I included `--silent`. There is still some output form `fftw`, I'm not sure if we can control it.

PR #170, opened 2023-09-19
Fix typelist::at

ivan-cukic:


RalphSteinhagen (approved):
Ok. We should look eventually into a more generic algorithm that can handle an arbitrary number of `type_list<...>` parameter. I seem to recall a talk where Daisy Hollman illustrated such an algorithm...

RalphSteinhagen:
> With this patch, I can declare e.g. 16 ports (next feature request: arrays :) ) and it compiles.

@frankosterfeld you should be able to do so even before with field defined ports...

PR #171, opened 2023-09-21
refactored port<T> interface

RalphSteinhagen:
... as outlined by GR Architecture WG and https://github.com/fair-acc/graph-prototype/issues/148

tackled items:
 * refactored port structure (mandatory enum NTTPs vs. optional type-wrapped arguments)
     ```cpp
    class template<typename T, fixed_string PortName, port_type_t PortType, port_direction_t PortDirection, 
                                typename...  Arguments>
    struct Port { /* ... */ };
    ```
 * added optional domain argument
 * added default init value (needed for cyclic graphs)
 * add isOptional() annotation
 * fixed repeated_port name -> name0, name1, name2, ...
 * fixed `0 -> 1` for the required min port samples in various tests & code examples (now ensured via a `static_assert(..)`)
 * added 'Async' port annotation and 'isSynchronous()' function
 * renamed IN,OUT,... short-hand aliases to more explicit/hopefully descriptive PortIn, PortOut names
     ```cpp
    template<typename T, typename... Arguments>
    using PortIn = Port<T, "", port_type_t::STREAM, port_direction_t::INPUT, Arguments...>;
    template<typename T, typename... Arguments>
    using PortOut = Port<T, "", port_type_t::STREAM, port_direction_t::OUTPUT, Arguments...>;
    template<typename... Arguments>
    using MsgPortIn = Port<property_map, "", port_type_t::MESSAGE, port_direction_t::INPUT, Arguments...>;
    template<typename... Arguments>
    using MsgPortOut = Port<property_map, "", port_type_t::MESSAGE, port_direction_t::OUTPUT, Arguments...>;
    // ...
    ```
 * updated the developer [Readme.md](https://github.com/fair-acc/graph-prototype/tree/missing_port_interfaces/include#user-content-ports) and code documentation where necessary (-> there's more work for technical writers):
     Some of the possible optional port annotation attributes are:
     * `RequiredSamples<size_t, size_t>` to describe the min/max number of samples required from this port before invoking the blocks work function,
     * `Optional` informing the graph/scheduler that a given port does not require to be connected,
     * `PortDomain<fixed_string>` described whether the port can be handled within the same scheduling domain (e.g. `CPU` or `GPU`),
     * `StreamBufferType` and `TagBufferType` to inject specific user-provided buffer implementations to the port, or
     * `Async` for making a port asynchronous in a signal flow-graph block.
 * changed to capitalised class naming following [C++ Core guideline item](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#example-389) and Bjarne Stroustrup's [style naming](https://www.stroustrup.com/Programming/PPP-style.pdf) convention where ISO-C++/STL-style classes (or in our case classes that could be) are lower-case but own declared types being declared capitalised. N.B. this helps to distinguish whether classes are standard ISO-C++ (and should perhaps be known by the user) and/or are defined in the GNU Radio context only.
 
 Thanks for the in-person and on-line pre-comments and feedbacks by @ivan-cukic, @drslebedev, and @wirew0rm.
 
 One of the follow-ups would be to implement the `Async` function in the node/block code.

### DIFF
--- a/include/README.md
+++ b/include/README.md
@@ -19,10 +19,10 @@ continuously improve this document.
   through a general `work()` function. Blocks are the building blocks of a flow-graph and can be thought of as vertices
   in a graph, and *ports* are their input/output connections to neighboring blocks for data streaming, streaming tags,
   and asynchronous messages. For the specific implementation, see [node.hpp](node.hpp).
-* [port](#Ports) is an interface through which data flows into or out of a block. Each block may have zero, one or
-  more input ports, and zero, one or more output ports. Data is passed between blocks by connecting the output port of
-  one block to the input port of another. For the specific implementation, see [port.hpp](port.hpp).
-* [buffer](#Buffer)  is an area of memory where data is temporarily stored in the runtime-connected graph. Each port
+* [Port](#Ports) is an interface through which data flows into or out of a block. Each block may have zero, one or
+  more input ports, and zero, one or more output ports. Data is passed between blocks by connecting the output Port of
+  one block to the input Port of another. For the specific implementation, see [port.hpp](port.hpp).
+* [buffer](#Buffer)  is an area of memory where data is temporarily stored in the runtime-connected graph. Each Port
   has its own buffer to store the data, tags, or other messages it needs to perform its computations. Buffer
   implementations are typically domain-specific (e.g. for blocks/ports implemented on the CPU, GPU, etc.) and are often,
   but not necessarily, implemented as circular buffers. For the specific interface see [Buffer.hpp](Buffer.hpp) and one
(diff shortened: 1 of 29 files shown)


ivan-cukic (approved):
Looks good

include/utils.hpp
```diff
@@ -104,6 +104,44 @@ operator+(const fixed_string<CharT, N1> &lhs, const fixed_string<CharT, N2> &rhs
...
+constexpr int
+log10(int n) noexcept {
+    if (n < 10) return 0;
+    return 1 + log10(n / 10);
+}
+
+constexpr int
+pow10(int n) noexcept {
+    if (n == 0) return 1;
+    return 10 * pow10(n - 1);
```
wirew0rm, line 117: Should these math helper functions fall back to a non recursive implementation if they are evaluated at runtime? Or be `consteval` or an assertion to prevent accidental runtime use?

include/port.hpp
```diff
@@ -346,39 +504,52 @@ namespace detail {
...
+using PortIn = Port<T, "", port_type_t::STREAM, port_direction_t::INPUT, Arguments...>;
+template<typename T, typename... Arguments>
+using PortOut = Port<T, "", port_type_t::STREAM, port_direction_t::OUTPUT, Arguments...>;
+template<typename... Arguments>
+using MsgPortIn = Port<property_map, "", port_type_t::MESSAGE, port_direction_t::INPUT, Arguments...>;
+template<typename... Arguments>
+using MsgPortOut = Port<property_map, "", port_type_t::MESSAGE, port_direction_t::OUTPUT, Arguments...>;
+
+template<typename T, fixed_string PortName = "", typename... Arguments>
+using PortInNamed = Port<T, PortName, port_type_t::STREAM, port_direction_t::INPUT, Arguments...>;
```
wirew0rm, line 534: Why is `PortInNamed` explicitly necessary and not just have the defaulted argument for the name on `PortIn`?
Probably because it would prevent having a string as the first Argument?

In any case, since the only purpose of this Template in comparison to `PortIn`, it should probably not be defaulted to the empty string?

include/README.md
```diff
@@ -19,10 +19,10 @@ continuously improve this document.
   through a general `work()` function. Blocks are the building blocks of a flow-graph and can be thought of as vertices
   in a graph, and *ports* are their input/output connections to neighboring blocks for data streaming, streaming tags,
   and asynchronous messages. For the specific implementation, see [node.hpp](node.hpp).
-* [port](#Ports) is an interface through which data flows into or out of a block. Each block may have zero, one or
-  more input ports, and zero, one or more output ports. Data is passed between blocks by connecting the output port of
-  one block to the input port of another. For the specific implementation, see [port.hpp](port.hpp).
-* [buffer](#Buffer)  is an area of memory where data is temporarily stored in the runtime-connected graph. Each port
+* [Port](#Ports) is an interface through which data flows into or out of a block. Each block may have zero, one or
```
wirew0rm, line 22: I assume Port is capitalized because you refer to the actual type and not the general concept? I it is supposed to be code I would expect it to be surrounded by single ticks (`Port`). Not super important but if we decide on one style there we can make the effort to keep it consistent everywhere.

wirew0rm (approved):
Looks good and makes the port definitions nicer. Thanks also for taking care of updating the documentation :+1:

PR #175, opened 2023-09-29
Fix test fails, switch off fsanitize=address for emscriptem.

drslebedev:
Fix test fails. It seems that `-fsanitize=address` flag causes that too many locals are created. 
For the moment just do not use `-fsanitize=address` flag for emscriptem.


RalphSteinhagen (approved):
Normally disabling warnings/sanitizers would be a workaround but since this is on for the other gcc/clang cases this is OK. Thanks @drslebedev for looking into this. 😀

PR #177, opened 2023-10-02
updated FFT block and helper functions

RalphSteinhagen:
* reduced the API towards a more clean and lean implementation
  - reduced 'using' declarations
  - removed redundant functions in the block implementation
* changed fftSize from problematic/non-portable `std::size_t` to `std::uint32_t`
* fixed sample_rate, and signal_... names to follow the tag::DEFAULT_TAGS conventions
* added Kaiser window function (pre-requisite for FIR filter calculations)
* updated block documentation

Future Todos:
 * semi-private caching vectors (need to be public for unit-test) should be moved to FFT implementations, notably casting from T -> U::value_type should be done there
 * use of SIMD primitives where possible
 * change window function from int to std::string (parsing + default fall back) function

### DIFF
--- a/bench/bm_fft.cpp
+++ b/bench/bm_fft.cpp
@@ -55,11 +55,12 @@ template<typename T>
 void
 testFFT() {
     using namespace benchmark;
+    using namespace boost::ut;
     using namespace boost::ut::reflection;
     using namespace fair::graph;
     using namespace gr::algorithm;
 
-    constexpr std::size_t N{ 1024 }; // must be power of 2
+    constexpr std::uint32_t N{ 1024U }; // must be power of 2
     constexpr double      sampleRate{ 256. };
     constexpr double      frequency{ 100. };
     constexpr double      amplitude{ 1. };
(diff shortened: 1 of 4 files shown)


ivan-cukic (approved):

bench/bm_fft.cpp
```diff
@@ -75,45 +76,28 @@ testFFT() {
...
-
-        double sum{ 0. };
-        fft1.computeMagnitudeSpectrum();
-        fft1.computePhaseSpectrum();
-        ::benchmark::benchmark<nRepetitions>(fmt::format("{} - dataset", type_name<T>())) = [&fft1, &sum] {
-            const auto ds1 = fft1.createDataset();
-            sum += static_cast<double>(ds1.signal_values[0]);
-        };
-    }
-
```
drslebedev, line 116: Why don't you want to keep this performance measurements? At least as a reminder that calculation of phase and magnitude can be also at the same order as FFT itself?

drslebedev (approved):
The code looks good to me! Very nice work in refactoring FFT, everything seems clear.

bench/bm_fft.cpp
RalphSteinhagen: From a clean and lean code perspective, my decision was based on the following considerations:

1. The tests don't provide substantial additional value to the ones above, particularly as we're benchmarking our FFT against FFTW. Nuances such as magnitude and phase calculations might be better placed in a dedicated test suite.
2. The granular breakdown, often into one-liner methods, makes the FFT GR block's core functionality harder to read for new developers or those unfamiliar with the intricacies of FFT. Or expressed differently, since this is supposed to be a reference implementation/benchmark, what would the library become for hundreds of more blocks that we need to maintain if we maintain this high verbosity?
3. The current tests don't compare various implementations.
4. If the phase and magnitude calculations are potential performance bottlenecks, they should be benchmarked as stand-alone functions.
5. In striving for clean and lean code, it's vital that our benchmarks reflect the primary use cases and core functionalities of the components we're testing. Introducing benchmarks for subsidiary operations could detract from this primary focus.

I just wanted to let you know that your existing tests are good. At the same time, our tests must be clear, targeted, and underscore primary functionalities. Sometimes, *less is more*.

drslebedev: Thanks a lot for a very detailed answer!

PR #179, opened 2023-10-04
added window function to_string & parse API

RalphSteinhagen:
... and corresponding unit-tests. This is necessary as we cannot safely transmit C++ enum values across different domains, platforms and languages (e.g. C++, Python, or Java)


wirew0rm (approved):
Looks good!

PR #183, opened 2023-10-07
RequiredSamples, ResamplingRatio, Stride, and other improvements

RalphSteinhagen:
## RequiredSamples, ResamplingRatio, Stride, and other improvements

* generalised Annotated wrapper constructor and assignment operators through perfect forwarding and
  * std::constructible_from
  * std::assignable_from
* extended RequiredSamples to support explicit compile-time const limits
* renamed & extended PerformDecimationInterpolation to default (optionally compile-constant) ResamplingRatio
   ```cpp
   template<std::size_t numerator = 1LU, std::size_t denominator = 1LU, bool isConst = false>
   struct ResamplingRatio { /*..*/ };
   ```
* renamed & extended PerformStride to default (optionally compile-constant) Stride
   ```cpp
   template<std::size_t stride = 0LU, bool isConst = false>
   struct Stride { /*..*/ };
   ```
* added support for reflectable const setting types e.g. `const size_t`, `Annotated<const size_t,...>`
* addressed float/double-conversion compiler warnings in window function unit tests

**Basic example**:
```cpp
struct TestBlock : gr::node<TestBlock, ResamplingRatio<2LU, 1LU, true>, Stride<2LU, false>> {} testBlock;

static_assert(std::is_const_v<decltype(testBlock.numerator.value)>); 
static_assert(std::is_const_v<decltype(testBlock.denominator.value)>);
static_assert(!std::is_const_v<decltype(testBlock.stride.value)>);

expect(eq(testBlock.numerator, 2LU));
expect(eq(testBlock.denominator, 1LU));
expect(eq(testBlock.stride, 2LU));
```


wirew0rm (approved):
:+1:

Added one small change in a separate commit to fix a potential problem that sonarcloud found with the annotated wrapper constructor being too greedy. Feel free to merge if the fix seems valid or drop it if this was intentional or does not correctly resolve the issue.

PR #186, opened 2023-10-09
Frank/settings vector

frankosterfeld:
Don't print warning about std::vector<std::string> being unsupported and add it to test.
Minor code simplifications.

### DIFF
--- a/core/include/gnuradio-4.0/settings.hpp
+++ b/core/include/gnuradio-4.0/settings.hpp
@@ -269,7 +269,7 @@ public:
                         auto iterate_over_member = [&](auto member) {
                             using RawType = std::remove_cvref_t<decltype(member(*_node))>;
                             using Type    = unwrap_if_wrapped_t<RawType>;
-                            if constexpr ((!traits::node::detail::is_port_or_collection<Type>()) && is_writable(member) && (std::is_arithmetic_v<Type> || std::is_same_v<Type, std::string> || gr::meta::vector_type<Type>) ) {
+                            if constexpr (!traits::node::detail::is_port_or_collection<Type>() && is_writable(member) && is_supported_type<Type>()) {
                                 auto matchesIgnoringPrefix = [](std::string_view str, std::string_view prefix, std::string_view target) {
                                     if (str.starts_with(prefix)) {
                                         str.remove_prefix(prefix.size());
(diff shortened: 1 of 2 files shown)


core/include/gnuradio-4.0/settings.hpp
```diff
@@ -269,7 +269,7 @@ class basic_settings : public settings_base {
                         auto iterate_over_member = [&](auto member) {
                             using RawType = std::remove_cvref_t<decltype(member(*_node))>;
                             using Type    = unwrap_if_wrapped_t<RawType>;
-                            if constexpr ((!traits::node::detail::is_port_or_collection<Type>()) && is_writable(member) && (std::is_arithmetic_v<Type> || std::is_same_v<Type, std::string> || gr::meta::vector_type<Type>) ) {
```
RalphSteinhagen, line 272: The original restrictive list was due to the fact that `is_supported_type` supports way more than just strings, integral, floating point, and vectors thereof but also maps etc. 

Opening the settings that wide may need some additional unit-tests and/or we would have holes in the implementation that fail the first time when someone sets -- for example -- a map.

RalphSteinhagen (approved):
LGTM! :+1: 

*A minor comment: @frankosterfeld you opened the settings interface to all pmt types not just `std::vector<std::strings>`. Some cases like `map<std::string, pmt_t>` and recursions are enabled and untested now.`

core/include/gnuradio-4.0/settings.hpp
frankosterfeld: Hmm, it wasn't the idea to change anything about the supported types.

I replaced `std::is_arithmetic_v<Type> || std::is_same_v<Type, std::string> || gr::meta::vector_type<Type>` here by is_supported_types(), which is `std::is_arithmetic_v<Type> || std::is_same_v<Type, std::string> || gr::meta::vector_type<Type>`, so it should be equivalent? The only change in did in there is to shorten `std::integral<Type> || std::floating_point<Type>` to `std::is_arithmetic_v<Type>`, which is equivalent, if I read the docs correctly.
Do I miss something?

(Actually this change originates in me trying to add property_map support, which I then abandoned when going for the individual channel_* vectors for the channel settings)

RalphSteinhagen: The 'supported types' definition is actually from pmt which we copied/simplified but needs to be rolled back to pmt.

Will merge this as is because it solves your specific problem/application.

PR #189, opened 2023-10-10
FFT block and algorithms refactoring

drslebedev:
FFT block and algorithms refactoring:

* Change signature of the FFT algorithm classes. Add `TInput` and `TOutput` template parameters
* Precision of the calculation is defined by the `TOutput` precision
* Clean up FFT unit tests


algorithm/include/gnuradio-4.0/algorithm/fourier/fft_common.hpp
```diff
@@ -6,33 +6,74 @@
...
+computeMagnitudeSpectrum(const TContainerIn &fftIn, ConfigMagnitude config) {
+    return computeMagnitudeSpectrum(fftIn, {}, config);
 }
 
-template<ComplexType T, typename U>
-void
-computePhaseSpectrum(const std::vector<T> &fftOut, std::vector<U> &phaseSpectrum) {
-    if (fftOut.size() < phaseSpectrum.size()) {
-        throw std::invalid_argument(fmt::format("FFT vector size ({}) must be more or equal than phase vector size ({}).", fftOut.size(), phaseSpectrum.size()));
+struct ConfigPhase {
```
RalphSteinhagen, line 50: I was initially hesitant when seeing an additional config type to `ConfigMagnitude` but the `ConfigPhase` definition here is warranted, since the `outputInDb` wouldn't make sense here. As a follow-up thought, what would make a lot of sense sense is to further specify:
 1. outputInDeg (default: false)
 2. unwrapPhase (default: false)
 
 As these are widespread post-processing options in that domain. Regarding naming: I am undecided whether specifying the prefix `outputXXX` is really necessary for brevity and context reasons, but I am OK if it is left as is.
 
**As a long-term goal*: we should look into, introduce and use physical unit safety. The [mp-units (v2)](https://github.com/mpusz/mp-units) is a very fine library, also for learning C++ meta-programming at its best. 
However, there are unresolved questions and challenges, particularly when it comes to:
 * interoperability w.r.t. runtime polymorphism (especially regarding type multiplicity because of additional quantity & unit NTTPs combinations)
 * interoperability w.r.t. complex data types (`std::simd`, `std::complex`, linear algebra, ...)

In my opinion, this would help constrain many of our block parameters and make GNU Radio in its various applications much safer to use also in industrial/critical applications.
Maybe a  leisurely discussion to be had for whether this is something for gnuradio-4.0 or rather 4.1. Maybe this could be done minimally invasive, with end-users do not having to pay a price for this. I think there is still a bit time as this would would need to be finalised only with the beta-release which I hope we can give to the wider GR community late spring (N.B. 'alpha' being what we want to push for the GR Architecture team and GR enthusiasts this autumn). Having this feature in-built could attract many commercial, and public-sponsored users and show-case GR & C++ at its best. For the time being, food for thought ...

@drslebedev, @wirew0rm, @ivan-cukic, @mattkretz, @mormj @jsallay

algorithm/include/gnuradio-4.0/algorithm/fourier/fft_types.hpp
```diff
@@ -1,28 +1,12 @@
 #ifndef GNURADIO_ALGORITHM_FFT_TYPES_HPP
 #define GNURADIO_ALGORITHM_FFT_TYPES_HPP
 
-#include <type_traits>
 #include <complex>
+#include <type_traits>
```
RalphSteinhagen, line 5: Is this include (nearly) empty? If yes, please remove...

RalphSteinhagen (approved):
This looks very nice, I made only minor comments, but it could be merged as is. :+1: 

One of the things reminded me and triggered some thoughts that we may pick up in a larger group w.r.t. feasibility etc.

algorithm/include/gnuradio-4.0/algorithm/fourier/fft_types.hpp
drslebedev: Basically there is only one concept definition inside for complex numbers:
```cpp
concept ComplexType = std::is_same_v<T, std::complex<float>> || std::is_same_v<T, std::complex<double>>;
```

But it is widely in use in FFT classes. 
Where can I move it? What would you propose?

PR #191, opened 2023-10-11
fixes bugs reported by sonar

RalphSteinhagen:
* missing 'requires'
* sloppy '-1_UZ' cast
* erroneously use of std::move


wirew0rm (approved):
Looks good, thanks for following these up! :+1:

PR #192, opened 2023-10-11
Updated syntax for connections, for ports in collections as well

ivan-cukic:
- Removed the connect syntax with pointers to members
- Added the support for vectors and arrays of ports to the static connect syntax
- Exposing static port names to dynamic ports
- Added syntax for dynamic_connect that works with port names, with the support for port collections

### DIFF
--- a/core/include/gnuradio-4.0/Block.hpp
+++ b/core/include/gnuradio-4.0/Block.hpp
@@ -480,6 +480,33 @@ public:
     init(std::shared_ptr<gr::Sequence> progress_, std::shared_ptr<gr::thread_pool::BasicThreadPool> ioThreadPool_) {
         progress     = std::move(progress_);
         ioThreadPool = std::move(ioThreadPool_);
+
+        // Set names of port member variables
+        // TODO: Refactor the library not to assign names to ports. The
+        // block and the graph are the only things that need the port name
+        auto setPortName = [&]([[maybe_unused]] std::size_t index, auto &&t) {
+            using CurrentPortType = std::remove_cvref_t<decltype(t)>;
+            if constexpr (traits::port::is_port_v<CurrentPortType>) {
+                using PortDescriptor = typename CurrentPortType::ReflDescriptor;
+                if constexpr (refl::trait::is_descriptor_v<PortDescriptor>) {
+                    auto &port = (self().*(PortDescriptor::pointer));
+                    port.name  = CurrentPortType::Name;
+                }
+            } else {
+                using PortCollectionDescriptor = typename CurrentPortType::value_type::ReflDescriptor;
+                if constexpr (refl::trait::is_descriptor_v<PortCollectionDescriptor>) {
+                    auto       &collection     = (self().*(PortCollectionDescriptor::pointer));
+                    std::string collectionName = refl::descriptor::get_name(PortCollectionDescriptor()).data;
+                    for (auto &port : collection) {
+                        port.name = collectionName;
+                    }
+                }
+            }
+        };
+        traits::block::input_ports<Derived>::template apply_func(setPortName);
+        traits::block::output_ports<Derived>::template apply_func(setPortName);
+
+        // Handle settings
         if (const auto forward_parameters = settings().applyStagedParameters(); !forward_parameters.empty()) {
             std::for_each(_tags_at_output.begin(), _tags_at_output.end(), [&forward_parameters](Tag &tag) {
                 for (const auto &[key, value] : forward_parameters) {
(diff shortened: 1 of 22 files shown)


include/node.hpp
```diff
@@ -473,6 +473,31 @@ struct node : protected std::tuple<Arguments...> {
     init(std::shared_ptr<gr::Sequence> progress_, std::shared_ptr<fair::thread_pool::BasicThreadPool> ioThreadPool_) {
         progress     = std::move(progress_);
         ioThreadPool = std::move(ioThreadPool_);
+
+        // Set names of port member variables
+        auto setPortName = [&]([[maybe_unused]] std::size_t index, auto &&t) {
```
RalphSteinhagen, line 478: Why do we need this? aren't the reflected names enough?

RalphSteinhagen (changes requested):
please rebase and change to the new naming rules where necessary.

bench/bm_case1.cpp
```diff
@@ -495,7 +495,7 @@ inline const boost::ut::suite _runtime_tests = [] {
         fg::graph flow_graph;
         auto     &src  = flow_graph.make_node<test::source<float>>(N_SAMPLES);
         auto     &sink = flow_graph.make_node<test::sink<float>>();
-        expect(eq(fg::connection_result_t::SUCCESS, flow_graph.connect<"out">(src).to<"in">(sink)));
+        flow_graph.connect<"out">(src).to<"in">(sink);
```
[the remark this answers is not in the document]
ivan-cukic: No, `connect` always succeeds as it is a lazy connection - it does nothing. The failure can happen later - when the connections are actually performed.  That part is done either by the scheduler or manually in some tests - and there, they are checked.

include/node.hpp
ivan-cukic: The port instance does not know its variable name. This is a bridge between the compile-time refl-name and the runtime name of the port.

include/graph.hpp
```diff
@@ -426,28 +527,58 @@ struct graph {
...
     }
 
-    template<std::size_t src_port_index, std::size_t dst_port_index, typename Source, typename SourcePort, typename Destination, typename DestinationPort>
-    [[nodiscard]] connection_result_t
-    connect_impl(Source &src_node_raw, SourcePort &source_port, Destination &dst_node_raw, DestinationPort &destination_port, std::size_t min_buffer_size = 65536, std::int32_t weight = 0,
-                 std::string_view name = "unnamed edge") {
-        static_assert(std::is_same_v<typename SourcePort::value_type, typename DestinationPort::value_type>, "The source port type needs to match the sink port type");
+    template<typename Node>
+    [[nodiscard]] std::size_t
+    dynamic_output_port_index(Node &node, std::string_view name) {
```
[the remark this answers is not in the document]
ivan-cukic: Was just wrapper that provided some extra checks. Removed all of these.

bench/bm_case1.cpp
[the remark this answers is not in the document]
RalphSteinhagen: We should/must fail early, e.g. if the ports are incompatible (in vs. in, wrong data type, wrong port type etc.).

include/node.hpp
RalphSteinhagen: There is already API for that in the Port. Either we need to use 
 * statically 'named' ports (NTTP Port API), 
 * names that are derived from their field names, or
 * names that are derived from their field names and their index. Here the implicit assumption is that e.g. the 2 port in an array of ports would be called "in1". This minimises ambiguity and crafting unconventional names.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -169,22 +170,22 @@ struct Result {
 
 template<typename T>
 concept BlockLike = requires(T t, std::size_t requested_work) {
-    { t.unique_name } -> std::same_as<const std::string &>;
-    { unwrap_if_wrapped_t<decltype(t.name)>{} } -> std::same_as<std::string>;
-    { unwrap_if_wrapped_t<decltype(t.meta_information)>{} } -> std::same_as<property_map>;
-    { t.description } noexcept -> std::same_as<const std::string_view &>;
+                        { t.unique_name } -> std::same_as<const std::string &>;
```
RalphSteinhagen, line 173: clang-format16/17 did this? :open_mouth:

RalphSteinhagen: Not an issue ... just a comment.

RalphSteinhagen (approved):
@ivan-cukic thanks for the changes. :+1:

PR #193, opened 2023-10-12
CI: configure sonarclound coverage paths

wirew0rm:
Define coverage paths s.t. tests and benchmarks are not counted towards general coverage and line counts, but still analyzed in the test scope. see: https://docs.sonarcloud.io/advanced-setup/analysis-scope/


RalphSteinhagen (approved):

RalphSteinhagen (approved):
@wirew0rm thanks for improving the CI and adding the excludes. Makes sense. :+1:

PR #195, opened 2023-10-13
CI: correctly set coverage paths

wirew0rm:
hopefully the last time.

Trying to set sonartype scopes for sources and tests to get correct coverage reporting: https://docs.sonarcloud.io/advanced-setup/analysis-scope/

The documented way of setting tests and sources where they are distributed around the tree breaks the exclusion of the benchmark which crashes the sonar analysis process, so this needs some convincing to get everything set up correctly.

Should only be merged once the gcc13|Debug CI has successfully passed.


RalphSteinhagen (approved):
Hey, this looks great and these changes are sensible choices. :+1: 

Let's see whether we can get the Meson-based build system up and going again soon. Maybe @mormj or other Meson experts might be able to help with that.

PR #205, opened 2023-10-20
Implement SignalSource block

drslebedev:
The 'SignalSource 'class generates various types of signal waveforms, including sine, cosine, square, constant, saw, and triangle signals.
Users can set parameters such as 'amplitude', 'frequency', 'offset,' and 'phase' for the desired waveform.
Note that not all parameters are supported for all signals.

The following signal shapes are supported (A = amplitude, f = frequency, P = phase, O = offset, t = time):

For the Square, Saw and Triangle t corresponds to phase adjusted time t = t + P / (pi2 * f)

* Sine
This waveform represents a smooth periodic oscillation. It is one of the most basic and pure forms of a periodic waveform.
formula: s(t) = A * sin(2 * pi * f * t + P) + O

* Cosine
Similar to the sine signal but shifted by pi/2 radians.
formula: s(t) = A * cos(2 * pi * f * t + P) + O

* Square
Represents a signal that alternates between two levels: -amplitude and amplitude.
If timeWithinPeriod < halfPeriod, y(t) = A + O else, y(t) = -A + O

* Constant
A signal that always returns a constant value, irrespective of time.
s(t) = A + O

* Saw (Sawtooth)
This waveform ramps upward and then sharply drops. It's called a sawtooth due to its resemblance to the profile of a saw blade.
s(t) = 2 * A * (t * f - floor(t * f + 0.5)) + O

* Triangle
This waveform linearly increases from -amplitude to amplitude in the first half of its period and then decreases back to -amplitude in the second half, forming a triangle shape.
s(t) = A * (4 * abs(t * f - floor(t * f + 0.75) + 0.25) - 1) + O

Note to reviewer: This block should be also tested  visually with plots.


RalphSteinhagen (approved):
Nice PR. :+1: 

Good as is. We'd need to tackle the merged 'ClockSource + SignalGenerator ' -> SignalSource once the message ports are integrated and the `Block<T>` CRTP is cleaned-up.

PR #206, opened 2023-10-23
Support port collections in GRC YAML   

frankosterfeld:
Make sure we can load/save connections involving port collections, i.e. serialize a connection from block1, port collection 0, sub-index 1 to block2, port collection 0, sub-index 2 as
    
```
 connections:
   - [block1, [0, 1], block2, [0, 2]]
```

### DIFF
--- a/core/include/gnuradio-4.0/Graph.hpp
+++ b/core/include/gnuradio-4.0/Graph.hpp
@@ -431,18 +431,26 @@ public:
     }
 };
 
+template<typename T>
+struct PortIndexDefinition {
+    T           topLevel;
+    std::size_t subIndex;
+
+    constexpr PortIndexDefinition(T _topLevel, std::size_t _subIndex = meta::invalid_index) : topLevel(std::move(_topLevel)), subIndex(_subIndex) {}
+};
+
 class edge {
 public: // TODO: consider making this private and to use accessors (that can be safely used by users)
     using PortDirection::INPUT;
     using PortDirection::OUTPUT;
-    BlockModel  *_sourceBlock;
-    BlockModel  *_destinationBlock;
-    std::size_t  _sourcePortIndex;
-    std::size_t  _destinationPortIndex;
-    std::size_t  _minBufferSize;
-    std::int32_t _weight;
-    std::string  _name; // custom edge name
-    bool         _connected;
+    BlockModel                      *_sourceBlock;
+    BlockModel                      *_destinationBlock;
+    PortIndexDefinition<std::size_t> _sourcePortDefinition;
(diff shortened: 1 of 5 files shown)


core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -431,39 +431,48 @@ class BlockWrapper : public BlockModel {
     }
 };
 
-class edge {
+template<typename T>
+struct PortIndexDefinition {
+    T           topLevel;
```
RalphSteinhagen, line 436: If you could change `T` -> `std::size_t` and remove the unnecessary templatisation and constructor.

core/test/app_grc.cpp
```diff
@@ -13,13 +13,45 @@
 
 namespace grg = gr;
 
+template<typename T>
+struct ArraySource : public grg::Block<ArraySource<T>> {
```
RalphSteinhagen, line 17: boy-scout-rule: could you also rename `grg` -> `gr`

core/test/app_grc.cpp
```diff
@@ -13,13 +13,45 @@
...
+
+template<typename T>
+struct ArraySink : public grg::Block<ArraySink<T>> {
+    std::array<grg::PortIn<T>, 2> inA;
+    std::array<grg::PortIn<T>, 2> inB;
+};
+
+ENABLE_REFLECTION_FOR_TEMPLATE(ArraySink, inA, inB);
+
 struct test_context {
```
RalphSteinhagen, line 32: boy-scout-rule: if you could refactor and change `test_context` -> `TestContext`

RalphSteinhagen (approved):
LGTM! :+1: 

Some minor 'boy-scout-rule' things. If you could address them and then merge. You have my OK.

core/test/app_grc.cpp
frankosterfeld: Done

core/test/app_grc.cpp
frankosterfeld: Done

core/include/gnuradio-4.0/Graph.hpp
frankosterfeld: PortIndexDefinition is used with T=std::size_t and T=std::string, as an argument to the connect() calls, so it's desirable from an API POV that the passed string/std::size_t parameters are implicitly converted to PortIndexDefinition. (thus the non-explicit ctor)

RalphSteinhagen: OK.

PR #213, opened 2023-10-31
Use vir-simd 0.3.0

mattkretz:
* `qa_thread_affinity` fails on my system

The goal of this PR is to remove the code from GR that exists in vir-simd (`meta::simdize -> vir::simdize`). `vir::simdize<tuple<>>` and `vir::simdize<void>` are ill-formed whereas `meta::simdize<tuple<>>` produced a `tuple<>` (and `void` for `void`). I still believe this is the correct behavior for `simdize<T>` (if it can't be SIMD-ified, don't act as if it worked). This required a few adjustments in the GR code.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -17,8 +17,8 @@
 Language: Cpp
 AccessModifierOffset: -4
 AlignAfterOpenBracket: Align
-AlignConsecutiveAssignments: AcrossEmptyLines
-AlignConsecutiveDeclarations: AcrossEmptyLines
+AlignConsecutiveAssignments: Consecutive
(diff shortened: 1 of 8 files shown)


mattkretz:
Oh, it looks like the vir-simd tests found a bug in Clang 17? I only tested vir-simd with Clang 15, 16, and latest Emscripten (which is Clang 18 based). Off to add Clang 17 to the vir-simd CI...

mattkretz:
I can't reproduce the clang 17 failures locally. (exact same clang-17 packages) Neither inside vir-simd nor in graph-prototype. :shrug: How do I go from here?

PR #214, opened 2023-10-31
Block derived classes can optionally implement lifecycle methods

drslebedev:
Block derived classes can optionally implement any subset of the lifecycle methods ( `start()`, `stop()`, `reset()`, `pause()`, `resume()`).

The Scheduler invokes these methods on each Block instance, if they are implemented, just before invoking its corresponding method of the same name.
```cpp
 struct user_defined_block : public Block<user_defined_block> {
 void start()  {...}
 void stop()  {...}
 void pause() {...}
 void resume() {...}
 void reset() {...}
 };
```


frankosterfeld:
Great, can't wait for this to land :)

I'll cherry-picked it to frank/opendigitizer-service and will give it a try.

The qa_DataSink.cpp test contains a lot of these lines:

    sink.stop(); // TODO the scheduler should call this`

These could be removed now and the test should still pass.

frankosterfeld:
For multithreaded schedulers, the functions should be called from the worker thread, otherwise each block needs to do the synchronization for itself.

frankosterfeld:
> Thanks a lot for your comments!
> 
> > For multithreaded schedulers, the functions should be called from the worker thread, otherwise each block needs to do the synchronization for itself.
> 
> The Scheduler invokes these function in a single thread, multi-threading is available only for `work` function.

Hmm, as start()/stop() etc. will often access the same data structures as the work/processBulk() functions, that adds an additional complexity when implementing a block. In the case of DataSink for example, stop() iterates over the list of active listeners, also used by processBulk(), so I would have to lock a mutex in every processBulk() call to avoid races.

frankosterfeld:
> > Thanks a lot for your comments!
> > > For multithreaded schedulers, the functions should be called from the worker thread, otherwise each block needs to do the synchronization for itself.
> > 
> > 
> > The Scheduler invokes these function in a single thread, multi-threading is available only for `work` function.
> 
> Hmm, as start()/stop() etc. will often access the same data structures as the work/processBulk() functions, that adds an additional complexity when implementing a block. In the case of DataSink for example, stop() iterates over the list of active listeners, also used by processBulk(), so I would have to lock a mutex in every processBulk() call to avoid races.

I think this could be avoided if stop()/pause() could be called in waitDone() after waiting for _runningThreads going down to 0. That would then also cover my wish for stop() being called on normal runAndWait() termination.

frankosterfeld:
> > > Thanks a lot for your comments!
> > > > For multithreaded schedulers, the functions should be called from the worker thread, otherwise each block needs to do the synchronization for itself.
> > > 
> > > 
> > > The Scheduler invokes these function in a single thread, multi-threading is available only for `work` function.
> > 
> > 
> > Hmm, as start()/stop() etc. will often access the same data structures as the work/processBulk() functions, that adds an additional complexity when implementing a block. In the case of DataSink for example, stop() iterates over the list of active listeners, also used by processBulk(), so I would have to lock a mutex in every processBulk() call to avoid races.
> 
> I think this could be avoided if stop()/pause() could be called in waitDone() after waiting for _runningThreads going down to 0. That would then also cover my wish for stop() being called on normal runAndWait() termination.

I pushed my suggested changes here: https://github.com/fair-acc/graph-prototype/commit/c876c0c93f778f784853f4e7a84ab95d5d57f3aa

reset() still wouldn't be thread-safe, and I'm unsure how resume() is supposed to work (as resume() on the scheduler itself does nothing??)

core/test/qa_HierBlock.cpp
```diff
@@ -72,6 +72,21 @@ class HierBlock : public gr::BlockModel {
...
+
+    void
+    pause() override {}
+
+    void
+    resume() override {}
+
+    void
+    reset() override {}
+
```
RalphSteinhagen, line 89: Are these empty 'override's needed? If not, please remove for conciseness.

blocks/basic/include/gnuradio-4.0/basic/common_blocks.hpp
```diff
@@ -113,6 +114,21 @@ class multi_adder : public gr::BlockModel {
...
+
+    void
+    pause() override {}
+
+    void
+    resume() override {}
+
+    void
+    reset() override {}
+
```
RalphSteinhagen, line 131: Empty overrides, are these needed?

RalphSteinhagen (approved):
Looks really nice and I especially like the documentation and more stream-lined scheduler state-machine handling! Well done!

There are some superfluous functions that could be removed but otherwise. OK.

drslebedev:
Thanks a lot for the review. Very good point! 
Basically it should be a user responsibility to check state of the `Scheduler`. The transitions between states are restricted and can be checked via `canXXX()` methods, they are public. Probably even in your case it is a good illustration of incorrect use of API and how it can be easily found during the first run, which I think is a good thing.

A misuse of the API can be still an issue: one can wait for a `Scheduler` state which is impossible.
If we are talking about UI the buttons responsible for "start", "stop" etc. can be enabled/disabled depending on Scheduler state, which van be checked with `canXXX()` methods.

I will merge it as it is and one can revisit later if needed.

PR #215, opened 2023-11-01
Set ColumnLimit back to 200.

drslebedev:
Adjusted the ColumnLimit setting back to 200 as discussed with @RalphSteinhagen to address certain formatting issues we encountered. 

We should discuss this further and probably conduct a more systematic review of the overall code formatting.


RalphSteinhagen (approved):
As discussed, need to revise this more carefully to address also the other side-effects.

PR #216, opened 2023-11-01
Frank/scheduler fixes

frankosterfeld:
- Fixes terminating of multithreaded running graphs via stop() when the graphs wouldn't terminate by itself.
- Add test demonstrating issues when using more than two threads:
When using the default threadpool, the worker threads of dependent
blocks might all report that nothing happened and increase the
progress/ "done" count (in poolWorker()), before the producing blocks
have any chance to produce anything. That makes the dependent blocks
"shutdown", while the producing node produces something and then sleeps
forever.
- Fix by @wirew0rm for above issue
- Small feature: Add isProcessing() to test if a multithreaded scheduler finished processing.

### DIFF
--- a/core/include/gnuradio-4.0/Scheduler.hpp
+++ b/core/include/gnuradio-4.0/Scheduler.hpp
@@ -191,6 +191,10 @@ public:
                     const profiling::Options &profiling_options = {})
         : SchedulerBase<TProfiler>(std::move(graph), thread_pool, profiling_options) {}
 
+    bool isProcessing() const requires (executionPolicy == multiThreaded) {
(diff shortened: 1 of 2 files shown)


wirew0rm:
Fixed the progress logic in the scheduler and rebased to main to apply the updated clang format

frankosterfeld:
Mostly obsolete now, I think the isProcessing() bit is still relevant, keeping it as reminder.

wirew0rm:
> Mostly obsolete now, I think the isProcessing() bit is still relevant, keeping it as reminder.

since `isProcessing()` was added in #260 I suppose this PR can be closed now.

PR #218, opened 2023-11-06
IIR and FIR coefficient generator and time-domain filter (<800 SLOCs!)

RalphSteinhagen:
## IIR and FIR coefficient generator tool and time-domain filter implementation

This PR introduces implementation for both Infinite Impulse Response (IIR) and Finite Impulse Response (FIR) filters within the `gr::filter` namespace. 

The primary focus lies in:
  * Provide a robust, clean, and lean C++-only implementation (<800 SLOCs!) that is easy(/ier) to maintain and can be used within `Block<T>`s to derive default FIR or IIR filter parameters based on high-level filter parameter requirements w/o strings attached (e.g. needing Python to compute multiplications of complex numbers).
  * consistent evaluation and creation of the required IIR and FIR filter coefficients, e.g.
    ```cpp
    constexpr auto kFilterParams = FilterParameters{ .order = 4UZ, .fLow = 4.0, .fHigh = 6.0, .attenuationDb = 60, .fs = 1000.0 }
    const auto digitalBandPass1 = iir::designFilter<double>(BANDPASS, kFilterParams, /* optional */ BUTTERWORTH);
    const auto digitalBandPass2 = fir::designFilter<double>(BANDPASS, kFilterParams, /* optional */ Kaiser);
    ```
 * application of filter coefficients for digital signal processing, e.g.
   ```cpp
   for (const auto &frequency : { 3.0, 3.5, 5., 6.5, 7.0 }) {
     auto filter = Filter<double>(digitalBandPass);
     std::vector<double> yValue(xValues.size());
     std::transform(xValues.cbegin(), xValues.cend(), yFiltered.begin(), [&filter](double x) { 
       const double 3.0 * std::sin(2. * std::numbers::pi * frequency * t)L
       return filter.processOne(y); });
      chart.draw(xValues, yFiltered, fmt::format("filtered@{}Hz", frequency));
   }
   ```
 * Automatic derivation of 'transition width' in FIR filters to match user expectations based on high-level filter parameter: the API distinction between IIR and FIR are deliberately kept to a minimum to share a common infrastructure and transparent switching between filter design and implementation.
   ```cpp
    struct FilterParameters {
      std::size_t order{ 4UZ };                                      /// default filter order
      double      fLow{ std::numeric_limits<double>::quiet_NaN() };  /// Lower cutoff frequency [Hertz].
      double      fHigh{ std::numeric_limits<double>::quiet_NaN() }; /// Upper cutoff frequency [Hertz].
      double      gain{ 1.0 };                                       /// required total filter gain
      double      rippleDb{ 0.1 };                                   /// Maximum allowed ripple in the pass-band [dB].
      double      attenuationDb{ 40 };                               /// Minimum required attenuation in the stop-band [dB].
      double      beta{ 1.6 };                                       /// default beta for Kaiser-type windowing
      double      fs{ std::numeric_limits<double>::quiet_NaN() };    /// Sampling frequency for digital filters [Hertz].
    };
    ```
   In order to minimise the level of surprise to users that express and have expectations w.r.t. the high-level filter parameter and their performance, the 'transition width' typical in other FIR filter implementations has been changed and been made an implicit parameter and is automatically derived to match the required:
    * filter `.order`, i.e. to achieve 20dB attenuation/decade per order,
    * filter `.fLow` and `.fHigh` parameters to have a correct corner frequency (especially in the vicinity of DC and the Nyquist frequency), and
    * filter stop-band attenuation `.attenuationDb`, which is particularly important for `.fLow` and `.fHigh` being in the vicinity of DC, the Nyquist frequency, or close to each other in case of band-stop filters.
   This should always provide safe defaults, the choice of least surprise, and implementation that requires the least amount of filter design expertise or having to inspect and evaluate each filter for a given parameter combination.
   *N.B The actual number of generated filter taps is still visible via `FilterCoefficients::b::size()`.*

### IIR Filter Implementation
- The IIR section provides an extensive toolkit to design various types of filters such as Butterworth, Bessel, Chebyshev Type I, and Chebyshev Type II.
- It includes methods for analog filter design (`designAnalogFilter(filter::Type, FilterParaneters, filter::Design)`), transformation to digital filters using Tustin's bilinear transform (`analogToDigitalTransform`), or direct digital filter genration (`designFilter<T>(filter::Type, FilterParaneters, filter::Design)`) that yield the desired `FilterCoefficients<T>`:
  ```cpp
  /**
   * @brief Filter coefficients of a digital transfer function H(z) = B(z)/A(z) in the z-domain with:
   *   B(z) = b[0] + b[1]·z^{-1} + b[2]·z^{-2} + …
   *   A(z) = a[0] + a[1]·z^{-1} + a[2]·z^{-2} + …
   *
   * The difference equation representing the filter is:
   * y[n] = b[0]·x[n] + b[1]·x[n-1] + … - (a[1]·y[n-1] + a[2]·y[n-2] + …)
   *
   * @note Typically, a[0] is 1 for causal systems and a{
   */
  template<typename T>
  struct FilterCoefficients {
    using value_type = T;
    std::vector<T> b{};                    /// numerator coefficients
    std::vector<T> a{ static_cast<T>(1) }; /// denominator coefficients
  };
  ```
- The IIR filters are tailored through `FilterParameters`, allowing for specific definitions of filter order, cutoff frequencies, and other critical parameters.
- There are facilities to inspect the analog poles and zeros
imageimage
- as well as the filter performance for the low-pass, high-pass, band-pass, and band-stop (here showing the analog and IIR filter being numerically virtually identical): 
 image
- or example to illustrate the IIR band-pass filter using various input frequencies:
image

### FIR Filter Implementation
- The FIR component offers a filter coefficient generator ((`designFilter<T>(filter::Type, FilterParaneters, window::Type)`)) based on the window method and that behaves very similar to the IIR filter. The default apodisation is the `Kaiser` window that can be further configured via the `FilterParameters::beta` parameter or replaced by any other of the window functions (e.g. `Rectangular`, `Hamming`, `Hann`, `HannExp`, `Blackman`, `Nuttall`, `BlackmanHarris`, `BlackmanNuttall`, `FlatTop`, `Exponential`, `Kaiser`).
- Example low-, high-, and band-pass filter (`N` indicates the number of required taps):
  image
  on can see that the FIR requires significantly more taps to achieve the same performance as the equivalent IIR filter (see IIR example above).
- Example band-stop filter:
 image
- or example to illustrating the FIR band-pass filter using various input frequencies:
 image

### IIR vs. FIR Filter Performance

IIR filters are commonly misunderstood, and FIR filters are strongly favoured in GNU Radio applications due to their cognitive simplicity and auspicious linear phase response. However,  
  - IIR filters exhibit no side lobes in their response, leading to a more predictable and stable filter performance.
  - IIR filters are numerically more efficient, which is particularly beneficial for real-time signal processing applications.
  - While FIRs are presumed superior w.r.t. phase linearity, especially when using IIR filters like Bessel, which are optimized for phase linearity.

IIR are more predictable and correspond to their well-behaved analog counterparts and, for example, usually exhibit no side-lobs that are common and unavoidable for most FIR filters and produce artefacts in the presence of strong carrier waves. Benchmarks indicate significant performance advantages for IIR filters over their FIR counterparts, particularly in terms of computational complexity:
```plaintext
LOWPASS complexity: #IIR 9 vs. #FIR 733 multiplications & additions
HIGHPASS complexity: #IIR 9  vs. #FIR 489 multiplications & additions
BANDSTOP complexity: #IIR 18 vs. #FIR 7927 multiplications & additions
```
image

The benchmark indicates that -- for the IIR -- the performance is presently limited by the CPU's instructions bandwidth (i.e. CPU clock frequency), which could be further improved to roughly up to 500 MS/s for filter cascades and/or use of FFT-based filtering techniques (to be implemented/benchmarked).

While often associated with non-linear phase responses, the following plots indicate that IIR filters can achieve near-linear phase characteristics when appropriately designed, such as using a Bessel filter, and the errors kept well below the one-degree error over the pass-band (N.B. here the phase offsets are visualisation purposes only):
image

The comparison underlines that the phase linearity advantage, typically attributed to FIR filters, is minimal and can be further minimised with the correct IIR filter design choice. While FIR filters have their exclusive and/or preferred use cases, the above comparison suggests that IIR filters have their merits, especially in high-bandwidth low-, high-, band-pass and band-stop applications, and make them a preferable option for a wide range of filtering tasks. Please experiment, comment and enhance if you find a better, more efficient, cleaner or leaner implementation.

**For further details, please see the corresponding unit tests**

### Some known issues:
 * the present analog vs. digital filter comparison are less than perfect, especially for the `CHEBYSHEV2` implementation.
 * the filter computation using 'float` and `double` works fine using gcc/clang on x86_64 but seems to have some numeric rounding issues for WASM (Emscripten, 32/64 bit) -- the reason is not understood (Emscripten is Clang-based :thinking: )
* IIR-biquads are fine for all orders, but the filter produce stable results w/o using biquads only up to ~ 5th order depending on the lower- or upper-cut-off frequency -- other implementations suggest stable IIR filter up to the 10th order or beyond (-- which I find doubtful... if someone could comment on that).

Your review and feedback on this implementation would be greatly appreciated to refine and optimise the functionality provided. Especially if you can find further improvements or alternative implementations that are more robust in terms of numerics. :pray: :+1:

### DIFF
--- a/algorithm/include/gnuradio-4.0/algorithm/filter/FilterTool.hpp
+++ b/algorithm/include/gnuradio-4.0/algorithm/filter/FilterTool.hpp
@@ -1,0 +1,890 @@
+#ifndef GNURADIO_FILTERTOOL_HPP
+#define GNURADIO_FILTERTOOL_HPP
+
+#include <algorithm>
+#include <cmath>
+#include <complex>
+#include <execution>
+#include <iterator>
+#include <numbers>
+#include <numeric>
+#include <ranges>
+#include <unordered_set>
+#include <vector>
+
+#include <fmt/format.h>
+#include <fmt/ranges.h>
+#include <fmt/std.h>
+#include <magic_enum.hpp>
+#include <magic_enum_utility.hpp>
+
+#include <gnuradio-4.0/algorithm/fourier/window.hpp>
+#include <gnuradio-4.0/HistoryBuffer.hpp>
+#include <gnuradio-4.0/meta/formatter.hpp>
+
+// this mocks the execution policy until Emscripten's libc++ does support this (Clang already does)
+#if defined(__EMSCRIPTEN__)
+
+namespace std {
+
+namespace execution {
+class mock_execution_policy {};
+
+inline constexpr mock_execution_policy seq{};
+inline constexpr mock_execution_policy unseq{};
+inline constexpr mock_execution_policy par{};
+} // namespace execution
+
+template<typename InputIt1, typename InputIt2, typename T, typename BinaryOp1, typename BinaryOp2>
+inline T
+transform_reduce(auto, InputIt1 first1, InputIt1 last1, InputIt2 first2, T init, BinaryOp1 binary_op1, BinaryOp2 binary_op2) {
+    return std::transform_reduce(first1, last1, first2, init, binary_op1, binary_op2);
+}
+
+} // namespace std
+#endif
+
+namespace gr::filter {
+
+enum class Frequency {
+    Hertz,        /// frequency in cycles per second. Standard unit for frequency.
+    RadianPerSec, /// angular frequency, indicating the rate of change per second in radians.
+    Normalised    /// frequency as a fraction of the sampling rate, typically in the range [0, 0.5].
+};
+
+enum class Type { LOWPASS, HIGHPASS, BANDPASS, BANDSTOP };
+
+struct FilterParameters {
+    std::size_t order{ 4UZ };                                      /// default filter order
+    double      fLow{ std::numeric_limits<double>::quiet_NaN() };  /// Lower cutoff frequency [Hertz].
+    double      fHigh{ std::numeric_limits<double>::quiet_NaN() }; /// Upper cutoff frequency [Hertz].
+    double      gain{ 1.0 };                                       /// required total filter gain
+    double      rippleDb{ 0.1 };                                   /// Maximum allowed ripple in the pass-band [dB].
+    double      attenuationDb{ 40 };                               /// Minimum required attenuation in the stop-band [dB].
+    double      beta{ 1.6 };                                       /// default beta for Kaiser-type windowing
+    double      fs{ std::numeric_limits<double>::quiet_NaN() };    /// Sampling frequency for digital filters [Hertz].
+};
+
+/**
+ * @brief Filter coefficients of a digital transfer function H(z) = B(z)/A(z) in the z-domain with:
+ *   B(z) = b[0] + b[1]·z^{-1} + b[2]·z^{-2} + …
+ *   A(z) = a[0] + a[1]·z^{-1} + a[2]·z^{-2} + …
+ *
+ * The difference equation representing the filter is:
+ * y[n] = b[0]·x[n] + b[1]·x[n-1] + … - (a[1]·y[n-1] + a[2]·y[n-2] + …)
+ *
+ * @note Typically, a[0] is 1 for causal systems and a{
+ */
+template<typename T>
+struct FilterCoefficients {
+    using value_type = T;
+    std::vector<T> b{};                    /// numerator coefficients
+    std::vector<T> a{ static_cast<T>(1) }; /// denominator coefficients
+};
+
+template<typename T>
+concept HasFilterCoefficients = requires(T t) {
+    typename T::value_type;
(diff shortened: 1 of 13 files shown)


wirew0rm (approved):
Very good, with links to the relevant literature, nice use of modern std::algorithm primitives and educative and comprehensive unit testing :+1:

PR #219, opened 2023-11-07
Inline header only dependencies

RalphSteinhagen:
### Inline header only dependencies to `./third_party/...` directory.

Presently only 'refl-cpp' and 'magic_enum' -- both expected to be replaced by C++26, for details see: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2996r0.html

Compiler warnings for these external dependencies are enabled by default but can be disabled by the `-DDISABLE_EXTERNAL_DEPS_WARNINGS=ON` cmake flag.

Replaced some custom [from, to] string<->enum by magic_enum ... more to follow. 

This PR focuses more on the CI, build mechanics, and coverage/sonar-line disabling for the inlined header-only libraries.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
         if: matrix.configurations.compiler != 'emscripten'
         shell: bash
         run: |
-          cmake -S . -B ../build -DCMAKE_BUILD_TYPE=${{ matrix.cmake-build-type }} -DENABLE_COVERAGE=${{ matrix.configurations.compiler == 'gcc13' &&  matrix.cmake-build-type == 'Debug'}}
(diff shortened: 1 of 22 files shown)


wirew0rm (approved):
Looks good, the download script works nicely.

Both libs are MIT :heavy_check_mark: 

What we loose in comparison with cmake fetch is the ability to override the dependencies with (standard) cmake options. (For distributions that have a policy against vendored dependencies, the only way now is to patch the buildscripts which will break everytime we change them.) On the other hand this is a lot more lightweight than cmake-fetch.

Code Factor does not exclude the directory and includes lots of issues in the vendored libraries, so we should probably drop that one (or add exclusion rules/fix issues upstream). Probably a repo configuration issue anyway, so out of scope for the PR.

PR #220, opened 2023-11-09
DataSink: Do not hand out unfinished pollers after stop()

frankosterfeld:
Make sure the "finished" flag in the poller is set to true if the poller was created after stop().


RalphSteinhagen (approved):
LGTM!

PR #223, opened 2023-11-20
Implement new DONE logic for the work function using END_OF_STREAM tag.

drslebedev:


core/include/gnuradio-4.0/Block.hpp
```diff
@@ -644,6 +653,30 @@ struct Block : protected std::tuple<Arguments...> {
...
+        } else {
+            if (numerator != 1_UZ || denominator != 1_UZ) {
+                throw std::runtime_error(fmt::format("Block is not defined as `ResamplingRatio<>`, but numerator = {}, denominator = {}, they both must equal to 1.", numerator, denominator));
+            }
+        }
+
+        if constexpr (StrideControl::kEnabled) {
+            static_assert(!kIsSourceBlock, "Stride is not available for source blocks. Remove 'Stride<>' from the block definition.");
+        } else {
+            if (stride != 0_UZ) {
```
RalphSteinhagen, line 674: TODO: should constrain `stride` to be non-zero using the `Annotated<>` interface. This way this is only checked when stride is changed and not during each iteration.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -979,95 +1107,65 @@ struct Block : protected std::tuple<Arguments...> {
...
-            const bool success = consumeReaders(self(), n_samples_to_consume);
-            forward_tags();
-            return { requested_work, ports_status.in_samples, success ? ret : work::Status::ERROR };
+            write_to_outputs(portsStatus.out_samples, writersTuple);
+            const bool success = consumeReaders(self(), nSamplesToConsume);
+            forwardTags();
+            if constexpr (kIsSourceBlock) {
+                if (ret == work::Status::DONE) {
+                    state = LifeCycleState::STOPPED;
+                    publishEOSTag(1); // offset 1 -> just after the last sample
```
RalphSteinhagen, line 1157: :grin: good that you found this... :+1:

core/include/gnuradio-4.0/Port.hpp
```diff
@@ -868,9 +868,9 @@ samples_to_next_tag(const PortLike auto &port) {
     }
 
     // at least one tag is present -> if tag is not on the first tag position read up to the tag position
-    const auto &tagData           = port.tagReader().get();
```
RalphSteinhagen, line 871: This return by reference is important because the port takes care of the lifetime and keeps a reader copy. Dropping the reference here invokes the copy constructor which will create and register a new `CircularBuffer::buffer_reader`

Notably see [here](https://github.com/fair-acc/graph-prototype/blob/caccc351279d55a9f825830f25d79ada6badf676/core/include/gnuradio-4.0/CircularBuffer.hpp#L482-L486):
```cpp
        buffer_reader(std::shared_ptr<buffer_impl> buffer) noexcept :
            _buffer(buffer), _size(buffer->_size) {
            gr::detail::addSequences(_buffer->_read_indices, _buffer->_cursor, {_read_index});
            _read_index_cached = _read_index->value();
        }
```

core/include/gnuradio-4.0/Port.hpp
```diff
@@ -881,6 +881,38 @@ samples_to_next_tag(const PortLike auto &port) {
     }
 }
 
+constexpr std::size_t
+samples_to_eos_tag(const PortLike auto &port) {
+    if (port.tagReader().available() == 0) [[likely]] {
+        return std::numeric_limits<std::size_t>::max(); // default: no tags in sight
+    }
+    const auto tags    = port.tagReader().get();
```
RalphSteinhagen, line 889: same here should be by reference.

core/test/qa_Block.cpp
```diff
@@ -35,7 +35,7 @@ struct IntDecTestData {
 
     std::string
     to_string() const {
-        return fmt::format("n_samples: {}, numerator: {}, denominator: {}, out_port_min: {}, out_port_max: {}, exp_in: {}, exp_out: {}, exp_counter: {}", n_samples, numerator, denominator,
```
RalphSteinhagen, line 38: naming convention.

core/test/qa_Scheduler.cpp
```diff
@@ -8,153 +8,180 @@ template<>
...
 #endif
 
-using trace_vector_type = std::vector<std::string>;
+using TraceVectorType = std::vector<std::string>;
 
-class tracer {
-    std::mutex        _trace_mutex;
-    trace_vector_type _trace_vector;
+class Tracer {
+    std::mutex      _traceMutex;
```
RalphSteinhagen, line 14: N.B. camelCase is ok here since these are private member fields.

core/test/qa_Scheduler.cpp
```diff
@@ -8,153 +8,180 @@ template<>
...
+        if (count >= nSamplesMax) {
+            this->requestStop();
+        }
+        tracer->trace(this->name);
+        return static_cast<int>(count);
     }
 };
 
-static_assert(gr::BlockLike<count_source<float, 10U>>);
+ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T), (CountSource<T>), out, nSamplesMax);
```
RalphSteinhagen, line 54: naming convention.

core/test/qa_Scheduler.cpp
```diff
@@ -208,50 +243,85 @@ checkBlockNames(const std::vector<TBlock> &joblist, std::set<std::string> set) {
     }
 }
 
+template<typename T>
```
RalphSteinhagen, line 246: Since this class is useful also for other unit-tests, it should be moved/declared in the blocklib (alongside the other test blocks).

Mind the naming convention and exported public member field names (snake_case)

RalphSteinhagen: Generally, we should aim to have most of the unit tests reuse the same test block definitions.

RalphSteinhagen (approved):
Very nice. We already discussed most of the changes and implementation in person so no need to repeat it here. 

I just added some minor comments: 
 * please revisit the by-reference usage of the buffer reader and writer. Copying by value creates independent readers while the port should keep a/the one copy.
 * you applied the 'boy-scout-rule' pretty well. Thanks for changing the class names etc. where needed. For public member fields the naming guide prescribes 'snake_case' to maintain compatibility with Python where snake_case is preferred (probably the Python dogma ;-) )
 
I trust your judgement, and for my part once you the above changes are in place you could directly merge these changes.

Thanks again for the good and tremendous work and for identifying/cleaning-up the various corner cases. Well done. :+1:

core/test/qa_Scheduler.cpp
drslebedev: I will extend it in the next "LifeCycle" PR and move to the blocklib.

PR #224, opened 2023-11-24
Fix the problem in Scheduler

drslebedev:
This PR fixes the problem with `Scheduler` when using only one thread (`n_batches = 1`) with execution Policy=multiThreaded.
If only one thread is used no need to wait updates from other threads, it will never happen.

### DIFF
--- a/core/include/gnuradio-4.0/Scheduler.hpp
+++ b/core/include/gnuradio-4.0/Scheduler.hpp
@@ -138,15 +138,16 @@ public:
 
     void
     poolWorker(const std::function<work::Result()> &work, std::size_t n_batches) {
-        auto    &profiler_handler = _profiler.forThisThread();
+        auto &profiler_handler = _profiler.forThisThread();
 
-        uint32_t done             = 0;
-        uint32_t progress_count   = 0;
+        uint32_t done           = 0;
+        uint32_t progress_count = 0;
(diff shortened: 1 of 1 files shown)


RalphSteinhagen (approved):
Thanks for the PR. 

The explicit `{}`init is also good (maybe `= 0` would be easier to read for most). General/style-wise, we shouldn't have to variables declared on the same line.

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -171,6 +171,9 @@ class SchedulerBase {
                 if (progress_count == progress_count_old && done < n_batches) {
```
wirew0rm, line 171:
```suggestion
                if (progress_count == progress_count_old && done + 1 < n_batches) {
```

I think this is the root problem, since the done count is only incremented in the `progress_new` and not in the underlying `done` in line 165. This should also be a problem for more than one thread.

drslebedev: Thanks!  This is also valid change!
 If nothing happened, and progress remains the same and all threads are done (`done + 1 < n_batches`) no need to wait.

`done < n_batches` means that the last thread can still wait for the progress. It is strange why it did not show up in the test.

PR #226, opened 2023-11-28
Implementation of Propagation of Uncertainties

RalphSteinhagen:
### Implementation of [Propagation of Uncertainties](https://en.wikipedia.org/wiki/Propagation_of_uncertainty) via `UncertainValue<T>`

This pull request introduces the `UncertainValue<>` template class: 
```cpp
template<arithmetic_or_complex_like T>
struct UncertainValue {
    T value       = static_cast<T>(0); /// mean value
    T uncertainty = static_cast<T>(0); /// uncorrelated standard deviation
   // [..]
};
```
Here, `uncertainty` represents the estimated measurement range within which the true value of a measurement lies, often expressed as a standard deviation in the context of a Gaussian (normal) distribution. This standard deviation quantifies the expected variability or spread of repeated measurements around the mean value, assuming a normal distribution of errors.

The class is designed to facilitate the propagation of uncertainties in calculations, a concept crucial in scientific, engineering, and real-time feedback applications (e.g. gain-scheduling, conditioning abort situations, etc.). It is inspired by:
Evan Manning, *"Uncertainty Propagation in C++"*, NASA Jet Propulsion Laboratory, C/C++ Users Journal Volume 14, Number 3, March, 1996, [reference](http://www.pennelynn.com/Documents/CUJ/HTML/14.03/MANNING/MANNING.HTM)

#### Key Features:
1. **Support for Basic and Complex Arithmetic Types:** The `UncertainValue` class implements operators for basic arithmetic (`+`, `+=`, `-`, `-=`, `*`, `*=`, `/`, `/=`), and is compatible with both standard arithmetic and complex number types.

2. **Error Propagation for Uncorrelated Variables:** The class focuses on the propagation of errors between uncorrelated variables. It does not cover the calculation and propagation of covariance matrices, which would require a more rigorous approach.

3. **Template-Based Design:** The class is templated to work with any type that satisfies the `arithmetic_or_complex_like` concept, making it versatile and adaptable to various data types.

4. **Standard Mathematical Operations:** The implementation includes overloads for several standard mathematical functions like `pow`, `sqrt`, `sin`, `cos`, and `exp`, ensuring that `UncertainValue` objects can be used seamlessly in mathematical expressions.

5. **Efficient and Safe Implementation:** The class is designed with efficiency in mind, using `constexpr` and `noexcept` where applicable. It also includes safeguards against edge cases, such as division by zero.

Basic usage example:
```cpp
UncertainValue<double> value{ 4.0, 1.0 };

UncertainValue<double> sum1{ 0.0, 0.0 };
UncertainValue<double> sum2{ 0.0, 0.0 };
for (std::size_t i = 0UZ; i < 16UZ; i++) {
    sum1 += value / 16.0;
    sum2 = sum2 + value / 16.0;
}
expect(eq(4.0, sum1.value));
expect(eq(4.0, sum2.value));
expect(approx(value.uncertainty / sqrt(16), sum1.uncertainty, 1e-9));
expect(approx(value.uncertainty / sqrt(16), sum2.uncertainty, 1e-9));
```

**Note to Reviewers:**
The propagation of uncertainties follows the text-book approach for real-valued arithmetic types. Please review the implementation for mathematical correctness, notably for the propagation of `std::complex<...>`-based uncertainties.

Any feedback or suggestions for improvement are highly appreciated.

### DIFF
--- a/meta/include/gnuradio-4.0/meta/UncertainValue.hpp
+++ b/meta/include/gnuradio-4.0/meta/UncertainValue.hpp
@@ -34,7 +34,7 @@ concept arithmetic_or_complex_like = std::is_arithmetic_v<T> || meta::complex_li
 template<arithmetic_or_complex_like T>
 struct UncertainValue {
     using value_type = T;
-    //using T2 = typename std::conditional_t<std::is_arithmetic_v<T>, T, typename T::value_type>; // TODO: fix later for handling symmetric errors for std::complex
+    // using T2 = typename std::conditional_t<std::is_arithmetic_v<T>, T, typename T::value_type>; // TODO: fix later for handling symmetric errors for std::complex
 
     T value       = static_cast<T>(0); /// mean value
     T uncertainty = static_cast<T>(0); /// uncorrelated standard deviation (TODO: make T2)
@@ -60,7 +60,9 @@ struct UncertainValue {
         return *this;
     }
 
-    auto operator<=>(UncertainValue const&) const = default;
+    auto
+    operator<=>(UncertainValue const &) const
+            = default;
 };
 
 template<typename T>
@@ -87,12 +89,12 @@ using UncertainValueType_t = detail::UncertainValueValueType<T>::type;
 /********************** some basic math operation definitions *********************************/
 
 template<typename T, typename U, typename ValueTypeT = UncertainValueType_t<T>, typename ValueTypeU = UncertainValueType_t<U>>
-    requires (UncertainValueLike<T> || UncertainValueLike<U>) // TODO: && std::is_same_v<ValueTypeT, ValueTypeU>
+    requires(UncertainValueLike<T> || UncertainValueLike<U>) // TODO: && std::is_same_v<ValueTypeT, ValueTypeU>
 [[nodiscard]] inline constexpr auto
 operator+(const T &lhs, const U &rhs) noexcept {
     if constexpr (UncertainValueLike<T> && UncertainValueLike<U>) {
         using ResultType = decltype(lhs.value + rhs.value);
-        if constexpr  (meta::complex_like<ValueTypeT> || meta::complex_like<ValueTypeU>) {
+        if constexpr (meta::complex_like<ValueTypeT> || meta::complex_like<ValueTypeU>) {
             // we are dealing with complex numbers -> use the standard uncorrelated calculation.
             ResultType newUncertainty = { std::hypot(std::real(lhs.uncertainty), std::real(rhs.uncertainty)), std::hypot(std::imag(lhs.uncertainty), std::imag(rhs.uncertainty)) };
             return UncertainValue<ResultType>{ lhs.value + rhs.value, newUncertainty };
@@ -117,13 +119,23 @@ operator+=(T &lhs, const U &rhs) noexcept {
     return lhs;
 }
 
+template<UncertainValueLike T, typename ValueTypeT = UncertainValueType_t<T>>
+inline constexpr T
+operator+(const T &val) {
+    if (meta::complex_like<ValueTypeT>) {
+        return val;
+    } else {
+        return { std::abs(val.value), std::abs(val.uncertainty) };
+    }
+}
+
 template<typename T, typename U, typename ValueTypeT = UncertainValueType_t<T>, typename ValueTypeU = UncertainValueType_t<U>>
     requires UncertainValueLike<T> || UncertainValueLike<U> // TODO: && std::is_same_v<ValueTypeT, ValueTypeU>
 [[nodiscard]] inline constexpr auto
 operator-(const T &lhs, const U &rhs) noexcept {
     if constexpr (UncertainValueLike<T> && UncertainValueLike<U>) {
         using ResultType = decltype(lhs.value - rhs.value);
-        if constexpr  (meta::complex_like<ValueTypeT> || meta::complex_like<ValueTypeU>) {
+        if constexpr (meta::complex_like<ValueTypeT> || meta::complex_like<ValueTypeU>) {
             // we are dealing with complex numbers -> use the standard uncorrelated calculation.
             ResultType newUncertainty = { std::hypot(std::real(lhs.uncertainty), std::real(rhs.uncertainty)), std::hypot(std::imag(lhs.uncertainty), std::imag(rhs.uncertainty)) };
             return UncertainValue<ResultType>{ lhs.value - rhs.value, newUncertainty };
@@ -147,13 +159,19 @@ operator-=(T &lhs, const U &rhs) noexcept {
     return lhs;
 }
 
+template<UncertainValueLike T>
+inline constexpr T
+operator-(const T &val) {
+    return { -val.value, val.uncertainty };
+}
+
 template<typename T, typename U, typename ValueTypeT = UncertainValueType_t<T>, typename ValueTypeU = UncertainValueType_t<U>>
     requires UncertainValueLike<T> || UncertainValueLike<U> // TODO: && std::is_same_v<ValueTypeT, ValueTypeU>
 [[nodiscard]] inline constexpr auto
 operator*(const T &lhs, const U &rhs) noexcept {
     if constexpr (UncertainValueLike<T> && UncertainValueLike<U>) {
         using ResultType = decltype(lhs.value * rhs.value);
-        if constexpr  (meta::complex_like<ValueTypeT> || meta::complex_like<ValueTypeU>) {
+        if constexpr (meta::complex_like<ValueTypeT> || meta::complex_like<ValueTypeU>) {
             // we are dealing with complex numbers -> use standard uncorrelated calculation
             ResultType newUncertainty = { std::hypot(std::real(lhs.value) * std::real(rhs.uncertainty), std::real(rhs.value) * std::real(lhs.uncertainty)),
                                           std::hypot(std::imag(lhs.value) * std::imag(rhs.uncertainty), std::imag(rhs.value) * std::imag(lhs.uncertainty)) };
@@ -185,13 +203,13 @@ template<typename T, typename U, typename ValueTypeT = UncertainValueType_t<T>, 
 operator/(const T &lhs, const U &rhs) noexcept {
     if constexpr (UncertainValueLike<T> && UncertainValueLike<U>) {
         using ResultType = decltype(lhs.value * rhs.value);
-        if constexpr  (meta::complex_like<ValueTypeT> || meta::complex_like<ValueTypeU>) {
+        if constexpr (meta::complex_like<ValueTypeT> || meta::complex_like<ValueTypeU>) {
             // we are dealing with complex numbers -> use standard uncorrelated calculation
             ResultType newUncertainty;
             if constexpr (std::is_arithmetic_v<ValueTypeT> && meta::complex_like<ValueTypeU>) {
                 // LHS is real, RHS is complex
                 newUncertainty = { std::sqrt(std::pow(lhs.uncertainty / std::real(rhs.value), 2)), std::sqrt(std::pow(std::imag(rhs.uncertainty) * lhs.value / std::norm(rhs.value), 2)) };
-            } else if constexpr  (meta::complex_like<ValueTypeT> && std::is_arithmetic_v<ValueTypeU>) {
+            } else if constexpr (meta::complex_like<ValueTypeT> && std::is_arithmetic_v<ValueTypeU>) {
                 // LHS is complex, RHS is real
                 newUncertainty = { std::hypot(std::real(lhs.uncertainty) / rhs.value, rhs.uncertainty * std::real(lhs.value) / std::pow(rhs.value, 2)),
                                    std::sqrt(std::pow(std::imag(lhs.uncertainty) / rhs.value, 2)) };
@@ -223,7 +241,7 @@ operator/=(T &lhs, const U &rhs) noexcept {
     return lhs;
 }
 
-} // namespace gr::meta
+} // namespace gr
 
 namespace std { // std:: basic math overloads
 
--- a/meta/include/gnuradio-4.0/meta/UncertainValue.hpp
+++ b/meta/include/gnuradio-4.0/meta/UncertainValue.hpp
@@ -34,10 +34,9 @@ concept arithmetic_or_complex_like = std::is_arithmetic_v<T> || meta::complex_li
 template<arithmetic_or_complex_like T>
 struct UncertainValue {
     using value_type = T;
-    // using T2 = typename std::conditional_t<std::is_arithmetic_v<T>, T, typename T::value_type>; // TODO: fix later for handling symmetric errors for std::complex
 
     T value       = static_cast<T>(0); /// mean value
-    T uncertainty = static_cast<T>(0); /// uncorrelated standard deviation (TODO: make T2)
+    T uncertainty = static_cast<T>(0); /// uncorrelated standard deviation
 
     // Default constructor
     constexpr UncertainValue() noexcept = default;
@@ -89,7 +88,7 @@ using UncertainValueType_t = detail::UncertainValueValueType<T>::type;
 /********************** some basic math operation definitions *********************************/
 
 template<typename T, typename U, typename ValueTypeT = UncertainValueType_t<T>, typename ValueTypeU = UncertainValueType_t<U>>
-    requires(UncertainValueLike<T> || UncertainValueLike<U>) // TODO: && std::is_same_v<ValueTypeT, ValueTypeU>
+    requires(UncertainValueLike<T> || UncertainValueLike<U>) && std::is_same_v<meta::fundamental_base_value_type_t<ValueTypeT>, meta::fundamental_base_value_type_t<ValueTypeU>>
 [[nodiscard]] inline constexpr auto
 operator+(const T &lhs, const U &rhs) noexcept {
     if constexpr (UncertainValueLike<T> && UncertainValueLike<U>) {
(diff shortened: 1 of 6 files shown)


meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+    constexpr UncertainValue &
+    operator=(const T &other) noexcept {
+        value       = other;
+        uncertainty = static_cast<T>(0);
+        return *this;
+    }
+
+    constexpr UncertainValue &
+    operator=(T &&other) noexcept {
```
ivan-cukic, line 66: If this is meant to work only with arithmetic and complex, not sure we need the move operations as copy and move are the same for those.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+    auto operator<=>(UncertainValue const&) const = default;
+};
+
+template<typename T>
+UncertainValue(T, T) -> UncertainValue<T>;
+
+template<typename T>
+concept UncertainValueLike = requires(T t) {
+    typename T::value_type;                                           // UncertainValue must have a nested value_type
+    { t.uncertainty } -> std::convertible_to<typename T::value_type>; // Must have an 'uncertainty' field
```
ivan-cukic, line 81: It should also have the value field.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+};
+
+namespace detail {
+template<typename T>
+struct UncertainValueTypeType {
```
ivan-cukic, line 86: UncertainValueValueType? (as in ValueType for UncertainValue)

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+template<typename T>
+using UncertainValueType_t = detail::UncertainValueTypeType<T>::type;
+
+/********************** some basic math operation definitions *********************************/
+
+template<typename T, typename U>
+    requires UncertainValueLike<T> || UncertainValueLike<U>
+[[nodiscard]] inline constexpr auto
+operator+(const T &lhs, const U &rhs) noexcept {
```
ivan-cukic, line 104: Do we want to enforce these having the same type?

The single parameter constructor T -> UncertainValue<T> could even be implicit if you want to allow mixing uncertain and normal values in one thing.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+};
+} // namespace detail
+
+template<typename T>
+using UncertainValueType_t = detail::UncertainValueTypeType<T>::type;
+
+/********************** some basic math operation definitions *********************************/
+
+template<typename T, typename U>
+    requires UncertainValueLike<T> || UncertainValueLike<U>
```
ivan-cukic, line 102: Instead of a concept-based check, I'd rather have a check that uses `is_instantiation_of`.

This way, if someone has a custom type with value and uncertainty fields in their type, this will create a operator+ for them quite unexpectedly. Potentially even collide with their type's operator+.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+    using ValueTypeU = UncertainValueType_t<U>;
+    if constexpr (UncertainValueLike<T> && UncertainValueLike<U>) {
+        using ResultType = decltype(lhs.value + rhs.value);
+        if constexpr (complex_like<ValueTypeT> || complex_like<ValueTypeU>) {
+            // we are dealing with complex numbers -> use the standard uncorrelated calculation.
+            ResultType newUncertainty = { std::hypot(std::real(lhs.uncertainty), std::real(rhs.uncertainty)), std::hypot(std::imag(lhs.uncertainty), std::imag(rhs.uncertainty)) };
+            return UncertainValue<ResultType>{ lhs.value + rhs.value, newUncertainty };
+        } else {
+            // both ValueType[T,U] are arithmetic uncertainties
+            return UncertainValue<ResultType>{ lhs.value + rhs.value, std::hypot(lhs.uncertainty, rhs.uncertainty) };
```
ivan-cukic, line 115: IMO, this is not universal. The numeric error can be calculated in different ways, not only with standard deviation.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+    if constexpr (UncertainValueLike<T> && UncertainValueLike<U>) {
+        using ResultType = decltype(lhs.value * rhs.value);
+        if constexpr (complex_like<ValueTypeT> || complex_like<ValueTypeU>) {
+            // we are dealing with complex numbers -> use standard uncorrelated calculation
+            ResultType newUncertainty = { std::hypot(std::real(lhs.value) * std::real(rhs.uncertainty), std::real(rhs.value) * std::real(lhs.uncertainty)),
+                                          std::hypot(std::imag(lhs.value) * std::imag(rhs.uncertainty), std::imag(rhs.value) * std::imag(lhs.uncertainty)) };
+            return UncertainValue<ResultType>{ lhs.value * rhs.value, newUncertainty };
+        } else {
+            // both ValueType[T,U] are arithmetic uncertainties
+            auto combinedUncertainty = std::hypot(lhs.value * rhs.uncertainty, rhs.value * lhs.uncertainty);
```
ivan-cukic, line 182: Maybe this would be better (multiplication is great with relative errors, so going to relative and back to absolute):

```
    lhs.value * rhs.value * std::hypot(lhs.uncertainty / lhs.value, rhs.uncertainty / rhs.value);
```

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+template<UncertainValueLike T, typename U>
+inline constexpr T &
+operator/=(T &lhs, const U &rhs) noexcept {
+    lhs = lhs / rhs;
+    return lhs;
+}
+
+} // namespace gr::meta
+
+namespace std { // std:: basic math overloads
```
ivan-cukic, line 251: Leave it in the the gr::meta namespace please. Defining things in std:: is generally UB, and allowed only for specific situations. I can't find that that pow ect. were meant to be customization points.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+template<typename T>
+    requires std::is_floating_point_v<T>
+[[nodiscard]] inline constexpr T
+hypot(const std::complex<T> &a, const std::complex<T> &b) noexcept { // doesn't exist in the STL for some reason
```
ivan-cukic, line 256: This is even more dangerous. All exceptions to the 'don''t define things in std::' require at least one user type. And here, all types are std ones. (so, even if the general comment from above is something that we could ignore, this isn't)

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+#include <complex>
+#include <concepts>
+#include <cstdint>
+#include <numbers>
+#include <optional>
+#include <type_traits>
+
+#include <gnuradio-4.0/meta/utils.hpp>
+
+namespace gr::meta {
```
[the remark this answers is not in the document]
RalphSteinhagen: what would your propose?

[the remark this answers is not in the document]
ivan-cukic: Probably just gr, or a new namespace for types like these - meta is for meta-programming things.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+            return T{ 1, 0 };
+        } else {
+            return T{ 0, 0 };
+        }
+    }
+
+    ValueTypeT newValue = std::pow(base.value, exponent);
+    if constexpr (gr::meta::complex_like<ValueTypeT>) {
+        auto val = exponent / base.value * newValue;
+        return T{ newValue, std::sqrt(val * std::conj(val)) * base.uncertainty };
```
marcusmueller, line 274: `std::abs` is defined for `std::complex` in `<complex>`, not sure why we have to special case that?

marcusmueller (changes requested):
I like this idea a lot but,

- ~code-wise, I don't think we're allowed to populate the `std::` namespace,~ Covered by @ivan-cukic  already
- mathematically, I'm fairly certain that the underlying assumption on how `uncertainty` is handled for complex values here is that the real and imaginary part are independent. However, that breaks down for the multiplication and division operators, and it totally doesn't make sense for `pow`.

As an electrical engineer, I need references explaining what kind of math is done here – it's not something I'm accustomed to.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
RalphSteinhagen: done

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
RalphSteinhagen: added concept/static_assert to enforce that the value types are the same.

RalphSteinhagen: implicit conversion for casting artihmetic to UncertainValue (implicit uncertainty = 0) is optimised out by gcc but not clang. Example: `std::hypoth(...)` may need to address this later once clang's optimiser became a bit smarter.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
RalphSteinhagen: we handle only uncorrelated standard deviations (i.e. no asymmetric or correlated values).

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
RalphSteinhagen: TODO: check with @mattkretz  whether overloading std math functions with user-defined types is acceptable or not

Example:
```cpp
template<gr::UncertainValueLike T, typename ValueTypeT = gr::UncertainValueType_t<T>>
[[nodiscard]] inline constexpr T
std::sin(const T &x) noexcept {
    return T{ std::sin(x.value), std::abs(std::cos(x.value) * x.uncertainty) };
}
```

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+
+template<typename T>
+concept arithmetic_or_complex_like = std::is_arithmetic_v<T> || complex_like<T>;
+
+template<arithmetic_or_complex_like T>
+struct UncertainValue {
+    using value_type = T;
+
+    T value       = static_cast<T>(0); /// mean value
+    T uncertainty = static_cast<T>(0); /// uncorrelated standard deviation
```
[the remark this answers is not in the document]
marcusmueller: I wish I had – I'm really not a physicist with a use case for this. Yes, we do commonly deal with I/Q imbalance, have to scale one of the branches to fit the signal amplitude of the other, and thus have different noise variances in real and imaginary part:

It's just that if you treat them as independent errors, that goes southwards as soon as you do *any* operation that "rotates" the complex value; essentially, any multiplicative operation means that afterwards, your errors in I and Q are correlated, as explained somewhere further below :) And among the first things you have to do when you have complex baseband is to rotate the phase so that the complex signal-contained information is recoverable.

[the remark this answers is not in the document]
marcusmueller: See especially [this](https://github.com/fair-acc/graph-prototype/pull/226#discussion_r1409167020) comment below; your multiplication operation, as far as I read it, keeps the output I uncertainty independent from the inputs' Q uncertainty. That's impossible.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,331 @@
...
+};
+
+namespace detail {
+template<typename T>
+struct UncertainValueTypeType {
+    using type = T;
+};
+
+template<typename T>
+struct UncertainValueTypeType<UncertainValue<T>> {
```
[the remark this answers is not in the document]
RalphSteinhagen: follows [CORE_NAMING_GUIDELINE.md](https://github.com/fair-acc/graph-prototype/blob/main/CORE_NAMING_GUIDELINE.md)

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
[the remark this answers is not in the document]
RalphSteinhagen: Dear  @marcusmueller,

Thank you for your detailed comments, time, and effort in reviewing the implementation of `UncertainValue` for complex numbers. I want to clarify the assumptions and practical applications of this code.

The PR and code description clearly outline the implementation's basis and the propagation of uncorrelated symmetric errors, specifically Gaussian-type standard deviations. 
```cpp
/**
 *
 * @brief Propagation of Uncertainties
 *
 * original idea by: Evan Manning, "Uncertainty Propagation in C++", NASA Jet Propulsion Laboratory,
 * C/C++ Users Journal Volume 14, Number 3, March, 1996
 * http://www.pennelynn.com/Documents/CUJ/HTML/14.03/MANNING/MANNING.HTM
 *
 * implements +,-,*,/ operators for basic arithmetic and complex types, for details see:
 * https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Example_formulae
 * This implements only propagation of uncorrelated symmetric errors (i.e. gaussian-type standard deviations).
 * A more rigorous treatment would require the calculation and propagation of the
 * corresponding covariance matrix which is out of scope of this implementation.
 */
```

This approach is tailored to our specific use case, where we map pairs of real-valued ADC samples with real-valued, uncorrelated uncertainties for each channel to `std::complex<T>`. In our context, complex numbers serve more as a convenient representation for pairs of independent values (akin to vectors or tuples) rather than traditional complex numbers with correlated real and imaginary parts.

In our specific applications, such as gain scheduling and signal conditioning in scientific and engineering contexts, we primarily deal with linear operations where the independence of uncertainties remains valid. This includes scenarios where complex numbers are added, subtracted or scaled by non-complex numbers, and the uncertainties of each component are independent, originating from different noise sources. For multiplicative or non-linear operations, while the post-operation correlation between real and imaginary parts might occur, our focus is on the initial step where the noise sources are uncorrelated. As an 80/20 engineering approximation, this holds for most relevant use cases in signal processing and feedback control applications.

Furthermore, in the context of operations like FFT, the linear treatment of uncertainties remains applicable, e.g.
$$\mathcal{FFT}(A(t)+B(t)) = \mathcal{FFT}(A(t)) + \mathcal{FFT}(B(t))$$ and notably
$$\mathcal{FFT}(\overline{x}(t) \pm \overline{\sigma_A}(t)) = \mathcal{FFT}((\overline{x}(t)) \pm FFT(\overline{\sigma_A}(t)).$$

This linear approach simplifies the otherwise complex task of handling large covariance matrices, which, while more rigorous, would be impractical and unnecessary for the most common use cases.

I acknowledge that this implementation does not cover all possible scenarios, especially those involving correlated uncertainties or non-linear operations. However, for our targeted applications (filtering, FFT, gain-scheduling, signal conditioning, ...), this approach balances accuracy and computational feasibility. 

For practical engineering purposes, at worst, this approach slightly underestimates the propagated error compared to a fully (anti-)correlated treatment — by only about 30%, not orders of magnitude. I feel this level of approximation is should be acceptable for most use cases. Moreover, it's crucial to emphasize that having an uncertainty estimate, even if not perfect, is significantly better than having none, which is currently the status quo in GR.

I hope this helps clear doubts and motivates the reasons behind the specific implementation choices. I am open to further discussion and am eager to hear any additional new thoughts or suggestions you might have.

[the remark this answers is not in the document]
marcusmueller: It remains wrong, and in a worse way, for the multiplication of two complexes with uncertainties, because that isn't even a linear operation.

However, when you look at your source code for the uncertainty of the imaginary part of the product, you see that the input's uncertainty in the real part doesn't seem to play a role. There's just some fundamentally wrong math here, and I don't really understand where the math comes from, so that's why I keep asking for an actual reference rather than just handwaving.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
RalphSteinhagen: moved to `namespace gr::math` -- don't want to break the world or annoy the C++ gods.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
```diff
@@ -0,0 +1,319 @@
...
+inline constexpr T &
+operator+=(T &lhs, const U &rhs) noexcept {
+    lhs = lhs + rhs;
+    return lhs;
+}
+
+template<UncertainValueLike T, typename ValueTypeT = UncertainValueType_t<T>>
+inline constexpr T
+operator+(const T &val) {
+    if (meta::complex_like<ValueTypeT>) {
```
ivan-cukic, line 124: if constexpr

ivan-cukic (approved):
One nit-pick remaining.  Otherwise, approved from the C++-code-side of things.

meta/include/gnuradio-4.0/meta/UncertainValue.hpp
RalphSteinhagen: done

PR #232, opened 2023-12-09
added HistoryBuffer::reset(<default>)

RalphSteinhagen:
... and filled some minor loop-holes in unit-tests (coverage)


wirew0rm (approved):
Thanks you, looks good and will be needed for implementing the state logic for the blocks using the history buffer and also make it easier to use in other contexts (e.g. plotting).

The decision to explicitly initialize the unused (size > i > capacity) memory corresponds to the behavior at construction and also some use-cases which do not want to do the size checks and therefore require initialization of all memory.

PR #233, opened 2023-12-14
Implementation of FunctionGenerator class

drslebedev:
The `FunctionGenerator` class generates a variety of functions and their combinations.
It supports multiple function types, including Constant, Linear Ramp, Parabolic Ramp, Cubic Spline, and Impulse Response.
Each function type is configurable with specific parameters, enabling precise tailoring to meet the user's requirements.
Users can create sequences of various functions using tags.
Some helper methods to create tags or property maps are provided: 1) `create{FunctionName}Tag(...)`, 2) `create{FunctionName}PropertyMap(...)`

Supported function types and their configurable parameters:

* Const
  - Constantly returns the same value, specified by 'startValue'.
  To generate a tag use `createConstTag(tagIndex, startValue)`.

* LinearRamp
  - Interpolates linearly between a start and a final value over a set duration.
  To generate a tag use `createLinearRampTag(tagIndex, startValue, finalValue, duration)`.

* ParabolicRamp
  - Produces a function with three sections: parabolic-linear-parabolic. The 'roundOffTime' parameter determines the length of the parabolic sections.
  To generate a tag use `createParabolicRampTag(tagIndex, startValue, finalValue, duration, roundOffTime)`.

* CubicSpline
  - Implements smooth cubic spline interpolation between start and final values over a specified duration.
  To generate a tag use `createCubicSplineTag(tagIndex, startValue, finalValue, duration)`.

* ImpulseResponse
  - Creates a function that spikes from 'time0' for a duration of 'time1'.
  To generate a tag use `createImpulseResponseTag(tagIndex, startValue, finalValue, time0, time1)`.

### 2 ways to create a chain of functions using `ClockSource`. 

1) Using times + property_map
Firstly, add time Tag entries, with time and command string:
```cpp
auto &clockSrc = testGraph.emplaceBlock<gr::basic::ClockSource<float>>({gr::tag::SAMPLE_RATE(sample_rate), { "n_samples_max", 100 }, { "name", "ClockSource" } });
clockSrc.addTimeTagEntry(1'000, "CMD_BP_START:FAIR.SELECTOR.C=1:S=1:P=1");
clockSrc.addTimeTagEntry(10'000, "CMD_BP_START:FAIR.SELECTOR.C=1:S=1:P=2");
clockSrc.addTimeTagEntry(30'000, "CMD_BP_START:FAIR.SELECTOR.C=1:S=1:P=3");
```

Secondly, create table to match command and functions:
```cpp
funcGen.addFunctionTableEntry({ { "context", "CMD_BP_START:FAIR.SELECTOR.C=1:S=1:P=1" } }, createConstPropertyMap(5.f));
funcGen.addFunctionTableEntry({ { "context", "CMD_BP_START:FAIR.SELECTOR.C=1:S=1:P=2" } }, createLinearRampPropertyMap(5.f, 30.f, .2f));
funcGen.addFunctionTableEntry({ { "context", "CMD_BP_START:FAIR.SELECTOR.C=1:S=1:P=3" } }, createConstPropertyMap(30.f));
```

Lastly create a matching predicate with the following signature:
```cpp
std::function<std::optional<bool>(const property_map &, const property_map &, std::size_t)>;
```

2) Using user defined ready-to-use Tags.
```cpp
auto &clockSrc = testGraph.emplaceBlock<gr::basic::ClockSource<float>>({ gr::tag::SAMPLE_RATE(sample_rate), { "n_samples_max", 100 }, { "name", "ClockSource" } });
clockSrc.tags = { Tag(0, createConstPropertyMap(5.f)),
                  Tag(100, createLinearRampPropertyMap(5.f, 30.f, .2f)),
                  Tag(300, createConstPropertyMap(30.f)),
                  Tag(350, createParabolicRampPropertyMap(30.f, 20.f, .1f, 0.02f)),
                  Tag(550, createConstPropertyMap(20.f)),
                  Tag(650, createCubicSplinePropertyMap(20.f, 10.f, .1f)),
                  Tag(800, createConstPropertyMap(10.f)),
                  Tag(850, createImpulseResponsePropertyMap(10.f, 20.f, .02f, .06f)) };
```

This code generates the following output:

func

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
@@ -1,0 +1,327 @@
+#ifndef GNURADIO_FUNCTION_GENERATOR_HPP
+#define GNURADIO_FUNCTION_GENERATOR_HPP
+
+#include <gnuradio-4.0/Block.hpp>
+#include <gnuradio-4.0/Tag.hpp>
+#include <magic_enum.hpp>
+#include <magic_enum_utility.hpp>
+#include <numbers>
+
+namespace gr::basic {
+
+using namespace gr;
+
+namespace function_generator {
+enum class SignalType : int { Const, LinearRamp, ParabolicRamp, CubicSpline, ImpulseResponse };
+// TODO: Enum values should be capitalized CamelStyle, for the moment they are the same as FunctionGenerator class members
+enum class ParameterType : int { signal_type, start_value, final_value, duration, round_off_time, impulse_time0, impulse_time1 };
+
+using enum SignalType;
+using enum ParameterType;
+constexpr auto SignalTypeList = magic_enum::enum_values<SignalType>();
+
+template<typename T>
+    requires std::is_same_v<T, SignalType> || std::is_same_v<T, ParameterType>
+constexpr std::string
+toString(T type) {
+    return std::string(magic_enum::enum_name(type));
+}
+
+template<typename EnumType, typename T>
+    requires std::is_same_v<EnumType, SignalType> || std::is_same_v<EnumType, ParameterType>
+[[nodiscard]] std::pair<std::string, pmtv::pmt>
+createPropertyMapEntry(EnumType enumType, T value) {
+    if constexpr (std::is_same_v<T, SignalType>) {
+        return { toString(enumType), static_cast<pmtv::pmt>(toString(value)) };
+    } else {
+        return { toString(enumType), static_cast<pmtv::pmt>(value) };
+    }
+}
+
+template<typename T>
+    requires std::is_same_v<T, SignalType> || std::is_same_v<T, ParameterType>
+constexpr T
+parse(std::string_view name) {
+    auto type = magic_enum::enum_cast<T>(name, magic_enum::case_insensitive);
+    if (!type.has_value()) {
+        throw std::invalid_argument(fmt::format("parser error, unknown function_generator::{} - '{}'", meta::type_name<T>(), name));
+    }
+    return type.value();
+}
+
+template<typename T>
+[[nodiscard]] property_map
+createConstPropertyMap(T startValue) {
+    return { createPropertyMapEntry(signal_type, Const), createPropertyMapEntry(start_value, startValue) };
+}
+
+template<typename T>
+[[nodiscard]] property_map
+createLinearRampPropertyMap(T startValue, T finalValue, T durationValue) {
+    return { createPropertyMapEntry(signal_type, LinearRamp), createPropertyMapEntry(start_value, startValue), createPropertyMapEntry(final_value, finalValue),
+             createPropertyMapEntry(duration, durationValue) };
+}
+
+template<typename T>
+[[nodiscard]] property_map
+createParabolicRampPropertyMap(T startValue, T finalValue, T durationValue, T roundOffTime) {
+    return { createPropertyMapEntry(signal_type, ParabolicRamp), createPropertyMapEntry(start_value, startValue), createPropertyMapEntry(final_value, finalValue),
+             createPropertyMapEntry(duration, durationValue), createPropertyMapEntry(round_off_time, roundOffTime) };
+}
+
+template<typename T>
+[[nodiscard]] property_map
+createCubicSplinePropertyMap(T startValue, T finalValue, T durationValue) {
+    return { createPropertyMapEntry(signal_type, CubicSpline), createPropertyMapEntry(start_value, startValue), createPropertyMapEntry(final_value, finalValue),
+             createPropertyMapEntry(duration, durationValue) };
+}
+
+template<typename T>
+[[nodiscard]] property_map
+createImpulseResponsePropertyMap(T startValue, T finalValue, T time0, T time1) {
+    return { createPropertyMapEntry(signal_type, ImpulseResponse), createPropertyMapEntry(start_value, startValue), createPropertyMapEntry(final_value, finalValue),
+             createPropertyMapEntry(impulse_time0, time0), createPropertyMapEntry(impulse_time1, time1) };
+}
+
+} // namespace function_generator
+
+using FunctionGeneratorDoc = Doc<R""(
+@brief The `FunctionGenerator` class generates a variety of functions and their combinations.
+It supports multiple function types, including Constant, Linear Ramp, Parabolic Ramp, Cubic Spline, and Impulse Response.
+Each function type is configurable with specific parameters, enabling precise tailoring to meet the user's requirements.
+Users can create sequences of various functions using tags.
+These tags are generated through dedicated helper methods, named `create{FunctionName}Tag(...)`.
+This method enables the flexible assembly of diverse function chains, offering significant versatility in signal generation.
+
+Supported function types and their configurable parameters:
+
+* Const
+  - Constantly returns the same value, specified by 'startValue'.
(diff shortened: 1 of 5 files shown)


blocks/basic/test/qa_sources.cpp
```diff
@@ -146,6 +162,68 @@ const boost::ut::suite TagTests = [] {
...
     };
+
+    "FunctionGenerator ImChart test"_test = [] {
+        const std::size_t        N           = 128; // test points
+        double                   start_value = 10.;
+        double                   final_value = 20.;
+        std::vector<std::string> signals{ "Const", "LinearRamp", "ParabolicRamp", "CubicSpline", "ImpulseResponse" };
+        for (const auto &sig : signals) {
+            FunctionGenerator<double> funcGen{};
+            std::ignore = funcGen.settings().set({ { "signal_type", sig },
```
RalphSteinhagen, line 173: Please don't 'std::ignore' keep the return and assert that it's `empty()` meaning that all settings have been accepted by the block (nice regression unit-test case)

RalphSteinhagen (changes requested):
Looks good! We should change the tag/timing context definitions right away to the ones defined for the FAIR timing. Added some infos in the comments. Let me know if you have questions regarding this.

blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
```diff
@@ -0,0 +1,297 @@
...
+    if (!signalType.has_value()) {
+        throw std::invalid_argument(fmt::format("unknown function generator type '{}'", name));
+    }
+    return signalType.value();
+}
+
+template<typename T>
+[[nodiscard]] property_map
+createConstPropertyMap(T startValue) {
+    return { { "signal_type", "Const" }, { "start_value", startValue } };
```
RalphSteinhagen, line 32: minor:  maybe derive the "Const" and other magic string constants directly from the enum using `magic_enum::enum_name(...)` This would minimise future refactoring/renaming issues and also indicate to the user/other developers that these are API definitions.

Similarly `signal_type`, `start_value`, ... etc. could be also defined via an enum.

blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
```diff
@@ -0,0 +1,297 @@
...
+
+Lastly create a matching predicate with the following signature:
+@code
+std::function<std::optional<bool>(const property_map &, const property_map &, std::size_t)>;
+@endcode
+
+
+2) Using user defined ready-to-use Tags.
+@code
+auto &clockSrc = testGraph.emplaceBlock<gr::basic::ClockSource<float>>({ { "sample_rate", 100.f }, { "n_samples_max", 100 }, { "name", "ClockSource" } });
```
RalphSteinhagen, line 116: minor (and also this is just within a code documentation): maybe we should re-use the `sample_rate` etc. definitions from the default tag definitions to indicate that these are also user-level API definitions and conventions.

blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
```diff
@@ -0,0 +1,297 @@
...
+                  Tag(850, createImpulseResponsePropertyMap(10.f, 20.f, .02f, .06f)) };
+@endcode
+
+)"">;
+
+template<typename T>
+    requires(std::floating_point<T>)
+struct FunctionGenerator : public gr::Block<FunctionGenerator<T>, BlockingIO<true>, FunctionGeneratorDoc> {
+    /**
+     * TODO: Taken from CtxSettings
```
RalphSteinhagen, line 133: fyi: need to follow-up this 'TODO' comment. Will do it in my branch since I also needed the tag-matching-predicate more intensely.

blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
```diff
@@ -0,0 +1,297 @@
...
+    using MatchPredicate = std::function<std::optional<bool>(const property_map &, const property_map &, std::size_t)>;
+
+    PortIn<T>  in; // ClockSource input
+    PortOut<T> out;
+
+    Annotated<float, "sample_rate", Visible, Doc<"sample rate">>                        sample_rate = 1000.f;
+    Annotated<std::string, "signal_type", Visible, Doc<"see function_generator::Type">> signal_type = "Const";
+
+    // Parameters for different functions
+    Annotated<T, "start_value", Visible>                                               start_value    = T(0.);
```
RalphSteinhagen, line 152: Similar to the `SampleRate` above but these should be derived from an enum and using statement that is declared in this function generator namespace.

RalphSteinhagen: N.B. review whether each of this settings needs to be 'Visible`. The idea behind the `Visible` is that it's and UI hint whether these values should be drawn on the top-level block representation. This is regardless of the accessibility/visibility of these fields within the contextt/block settings UI widgets. I.e. 'Visible` should be parameter that the user must see under any other conditions.

I'd say that there -- ideally -- should be as few as possible, mostly only 4-6 parameters total and more only for justified cases (e.g. digitizer, SDR sources etc.).

blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
```diff
@@ -0,0 +1,297 @@
...
+    int sampleCounter = 0;
+
+private:
+    function_generator::Type _signalType = function_generator::parse(signal_type);
+    T                        _timeTick   = T(1.) / static_cast<T>(sample_rate);
+
+public:
+    void
+    settingsChanged(const property_map & /*old_settings*/, const property_map &new_settings) {
+        if (new_settings.contains("trigger_meta_info")) {
```
RalphSteinhagen, line 176: A default-tag/enum-derived definition would help here instead of the literal string.

blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
```diff
@@ -0,0 +1,297 @@
...
+                return std::nullopt;
+            }
+            try {
+                return std::get<U>(it->second);
+            } catch (const std::bad_variant_access &) {
+                return std::nullopt;
+            }
+        };
+
+        signal_type    = getProperty(properties, "signal_type", std::string("")).value_or("Const");
```
RalphSteinhagen, line 237: use enums directly.

blocks/basic/test/qa_sources.cpp
```diff
@@ -164,6 +181,142 @@ const boost::ut::suite TagTests = [] {
...
```
RalphSteinhagen, line 192: as mentioned in the implemented code: we should consider replacing these string literals with either enum or default tag definitions that default convert to the required std::string key value. This would help readability (<-> public API/convention) and minimise typos.

blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
drslebedev: I removed the Visible flag from all function parameters.

blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
drslebedev: For the moment there is a problem with `gr::` prefix, this first need to be fixed. And then one should use where possible definitions from default tags.

PR #234, opened 2023-12-15
Integrate UncertainValue to IIR and FIR filter

RalphSteinhagen:
### Infinite-Impulse-Response (IIR) as well as Finite-Impulse-Response (FIR) filter based on a single or set of biquad filter coefficients.

This includes the computation of the propagation of uncertainty according to:

$$ \sigma_y^2[0] = \sum_{i=0}^{M} b^2[i]·σ_x^2[i] + \sum_{j=1}^{N} \sum_{k=1}^{N} a[j]·a[k]·R_{yy}[|j-k|]·σ_y[j]·σ_y[k] $$

with $R_{yy}[distance]$ being the auto-correlation function of the impulse response as an estimate of the covariance matrix.

#### Usage example:
```cpp
Filter<UncertainValue<double>> myFilter(filterSections);
UncertainValue<double> outputSample = myFilter.processOne({inputSample, noise});
double mean   = gr::value(outputSample);
double stddev = gr::uncertainty(outputSample);
 ```
 
 #### Visual step-responses examples for IIR and FIR filter for 1, 2, and 4 Hz low-pass bandwidth:
 
image
image

The more pronounced linear group-delay for the FIR is normal due to the usually larger filter kernel. N.B. the IIR filter has about a 50% poorer noise/uncertainty suppression which is also linked to the FIR filter kernel size rather than the low-pass cut-off frequency.

#### Preliminary Benchmark Performance:
image

Neither the regular nor `UncertainValue<T>`-enabled IIR/FIR filter implementation have been optimised for `std::simd` use. Hope to get some help/support/guidance by @mattkretz .

### DIFF
--- a/algorithm/include/gnuradio-4.0/algorithm/filter/FilterTool.hpp
+++ b/algorithm/include/gnuradio-4.0/algorithm/filter/FilterTool.hpp
@@ -21,6 +21,7 @@
 #include <gnuradio-4.0/algorithm/fourier/window.hpp>
 #include <gnuradio-4.0/HistoryBuffer.hpp>
 #include <gnuradio-4.0/meta/formatter.hpp>
+#include <gnuradio-4.0/meta/UncertainValue.hpp>
 
 // this mocks the execution policy until Emscripten's libc++ does support this (Clang already does)
 #if defined(__EMSCRIPTEN__)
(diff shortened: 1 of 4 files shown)


drslebedev (approved):
Thanks a lot for the PR!

I like the way you calculate uncertainties for FIR and IIR filters. I think the formula that you use is a good way to do it. The implementation of the calculation itself is also correct. 

The overall structure of the code is clear and modular, which facilitates future extensions.
The code is easy to understand and follow the logic.

I like the use of `std::transform_reduce` with execution policies.

The provided graphs for visualization in unit tests are very helpful.
It could be be beneficial to extend unit tests that address the uncertainty calculations with more checks. Is it possible to check the whole expected array of uncertainties?

Moreover, the performance benchmark results show that the inclusion of uncertainty does not significantly degrade the filter's throughput (less than 50%). This is rather impressive, considering the complexity of the uncertainty propagation formula.

Overall, this PR is a very good contribution and can be merged.

RalphSteinhagen:
> It could be be beneficial to extend unit tests that address the uncertainty calculations with more checks. Is it possible to check the whole expected array of uncertainties?

Yes correct. For the time being this only calculates/verifies that the uncertainty converges to the expected $1/sqrt{N}$ the steady-state value. For the FIR 'car-box-filter' the $\sigma(t)$ is rather trivial but for any other FIR and/or IIR I wouldn't know how to analytically derive the function of time other than computing the FIR/IIR with the filter itself ... :thinking: 

> Moreover, the performance benchmark results show that the inclusion of uncertainty does not significantly degrade the filter's throughput (less than 50%). This is rather impressive, considering the complexity of the uncertainty propagation formula.

Well, there is a performance penalty of about a factor 8 comparing w/ and w/o uncertainty calculation... hope to find a soluition via the 'std::simd'-way. The `transform_reduce` is already prepared but not (yet) using SIMD...

drslebedev:
> Well, there is a performance penalty of about a factor 8 comparing w/ and w/o uncertainty calculation... hope to find a soluition via the 'std::simd'-way. The `transform_reduce` is already prepared but not (yet) using SIMD...

Ah, sorry. I looked only at FIR performance and was impressed that the difference is only factor of 2. For the IRR you are right the difference is factor of 8 for `double`s and factor of 5 for `float`s.

PR #235, opened 2023-12-19
Fix relocating Port instances

vimpostor:
This PR is split into three semantically distinct commits:

- Convert qa_DynamicBlock to a real boost ut test
- Add a test for dynamically reallocating the ports collection
- Fix reallocating Port instances

The final commit is the patch with the actual fix, the move constructor of `Port` was missing some fields, which caused connections to break as soon as a Port was relocated.

This already fixes https://github.com/fair-acc/opendigitizer/issues/90.

However, there is more to the relationship between `DynamicPort` and its `Port` counterpart. Since they are so tightly coupled together, moving a Port needs to modify the accessor in `DynamicPort`. This is not needed in the above unit test, because the `DynamicPort`s [are always recreated](https://github.com/fair-acc/graph-prototype/blob/427aea104b912c02c13063110c8780f9847d8682/blocks/basic/include/gnuradio-4.0/basic/common_blocks.hpp#L94), when the `Port` moves so this is actually an unrelated issue, but it might be worthwhile to fix this in the future as well.

One possible (and naive) way to fix this problem is to add an reallocation callback to `Port`, that can be setup by `DynamicPort` to update its reference on `std::move`. More specifically this diff should do it:

```diff
diff --git a/core/include/gnuradio-4.0/Port.hpp b/core/include/gnuradio-4.0/Port.hpp
index e9a988d..227eb50 100644
--- a/core/include/gnuradio-4.0/Port.hpp
+++ b/core/include/gnuradio-4.0/Port.hpp
@@ -242,6 +242,7 @@ private:
     bool      _connected    = false;
     IoType    _ioHandler    = newIoHandler();
     TagIoType _tagIoHandler = newTagIoHandler();
+    std::function<void(Port<T, portName, portType, portDirection, Attributes...>&)> _moveCallback;
 
 public:
     [[nodiscard]] constexpr bool
@@ -298,6 +299,10 @@ public:
         return true;
     }
 
+    void setMoveCallback(decltype(_moveCallback) callback) {
+        this->_moveCallback = callback;
+    }
+
     constexpr Port()   = default;
     Port(const Port &) = delete;
     auto
@@ -309,7 +314,11 @@ public:
         static_assert(portName.empty(), "port name must be exclusively declared via NTTP or constructor parameter");
     }
 
-    constexpr Port(Port &&other) noexcept : name(std::move(other.name)), priority{ other.priority }, min_samples(other.min_samples), max_samples(other.max_samples) {}
+    constexpr Port(Port &&other) noexcept : name(std::move(other.name)), priority{ other.priority }, min_samples(other.min_samples), max_samples(other.max_samples), _connected(other._connected), _ioHandler(std::move(other._ioHandler)), _tagIoHandler(std::move(other._tagIoHandler)) {
+        if (_moveCallback) {
+            _moveCallback(*this);
+        }
+    }
 
     constexpr Port &
     operator=(Port &&other) noexcept {
@@ -783,7 +792,11 @@ public:
     // can not be reallocated
     template<PortLike T>
     explicit constexpr DynamicPort(T &arg, non_owned_reference_tag) noexcept
-        : name(arg.name), priority(arg.priority), min_samples(arg.min_samples), max_samples(arg.max_samples), _accessor{ std::make_unique<wrapper<T, false>>(arg) } {}
+        : name(arg.name), priority(arg.priority), min_samples(arg.min_samples), max_samples(arg.max_samples), _accessor{ std::make_unique<wrapper<T, false>>(arg) } {
+            arg.setMoveCallback([&](auto &newArg){
+                    this->_accessor = std::make_unique<wrapper<T, false>>(newArg);
+                });
+        }
 
     template<PortLike T>
     explicit constexpr DynamicPort(T &&arg, owned_value_tag) noexcept
```

However I have to admit I don't really like this solution, as it feels a lot like "hotpatching it in". A change in the architecture how we integrate `DynamicPort` and `Port` might be better.
In any case this is probably out of scope of this PR.

### DIFF
--- a/core/test/qa_DynamicBlock.cpp
+++ b/core/test/qa_DynamicBlock.cpp
@@ -47,7 +47,7 @@ ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T), (DebugSink<T>), lastValue, in)
 const boost::ut::suite DynamicBlocktests = [] {
     using namespace boost::ut;
     "Change number of ports dynamically"_test = [] {
-        constexpr const std::size_t sources_count = 10;
+        constexpr const int sources_count = 10;
         constexpr const std::size_t events_count  = 5;
 
         gr::Graph                   testGraph;
@@ -60,7 +60,7 @@ const boost::ut::suite DynamicBlocktests = [] {
 
         // Function that adds a new source node to the graph, and connects
         // it to one of adder's ports
-        std::ignore = adder.settings().set({ { "input_port_count", 10 } });
+        std::ignore = adder.settings().set({ { "input_port_count", sources_count } });
         std::ignore = adder.settings().applyStagedParameters();
 
         std::vector<fixed_source<double> *> sources;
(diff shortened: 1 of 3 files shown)


core/test/qa_DynamicBlock.cpp
```diff
@@ -29,54 +30,84 @@ static_assert(gr::traits::block::input_ports<fixed_source<int>>::size() == 0);
 static_assert(gr::traits::block::output_ports<fixed_source<int>>::size() == 1);
 
 template<typename T>
-struct cout_sink : public gr::Block<cout_sink<T>, gr::PortInNamed<T, "in">> {
-    std::size_t remaining = 0;
+struct debug_sink : public gr::Block<debug_sink<T>, gr::PortInNamed<T, "in">> {
```
RalphSteinhagen, line 33: Minor: 
* naming convention -> `debug_sink` -> `DebugSink`
* Boy-scout rule: 
    * check if there isn't already an existing sink that does the same (remove code duplication)
    * port should be a field member + refl-annotated <-> this is the user-code reference that we want to follow. The tuple-defined ports should be reserved primarily for meta-programming/non-user defined blocks since those C++ concepts are a bit harder to grasp for C++ beginners.

RalphSteinhagen (approved):
Looks good and thanks for hunting down the missing move operations in the rvalue constructor of `Port(&&)`.

This is unrelated to the fix but as a side note on `class multi_adder : public gr::BlockModel` (Boy-scout rule):
All blocks should derive from the Block<T> CRTP and not the gr::BlockModel type wrapper: 
The `BlockModel`-dervide blocks more often than not do not implement a correct `work(...)` behaviour such as including critical functionalities of tag->setting handling, tag forwarding, checking of in-out constraints etc. The direct `BlockModel` use should be slowly phased out.

core/test/qa_DynamicBlock.cpp
vimpostor: > check if there isn't already an existing sink that does the same (remove code duplication)

I checked and didn't find any. But it might still be worthwhile to move `DebugSink` from `qa_DynamicBlock` to `common_blocks.hpp`, as this kind of functionality is nice to have for automated tests. Maybe also rename it to something more suitable (`TestSink` ?) and generalize its interface so that it can debug/test multiple input values (not just one).

RalphSteinhagen: Agreed ... there are several multiple copies of these type of testing/debugging sinks that should be streamlined.

vimpostor: Maybe we could have a general `InspectBlock`, which acts as a pass-through block for debugging purposes that holds all the values coming in (so that they can be used in tests) and then pushes them to `n` output ports unaltered. If `n` is `0`, then it is like the sink described above.

PR #240, opened 2023-12-22
renamed some block-, file and public field names according to new NamingGuideline

RalphSteinhagen:


drslebedev (approved):
Thanks a lot for the PR!
The changes aligns with the new naming guidelines, which should also improve code readability.

PR can be merged after all tests passed.

PR #242, opened 2024-01-02
updated Tag[Source, Monitor, Sink]

RalphSteinhagen:
... and their re-use in different unit-tests.

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
@@ -289,8 +289,8 @@ public:
     Annotated<float, "sample rate", Doc<"signal sample rate">, Unit<"Hz">>           sample_rate = 1.f;
     Annotated<std::string, "signal name", Visible>                                   signal_name = "unknown signal";
     Annotated<std::string, "signal unit", Visible, Doc<"signal's physical SI unit">> signal_unit = "a.u.";
-    Annotated<T, "signal min", Doc<"signal physical min. (e.g. DAQ) limit">>         signal_min  = std::numeric_limits<T>::lowest();
-    Annotated<T, "signal max", Doc<"signal physical max. (e.g. DAQ) limit">>         signal_max  = std::numeric_limits<T>::max();
+    Annotated<float, "signal min", Doc<"signal physical min. (e.g. DAQ) limit">>     signal_min  = -1.0f;
+    Annotated<float, "signal max", Doc<"signal physical max. (e.g. DAQ) limit">>     signal_max  = +1.0f;
 
     PortIn<T, RequiredSamples<std::dynamic_extent, _listener_buffer_size>> in;
 
@@ -454,11 +454,11 @@ public:
     processBulk(std::span<const T> inData) noexcept {
         std::optional<property_map> tagData;
         if (this->input_tags_present()) {
-            assert(this->input_tags()[0].index == 0);
-            tagData = this->input_tags()[0].map;
+            assert(this->mergedInputTag().index == 0);
+            tagData = this->mergedInputTag().map;
             // signal info from settings overrides info from tags
             if (!_has_signal_info_from_settings) {
-                applySignalInfo(this->input_tags()[0].map);
+                applySignalInfo(this->mergedInputTag().map);
             }
         }
 
@@ -496,8 +496,8 @@ private:
         const auto rate_ = getProperty(properties, tag::SAMPLE_RATE.key(), float{});
         const auto name_ = getProperty(properties, tag::SIGNAL_NAME.key(), std::string());
         const auto unit_ = getProperty(properties, tag::SIGNAL_UNIT.key(), std::string());
-        const auto min_  = getProperty(properties, tag::SIGNAL_MIN.key(), T{});
-        const auto max_  = getProperty(properties, tag::SIGNAL_MAX.key(), T{});
+        const auto min_  = getProperty(properties, tag::SIGNAL_MIN.key(), float{});
+        const auto max_  = getProperty(properties, tag::SIGNAL_MAX.key(), float{});
         if (!rate_ && !name_ && !unit_ && !min_ && !max_) {
             return false;
         }
@@ -515,7 +515,7 @@ private:
                 listener->applySampleRate(sample_rate);
             }
             if (name_ || unit_ || min_ || max_) {
-                listener->setDataSetTemplate(DataSet<T>{ .signal_names = { signal_name }, .signal_units = { signal_unit }, .signal_ranges = { { signal_min, signal_max } } });
+                listener->setDataSetTemplate(DataSet<T>{ .signal_names = { signal_name }, .signal_units = { signal_unit }, .signal_ranges = { { T(signal_min), T(signal_max) } } });
             }
         }
 
@@ -543,7 +543,7 @@ private:
 
     void
     addListener(std::unique_ptr<AbstractListener> &&l, bool block) {
-        l->setDataSetTemplate(DataSet<T>{ .signal_names = { signal_name }, .signal_units = { signal_unit }, .signal_ranges = { { signal_min, signal_max } } });
+        l->setDataSetTemplate(DataSet<T>{ .signal_names = { signal_name }, .signal_units = { signal_unit }, .signal_ranges = { { T(signal_min), T(signal_max) } } });
         l->applySampleRate(sample_rate);
         if (block) {
             _listeners.push_back(std::move(l));
--- a/blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
@@ -40,48 +40,33 @@ ClockSource Documentation -- add here
     std::uint32_t                                                                n_samples_produced{ 0 };
     A<float, "avg. sample rate", Visible>                                        sample_rate = 1000.f;
     A<std::uint32_t, "chunk_size", Visible, Doc<"number of samples per update">> chunk_size  = 100;
-    std::thread                                                                  userProvidedThread;
+    std::shared_ptr<std::thread>                                                 userProvidedThread;
+    bool                                                                         verbose_console = true;
 
-    ~ClockSource() { stopThread(); }
-
-    [[maybe_unused]] bool
-    tryStartThread() {
-        if constexpr (useIoThread) {
-            return false;
+    void
+    start() {
+        if (verbose_console) {
+            fmt::println("starting {}", this->name);
         }
-        if (bool expectedThreadState = false; this->ioThreadShallRun.compare_exchange_strong(expectedThreadState, true, std::memory_order_acq_rel)) {
-            // mocks re-using a user-provided thread
-            fmt::print("mocking a user-provided io-Thread for {}\n", this->name);
-            std::atomic_store_explicit(&this->ioThreadShallRun, true, std::memory_order_release);
-            this->ioThreadShallRun.notify_all();
-            userProvidedThread = std::thread([this]() {
-                fmt::print("started user-provided thread\n");
-                for (int retry = 2; this->ioThreadShallRun.load() && retry > 0; --retry) {
-                    while (this->ioThreadShallRun.load()) {
-                        std::this_thread::sleep_until(nextTimePoint);
-                        // invoke and execute work function from user-provided thread
-                        if (this->invokeWork() == work::Status::DONE) {
-                            break;
-                        } else {
-                            retry = 2;
-                        }
-                    }
-                    // delayed shut-down in case there are more tasks to be processed
-                    std::this_thread::sleep_for(std::chrono::milliseconds(10));
-                }
-                this->stopThread();
-                fmt::print("stopped user-provided thread\n");
-            });
-            userProvidedThread.detach();
-            return true;
+        n_samples_produced = 0U;
+        tryStartThread();
+        if (verbose_console) {
+            fmt::println("started {}", this->name);
         }
-        return false;
     }
 
     void
-    stopThread() {
-        std::atomic_store_explicit(&this->ioThreadShallRun, false, std::memory_order_release);
-        this->ioThreadShallRun.notify_all();
+    stop() {
+        if (verbose_console) {
+            fmt::println("stop {}", this->name);
+        }
+        this->requestStop();
+        if constexpr (!useIoThread) {
+            if (verbose_console) {
+                fmt::println("joining user-provided {}joinable thread in block {}", userProvidedThread->joinable() ? "" : "non-", this->name);
+            }
+            userProvidedThread->join();
+        }
     }
 
     void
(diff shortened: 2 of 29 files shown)


core/include/gnuradio-4.0/Block.hpp
```diff
@@ -876,10 +875,30 @@ struct Block : protected std::tuple<Arguments...> {
...
+                             static_cast<double>(ports_status.in_min_samples) * ratio, static_cast<double>(ports_status.out_max_samples), numerator, denominator, ratio);
+                assert(false && "reduced min-required input sample needs to be <= max OutPort sample constraints");
+                return work::Status::ERROR;
+            }
+            if (static_cast<double>(ports_status.in_max_samples) * ratio < static_cast<double>(ports_status.out_min_samples)) { // TODO: convert to proper error message send to msgOut
+                fmt::println(stderr, "configuration error for block {}:  ports_status.in_max_samples * ratio {} > ports_status.out_min_samples {}, ratio (num {}/den {} = {})", name,
+                             static_cast<double>(ports_status.in_max_samples) * ratio, static_cast<double>(ports_status.out_min_samples), numerator, denominator, ratio);
+                assert(false && "reduced max-required input sample needs to be <= min OutPort sample constraints");
+                return work::Status::ERROR;
+            }
```
drslebedev, line 898: Does it makes sense to move all settings/parameter checking to some function like `checkSettings` and it should be called only if parameters are changed inside `settingsChanged`? Some additional ill-defined parameters checking can be also then added.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1400,32 +1441,28 @@ struct RegisterBlock {
...
-struct fmt::formatter<gr::work::Status> {
-    static constexpr auto
-    parse(const format_parse_context &ctx) {
-        const auto it = ctx.begin();
-        if (it != ctx.end() && *it != '}') throw format_error("invalid format");
-        return it;
-    }
+template<typename Function, typename Tuple, typename... Tuples>
+auto
+for_each_port(Function &&function, Tuple &&tuple, Tuples &&...tuples) {
```
drslebedev, line 1446: 👍

core/include/gnuradio-4.0/Port.hpp
```diff
@@ -845,76 +960,48 @@ class DynamicPort {
...
-    if (!port.isConnected() || port.tagReader().available() == 0) [[likely]] {
-        return std::numeric_limits<std::size_t>::max(); // default: no tags in sight
-    }
-    const auto tags    = port.tagReader().get();
-    const auto readPos = port.streamReader().position();
-
-    // find the smallest index of EOS tag, if no EOS tag is present then index = std::numeric_limits<std::size_t>::max()
-    std::size_t result = std::numeric_limits<std::size_t>::max();
-    // TODO: 'const auto tag' without reference is a workaround for gcc compiler
-    //  by unknown reasons only copy works, reference leads to unvalidated tag.map.end() iterator
```
drslebedev, line 899: Is this problem no longer valid because the logic was changed and only the first Tag is searched for?

core/test/qa_Block.cpp
```diff
@@ -263,7 +261,8 @@ const boost::ut::suite _stride_tests = [] {
         interpolation_decimation_test({ .n_samples =  100, .numerator =    101, .denominator =  101, .exp_in = 0 , .exp_out = 0, .exp_counter = 0 }, thread_pool);
 
         interpolation_decimation_test({ .n_samples =  100, .numerator = 5, .denominator = 11, .out_port_min = 10 , .out_port_max = 41, .exp_in = 88, .exp_out = 40, .exp_counter =  1 }, thread_pool);
-        interpolation_decimation_test({ .n_samples =  100, .numerator = 7, .denominator =  3, .out_port_min = 10 , .out_port_max = 10, .exp_in =  0, .exp_out =  0, .exp_counter =  0 }, thread_pool);
+// TODO: check with Semen if this test is possibly ill-defined and can never finish
+//        interpolation_decimation_test({ .n_samples =  100, .numerator = 7, .denominator =  3, .out_port_min = 10 , .out_port_max = 10, .exp_in =  0, .exp_out =  0, .exp_counter =  0 }, thread_pool);
```
drslebedev, line 265: As discussed privately, the test was intentionally ill-defined. One needs to think how to unit-test ill-defined settings.

drslebedev (approved):
Thanks a lot for a very nice PR! 
The general impression: a great job in refactoring and simplifying the code.
Updated tag processing looks good to me, as well as updated common blocks!

I added several comments for the PR when they are resolved the PR can be merged.

blocks/testing/include/gnuradio-4.0/testing/TagMonitors.hpp
```diff
@@ -59,128 +51,200 @@ mismatch_report(const IterType &mismatchedTag1, const IterType &mismatchedTag2,
...
     if (mismatchedTag1 != tags1.end()) {
-        mismatch_report(mismatchedTag1, mismatchedTag2, tags1.begin());
+        mismatch_report(mismatchedTag1, mismatchedTag2, tags1.begin(), ignoreKey);
         return false;
     }
     return true;
 }
 
-template<typename T, ProcessFunction UseProcessOne>
+template<typename T, ProcessFunction UseProcessOne = ProcessFunction::USE_PROCESS_BULK>
```
wirew0rm, line 86: naming: `UseProcessOne` -> `UseProcessVariant`, `ProcessFunctionVariant`...

wirew0rm (approved):
Looks good for the parts that I could follow completely, for Block and Port I didn't notice any problems, but I can't claim i followed though the whole code there.

I commented a few smaller things, nothing that looks like a major problem.

core/include/gnuradio-4.0/Port.hpp
RalphSteinhagen: Yes, this is fixed now. The primary issue seemed to be UB in the circular/mmap-ed buffer that does not work (by C++ design/standard) for non-trivially copyable data types. I will also inform the GR 3.X team (ping to @mormj) that this is also likely a UB bug in the existing GR 3.X and probably may reemerge also for fundamental types as the compiler will perform better/more aggressive optimisations (the double-mmap-ed version's end is nigh, and the <5% throughput benefit compared to the clean std-conforming implementation doesn't warrant the hassle much longer).

core/include/gnuradio-4.0/Block.hpp
RalphSteinhagen: Yes, it should be part of the next refactoring PR. I would leave it as is until we settled on the other items.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1058,49 +1058,90 @@ struct Block : protected std::tuple<Arguments...> {
...
+            }
+#ifdef _DEBUG
+            fmt::println("block {} - mark2 - in {} -> out {}", name, ports_status.in_samples, ports_status.out_samples);
+#endif
+            ports_status.in_samples = std::min(ports_status.in_samples, ports_status.nSamplesToEosTag); // clamp input until EOS tag
+            // TODO: recheck definitions of numerator and denominator
+            //   Provided we we define N := <numerator> and M := <denominator>, do we expect (assuming a simple in->out block) that
+            //  a) for N input samples M output samples are produced, or
+            //  b) for a given number of input samples X -> X * N/M output samples are produced.
+            //  Some of the logic conditions vary depending on whether we assume 'a)' or 'b)'.
```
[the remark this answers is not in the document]
RalphSteinhagen: OK, so we agree on the interpretation and that `b)` is presently implemented. 

This is a discussion to be had whether the variant `a)`  or `b)` is easier from an end-user perspective. I see you point for `b)`, at the same time `a)` would also allow a (perhaps) simpler interpretation and also -- at the same time -- allows to constrain the minimum/maximum required samples for each processing iteration w/o having to set these min / max configurations for each port separately. As discussed, there is some ill-defined cases to be identified where the resampling ratio (numerator/denominator) is incompatible/conflicts the port requirements. 

We should discuss this tomorrow how to best resolve these ambiguity and cases where the settings are overconstraint.

core/test/qa_Block.cpp
RalphSteinhagen: OK. Then my assessment was correct. Have to see how/where we handle this case: block, scheduler, ...elsewhere/not at all? -> tomorrow's discussion.

blocks/testing/include/gnuradio-4.0/testing/TagMonitors.hpp
RalphSteinhagen: I chose... `UseProcessVariant`

blocks/basic/test/qa_sources.cpp
```diff
@@ -70,23 +79,30 @@ const boost::ut::suite TagTests = [] {
         // TODO: last decimator/interpolator + stride addition seems to break the limiting the input samples to the min of available vs. n_samples-until next tags
         // expect(equal_tag_lists(src.tags, sink1.tags)) << "sink1 (USE_PROCESS_ONE) did not receive the required tags";
         // expect(equal_tag_lists(src.tags, sink2.tags)) << "sink2 (USE_PROCESS_BULK) did not receive the required tags";
+        if (verbose) {
+            fmt::println("finished ClockSource test w/ {}", useIoThreadPool ? "Graph/Block<T>-provided thread" : "user-provided thread");
+        }
     };
 
+    "ClockSource w/ Graph/Block<T> provided thread"_test = [&clockSourceTest] { clockSourceTest.template operator()<true>(false); };
```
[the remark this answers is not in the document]
RalphSteinhagen: I tried a different versions ... this one worked. If you have better options, please give it a try.

blocks/basic/test/qa_DataSink.cpp
```diff
@@ -399,14 +337,14 @@ const boost::ut::suite DataSinkTests = [] {
...
+        constexpr std::uint64_t kSamples = 200000;
 
         gr::Graph  testGraph;
-        auto      &src  = testGraph.emplaceBlock<Source<int32_t>>({ { "n_samples_max", kSamples } });
+        auto      &src  = testGraph.emplaceBlock<gr::testing::TagSource<int32_t>>({ { "n_samples_max", kSamples }, { "mark_tag", false } });
         const auto tags = std::vector<Tag>{ { 3000, { { "TYPE", "TRIGGER" } } }, { 8000, { { "TYPE", "NO_TRIGGER" } } }, { 180000, { { "TYPE", "TRIGGER" } } } };
         src.tags        = tags;
         auto &sink      = testGraph.emplaceBlock<DataSink<int32_t>>(
-                { { "name", "test_sink" }, { "signal_name", "test signal" }, { "signal_unit", "none" }, { "signal_min", int32_t{ 0 } }, { "signal_max", int32_t{ kSamples - 1 } } });
+                { { "name", "test_sink" }, { "signal_name", "test signal" }, { "signal_unit", "none" }, { "signal_min", float{ -2.0f } }, { "signal_max", float{ 2.0f } } });
```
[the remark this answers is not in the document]
RalphSteinhagen: done

blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
```diff
@@ -289,8 +289,8 @@ class DataSink : public Block<DataSink<T>> {
     Annotated<float, "sample rate", Doc<"signal sample rate">, Unit<"Hz">>           sample_rate = 1.f;
     Annotated<std::string, "signal name", Visible>                                   signal_name = "unknown signal";
     Annotated<std::string, "signal unit", Visible, Doc<"signal's physical SI unit">> signal_unit = "a.u.";
-    Annotated<T, "signal min", Doc<"signal physical min. (e.g. DAQ) limit">>         signal_min  = std::numeric_limits<T>::lowest();
-    Annotated<T, "signal max", Doc<"signal physical max. (e.g. DAQ) limit">>         signal_max  = std::numeric_limits<T>::max();
+    Annotated<float, "signal min", Doc<"signal physical min. (e.g. DAQ) limit">>     signal_min  = -1.0f;
+    Annotated<float, "signal max", Doc<"signal physical max. (e.g. DAQ) limit">>     signal_max  = +1.0f;
```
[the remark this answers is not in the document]
RalphSteinhagen: The signal_min and signal_max are defined for float only as part of the standardised default Tag definitions. In any case these are safe defaults and are overwritten by the data's actual min or max.

blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
```diff
@@ -40,48 +40,34 @@ ClockSource Documentation -- add here
     std::uint32_t                                                                n_samples_produced{ 0 };
     A<float, "avg. sample rate", Visible>                                        sample_rate = 1000.f;
     A<std::uint32_t, "chunk_size", Visible, Doc<"number of samples per update">> chunk_size  = 100;
-    std::thread                                                                  userProvidedThread;
+    std::shared_ptr<std::thread>                                                 userProvidedThread;
+    bool                                                                         debug = true;
```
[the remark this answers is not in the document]
RalphSteinhagen: done. changed to `verbose_console`. 

N.B. We should start thinking how to codify standard naming for settings (similar to the default tags).

RalphSteinhagen:
added the review comment. 

The mentioned issues of codifying tag, settings names and types should be done in a different PR/enhancements.

Similarly the interpretation of `nominator` , `denominator` and how to handle ill-defined block constraints and their compile-time & runtime checks.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -890,15 +909,20 @@ struct Block : protected std::tuple<Arguments...> {
...
+            if (ports_status.in_samples < in_min_wo_remainder) {
                 return work::Status::INSUFFICIENT_INPUT_ITEMS;
             }
-            ports_status.in_samples  = std::clamp(ports_status.in_samples, in_min_wo_reminder, in_max_wo_reminder);
+
+            if (in_min_wo_remainder < in_max_wo_remainder) {
+                ports_status.in_samples = std::clamp(ports_status.in_samples, in_min_wo_remainder, in_max_wo_remainder);
+            } else {
+                ports_status.in_samples = std::min(ports_status.in_samples, std::min(in_min_wo_remainder, in_max_wo_remainder));
+            }
```
drslebedev, line 925: I looked again to `std::clamp` docs and according to https://en.cppreference.com/w/cpp/algorithm/clamp: The behavior is undefined if the value of `lo` is greater than `hi`. Please check if my suggestion makes sense. With this change all tests passed.

```suggestion
            if (in_min_wo_remainder <= in_max_wo_remainder) {
                ports_status.in_samples = std::clamp(ports_status.in_samples, in_min_wo_remainder, in_max_wo_remainder);
            } else {
                return work::Status::ERROR;
            }
```

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1058,49 +1058,90 @@ struct Block : protected std::tuple<Arguments...> {
...
+        } else {                                                                         // end of kIsSourceBlock
+            ports_status.in_samples = std::min(ports_status.in_samples, requested_work); // clamp input to scheduler work constraint
+#ifdef _DEBUG
+            fmt::println("block {} - mark1 - in {} -> out {}", name, ports_status.in_samples, ports_status.out_samples);
+#endif
+            if constexpr (!Resampling::kEnabled) { // clamp input until (excluding) next tag unless the read-position is on the tag
+                // N.B. minimum size '1' because we could have two tags on adjacent samples.
+                ports_status.in_samples  = ports_status.nSamplesToNextTag == 0 ? std::min(ports_status.in_samples, std::max(ports_status.nSamplesToNextTagAfter, 1UZ))
+                                                                               : std::min(ports_status.in_samples, ports_status.nSamplesToNextTag);
+                ports_status.out_samples = ports_status.in_samples;
```
drslebedev, line 1070: It looks like this line can be removed because you update `ports_status.in_samples` directly after `if constexpr`, then `ports_status.out_samples` must be updated after as well.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1058,49 +1058,90 @@ struct Block : protected std::tuple<Arguments...> {
...
             if constexpr (Resampling::kEnabled) {
-                const bool isEOSTagPresent  = ports_status.in_samples_to_eos_tag < ports_status.in_samples;
-                const auto rasamplingStatus = doResampling();
-                if (rasamplingStatus != work::Status::OK) {
-                    if (rasamplingStatus == work::Status::INSUFFICIENT_INPUT_ITEMS && isEOSTagPresent) {
-                        state = LifeCycleState::STOPPED;
-                        updateInputAndOutputTags();
+                const auto resamplingStatus = doResampling();
+                if (resamplingStatus != work::Status::OK) {
+                    if (resamplingStatus == work::Status::INSUFFICIENT_INPUT_ITEMS || isEOSTagPresent) {
```
drslebedev, line 1104: I think there is a logic error here.
`isEOSTagPresent` is always false when reached this `if`.
if it's true `if (isEOSTagPresent || lifecycle::isShuttingDown(state))` is also true and it `return`s.

`isEOSTagPresent` should be separated for `Resampling::kEnabled`

PR #243, opened 2024-01-02
pmtv:map_t std::less<> follow-up update

RalphSteinhagen:
`pmtv::map_t` got a transparent comparator that `std::map<std::string, pmt_t, std::less<>>` required:
 * explicit comparators for fixed_string
 * DefaultTag

added corresponding unit-tests, removed now obsolete code, and added implicit equivalency for DefaultTag names such as `"XXX"` "and `"gr:XXX"`. N.B. the later is required for compatibility with SigMF.


jsallay (approved):
LGTM

PR #245, opened 2024-01-16
CI: adopt singleheader CI to updated filenames

wirew0rm:
This PR updates the single header generator configuration to adapt to the updated Header filenames and changed dependencies.
Also updates the readme with a new compiler explorer sample.

For this to be useful with compiler-explorer we'll have to look into compile times, the "sample" I added now is barely doing anything and still only compiles without timeout on `-O2`.


RalphSteinhagen (approved):
OK. 

* Minor changes `--sdt=...` -> `--std=...`
* update of copyright year 2022 -> 2024

PR #246, opened 2024-01-16
Frank/fixes

frankosterfeld:
Two small changes:

 - Enable gr-testing also when ENABLE_TESTING=OFF (so it can be used in gr-digitizers)
 - Use multi-threaded schedulers in qa_DataSink now that the issues seem to have been fixed.


RalphSteinhagen (approved):
OK. :+1:

PR #248, opened 2024-01-18
Add `property_map` as supported type to Settings and Transactions.

drslebedev:
- Add `property_map` as supported type to Settings and Transactions.
- Add `fmt::formatter` for `property_map`


RalphSteinhagen (approved):
LGTM! Thanks for adding the missing pmt `property_map` to the supported setting variable types. 

Sufficiently tested and can be merged as is once the CI passes for all compiler/architectures.

PR #250, opened 2024-01-22
UI integration tests/demo

RalphSteinhagen:
Added a small example/demo how UI widgets could be integrated using the existing GR `Block<T>` definitions and buffer handling. The key is that the blocks have the `BlockingIO<false>` attribute so that the `invokeWork()` function (and in turn the `process[One, Bulk](...)`) are called from within the UI thread. 

Peek 2024-01-22 20-09

Abbreviated example (full on is < 100 SLOCs):
```cpp
template<typename T>
struct ImChartMonitor : public Block<ImChartMonitor<T>, BlockingIO<false>> {
    PortIn<T>   in;
    // [...]
    HistoryBuffer<T>   _historyBufferX{ 1000 };
    HistoryBuffer<T>   _historyBufferY{ 1000 };
    HistoryBuffer<Tag> _historyBufferTags{ 1000 };

    // [...]

    constexpr void
    processOne(const T &input) noexcept {
        in.max_samples = 2 * sample_rate / 25; // poor man's rate limiter to avoid loosing/overriding samples when the UI is lagging behind.
        const float Ts = 1.0f / sample_rate;
        _historyBufferX.push_back(_historyBufferX[1] + Ts);
        _historyBufferY.push_back(input);

        if (this->input_tags_present()) { // received tag
            _historyBufferTags.push_back(this->mergedInputTag());
            _historyBufferTags[1].index = 0;
            this->_mergedInputTag.map.clear(); // TODO: provide proper API for clearing tags
        } else {
            _historyBufferTags.push_back(Tag(-1, property_map()));
        }
    }

    work::Status
    draw() noexcept {
        const work::Status status = this->invokeWork(); // calls work(...) -> processOne(...) (all in the same thread as this 'draw()'
        const auto [xMin, xMax]                    = std::ranges::minmax_element(_historyBufferX);
        const auto [yMin, yMax]                    = std::ranges::minmax_element(_historyBufferY);
        if (_historyBufferX.empty() || *xMin == *xMax || *yMin == *yMax) {
            return status; // buffer or axes' ranges are empty -> skip drawing
        }
        // invoke, for example, the ImGUI/ImPlot code here
        
        return status;
    }
};
```

Since both the `draw(...)` and `processOne(...)` are always called from the same thread, no further synchronisation is necessary and both function can share the same `HistoryBuffer<T>` instances.

The back-pressure and data reduction could/should be handled by dedicated Blocks that are handled by the regular GR processing graphs/scheduler and the UI element interact with the reducer/back-pressure properties (e.g. dropping samples when invisible) using the regular (WIP) `MsgPort msg[In,Out]`  API-

### DIFF
--- a/algorithm/include/gnuradio-4.0/algorithm/ImChart.hpp
+++ b/algorithm/include/gnuradio-4.0/algorithm/ImChart.hpp
@@ -319,8 +319,8 @@ public:
                 continue;
             }
 
-            const auto brailleColIndex = horAxisTransform::toScreen(x, axis_min_x, axis_max_x, horOffset, arrayWidth);
-            const auto brailleRowIndex = arrayHeight - verAxisTransform::toScreen(y, axis_min_y, axis_max_y, 0UZ, arrayHeight);
+            const auto brailleColIndex = horAxisTransform::toScreen(x, ValueType(axis_min_x), ValueType(axis_max_x), horOffset, arrayWidth);
+            const auto brailleRowIndex = arrayHeight - verAxisTransform::toScreen(y, ValueType(axis_min_y), ValueType(axis_max_y), 0UZ, arrayHeight);
             if (brailleRowIndex >= (arrayHeight - 1UZ) || brailleColIndex >= arrayWidth || (style == Style::Bars && brailleRowIndex >= horAxisPosY)) {
                 continue;
             }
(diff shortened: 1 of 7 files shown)


drslebedev (approved):
Thanks for the PR! It looks good to me. 
I really like this working concept and how straightforward it was to integrate UI. 
The PR can be merged.

PR #251, opened 2024-01-23
Changes in Block class as follow-up from #242

drslebedev:


RalphSteinhagen (approved):
Looks good and as discussed! :+1:

PR #254, opened 2024-01-25
Introducing message port handling to blocks and graphs

ivan-cukic:
- Added a default message port pair to gr::Block with example messages for setting properties, and lifetime control of blocks;
- Additional message port pair for gr::Graph to communicate with the blocks it owns -- used for message propagation into the graph and for exporting messages outward;
- Blocks handle the messages with (optionally) type-tagged ports -- when using named ports (or another way of port type differentiation, the processMessages handlers are chosen during compile-time and there is no need to check which port a message came on during runtime);
- Messages can be targeted to a specified unique_name. The '/' is used for separating unique_name of parent graph(s) and the block. '*' (or an empty target) is used to denote a message that is meant for all blocks;
- Since messages are arbitrary property maps, some convenience functions and definitions are provided;


core/include/gnuradio-4.0/PortTraits.hpp
```diff
@@ -50,6 +50,63 @@ using is_port = std::integral_constant<bool, is_port_v<Type>>;
...
+        return static_cast<typename T::value_type *>(nullptr);
+    } else {
+        meta::print_types<meta::message_type<"Is not a port or a collection of ports">, T>{};
+    }
+}
+
+template<typename T>
+using unwrap_port = std::remove_pointer_t<decltype(unwrap_port_helper<T>())>;
+
+struct port_kind {
```
frankosterfeld, line 68: `gr::traits::port::port_kind::Stream` reads  bit redundant, maybe rename to `kind`?

core/test/qa_DynamicBlock.cpp
```diff
@@ -14,7 +14,7 @@ struct fixed_source : public gr::Block<fixed_source<T>, gr::PortOutNamed<T, "out
 
     gr::work::Result
     work(std::size_t requested_work) {
-        auto &port   = gr::outputPort<0>(this);
+        auto &port   = gr::outputPort<0, gr::traits::port::port_kind::Stream>(this);
```
RalphSteinhagen, line 17: What about `gr::traits::port::port_kind::Stream` -> `gr::port::kind::Stream` or `gr::port_kind::Stream` for easier readability?

core/test/qa_Messages.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+template<typename T>
+struct CountSource : public gr::Block<CountSource<T>> {
+    gr::PortOut<T> out;
+    std::size_t    n_samples_max = 0;
+    std::size_t    count         = 0;
+
+    EventChecker *_eventChecker;
+
+    CountSource(EventChecker &eventChecker) : _eventChecker(std::addressof(eventChecker)) {}
```
RalphSteinhagen, line 92: Not ideal, since the CRTP constructor is not being called. We should refrain from declaring custom constructors/destructors in user-defined sources.

core/test/qa_Messages.cpp
```diff
@@ -0,0 +1,302 @@
...
+//  - test: settings update messages
+//
+//  - tests: custom message ports
+//  - processMessages overloads with 'different' message port types
+//
+//  - test: message processing when scheduler is stopped
+//
+//  idea: tests can use messages to trigger checks
+
+struct EventChecker {
```
RalphSteinhagen, line 58: If this important, move this to the `gr::testing::...` block definitions.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -130,54 +133,65 @@ invokeProcessOneWithOrWithoutOffset(T &node, std::size_t offset, const Us &...in
         return node.processOne(inputs...);
 }
 
-template<std::size_t Index, typename Self>
+template<std::size_t Index, auto Kind, typename Self>
```
RalphSteinhagen, line 136: could we constrain the `auto Kind` with a concept and/or use an explicite Type/NTTP for it?

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -130,54 +133,65 @@ invokeProcessOneWithOrWithoutOffset(T &node, std::size_t offset, const Us &...in
...
         using member_descriptor = traits::block::get_port_member_descriptor<Self, TRequestedPortType>;
         return member_descriptor()(*self);
     } else {
-        return std::get<TRequestedPortType>(*self);
+        return self->template getArgument<TRequestedPortType>();
     }
 }
 
-template<std::size_t Index, typename Self>
+template<std::size_t Index, auto Kind, typename Self>
```
RalphSteinhagen, line 148: ... similarly here 'auto' ?!?

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -282,6 +296,11 @@ concept PublishableSpan = std::ranges::contiguous_range<T> and std::ranges::outp
 
 static_assert(PublishableSpan<traits::block::detail::dummy_output_span<float>>);
 
+struct BuiltinMessagePortPair {
+    MsgPortInNamed<"__Builtin">  input;
```
RalphSteinhagen, line 300: Why the nested struct? Why not simply:

```cpp
MsgPortIn<>  msgIn;
MsgPortOut<>  msgOut;
```

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1356,6 +1431,72 @@ struct Block : protected std::tuple<Arguments...> {
             return workInternal(requested_work);
         }
     }
+
+    void
+    processMessages(MsgPortInNamed<"__Builtin"> &port, std::span<const Message> messages) {
+        fmt::print("{} got a message\n", self().unique_name);
```
RalphSteinhagen, line 1437: I guess this is a forgotten debug message. ;-)

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -591,16 +613,28 @@ class Edge {
...
     std::vector<std::function<ConnectionResult(Graph &)>> _connectionDefinitions;
-    std::vector<std::unique_ptr<BlockModel>>              _blocks;
     std::vector<Edge>                                     _edges;
 
+    std::vector<std::unique_ptr<BlockModel>> _blocks;
+
+    struct BlockData {
+        BlockModel                      *blockAccess = nullptr;
+        MsgPortOutNamed<"__ForChildren"> toChildMessagePort;
+        MsgPortInNamed<"__FromChildren"> fromChildMessagePort;
```
RalphSteinhagen, line 630: Why not drop the nesting and then re-use the existing msgIn/out definitions from `Block<T>`?

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -806,26 +864,30 @@ struct Graph {
...
+        }
     }
 
     // connect using the port name
 
     template<fixed_string sourcePortName, std::size_t sourcePortSubIndex, typename Source>
     [[nodiscard]] auto
     connect(Source &source) {
-        using source_output_ports             = typename traits::block::output_ports<Source>;
+        using source_output_ports             = typename traits::block::all_output_ports<Source>;
```
RalphSteinhagen, line 886: what about redining this from `traits::block::all_output_ports` to `(gr::)port::output::all<...>`?

RalphSteinhagen: This would keep it a bit more brief and similar also with e.g.
 * `port::output::message::all`
 * `port::output::stream::all`
  * ...
  
  :thinking:

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -864,6 +928,111 @@ struct Graph {
...
+        // fmt::print("==>>> Calling Graph's {} processMessages and passing to children, count = {}\n", this->unique_name, input.size());
+
+        auto splitFirstComponent = [](std::string_view name) -> std::optional<std::pair<std::string_view, std::string_view>> {
+            auto itSlash = std::ranges::find(name, '/');
+            if (itSlash == name.end()) return {};
+
+            return { { std::string_view(name.begin(), itSlash), std::string_view(itSlash + 1, name.end()) } };
+        };
+
+        auto notifyChild = [this](std::string_view childName, const auto &message) {
```
RalphSteinhagen, line 942: Wouldn't it be easier/better to notify every child with the same message and let the child decide whether the message is for the specific child? This way, we'd also have broadcasting-type messages.

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -444,12 +460,18 @@ class BlockWrapper : public BlockModel {
         return blockRef().work(requested_work);
     }
 
+    virtual void
+    processScheduledMessages() {
```
frankosterfeld, line 464: remove `virtual`, add `override`.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -130,54 +133,65 @@ invokeProcessOneWithOrWithoutOffset(T &node, std::size_t offset, const Us &...in
...
+    }
+    return outputPort<Index, traits::port::port_kind::Any, Self>(self);
 }
 
-template<typename Self>
+template<auto Kind, typename Self>
 [[nodiscard]] constexpr auto
 inputPorts(Self *self) noexcept {
-    return [self]<std::size_t... Idx>(std::index_sequence<Idx...>) { return std::tie(inputPort<Idx>(self)...); }(std::make_index_sequence<traits::block::input_ports<Self>::size>());
+    auto size = traits::block::ports_data<Self>::template for_kind<Kind>::input_ports::size;
```
frankosterfeld, line 183: Unused temporary (spams me with warnings)

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -994,6 +1035,38 @@ struct Block : protected std::tuple<Arguments...> {
...
+    constexpr void
+    processScheduledMessages() {
+        // fmt::print("processScheduledMessages for {}\n", unique_name);
+        auto processPort = [this]<PortLike TPort>(TPort &inPort) {
+            if constexpr (traits::block::can_processMessagesForPort<Derived, TPort>) {
+                const auto available = inPort.streamReader().available();
+                if (available > 0) {
+                    self().processMessages(inPort, inPort.streamReader().get(available));
+                }
+                inPort.streamReader().consume(available);
```
frankosterfeld, line 1047: These consume()s give me warnings because it's [[nodiscard]]

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -591,16 +613,28 @@ class Edge {
     }
 };
 
-struct Graph {
+class Graph : public gr::Block<Graph> {
```
frankosterfeld, line 616: The move assignment operator now gives me a sonar warning `Move assignment operator of ‘Graph’ is implicitly deleted because base class gr::Block<Graph> has deleted move assignment operator.`. opendigitizer's GnuRadioAcquisitionWorker relies on this (moving around graphs in/out of an optional)

core/test/qa_Messages.cpp
ivan-cukic: It calls Block(), which calls Block({}), but I can move it from the constructor..

core/include/gnuradio-4.0/Block.hpp
ivan-cukic: Ok

core/include/gnuradio-4.0/Block.hpp
ivan-cukic: - Easier to handle one pointer in BlockModel than two
- Can not accidentally be exposed grouped with other ports thus changing indices of other ports

core/include/gnuradio-4.0/Block.hpp
ivan-cukic: I added a TODO 'clean up debugging output' :)

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1356,6 +1431,72 @@ struct Block : protected std::tuple<Arguments...> {
...
+                    }
+                if constexpr (requires { self().resume(); })
+                    if (what == gr::message::control::Resume) {
+                        self().resume();
+                    }
+                if constexpr (requires { self().reset(); })
+                    if (what == gr::message::control::Reset) {
+                        self().reset();
+                    }
+
```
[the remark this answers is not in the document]
ivan-cukic: Meant more as an example than a real world use-case.

core/include/gnuradio-4.0/Graph.hpp
ivan-cukic: The separation is needed to differentiate between messages going in to those coming out of the Graph. If a block in a graph sends a message, we want it emitted to the graph's owner, not back to the Block.

core/include/gnuradio-4.0/Graph.hpp
ivan-cukic: This implementation does support multicast.
- `target=="*"` or `target=""` - sends a message to every block
- `target=="graph_unique_name/*"` - sends a message to every block in a graph
- `target=="graph_unique_name/block_unique_name"` or `target=="*/block_unique_name"` sends to a specific block (since the names are unique globally, not only per graph)
- (not supported now, but could be easily if we need it) `target=="**/..."` for multilevel Graph skipping

I'm not a fan of spam, so sending a message to everyone and then having most of them ignore it is not a good design IMO. This provides all the features with better control and no overhead of calling `O(n)` functions that will parse a string and then bail out when only one needs to be called (with n being the number of blocks)

core/include/gnuradio-4.0/Graph.hpp
ivan-cukic: Will see what can be done. If this can not be fixed (needs to see why Block is not moveable), GnuRadioAcquisitionWorker could use a unique_ptr<Graph> instead of optional.

core/include/gnuradio-4.0/Block.hpp
RalphSteinhagen: At the same time these message ports require a special handling compared to the other user-defined one. I'd recommend to use/implement them equal to the user-defined ones. This would remove a special case. 

Also, there is no functional need to mandatorially group two in/out message ports. The user may choose to connect only the input or only the output.

core/include/gnuradio-4.0/Block.hpp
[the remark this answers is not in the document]
RalphSteinhagen: OK. The lifecycle should be directly invoked by the scheduler -> graph -> blocks because these need to be synchronously processed (contract/guarantee) whereas the message ports are by designed asynchronous and their message processed at -- potentially -- undefined moment of time.

core/include/gnuradio-4.0/Graph.hpp
RalphSteinhagen: yes... forgot. These are the 'hidden' MsgPort that only the Graph uses to exchange data from/to it's children blocks. :+1:

core/include/gnuradio-4.0/Block.hpp
ivan-cukic: They are special. As we discussed before, user-defined message ports have user-defined handlers. The built-in message ports are the ports that we handle by default - that can be used for settings management etc.

The user-defined ports are not meant to be defined in pairs, but with the normal port syntax that we already have.

```
   PortMsgIn<> customMessageIn;
   PortMsgIn<> anotherMessageIn;
```

RalphSteinhagen: What about having a default handler for the msgIn that invokes the optional user handler?

core/include/gnuradio-4.0/Graph.hpp
frankosterfeld: I don't this resolved. Still get a warning.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -544,7 +585,39 @@ struct Block : protected std::tuple<Arguments...> {
         }
     }
 
-    Block(Block &&other) noexcept : std::tuple<Arguments...>(std::move(other)), _settings(std::move(other._settings)) {}
+    Block(Block &&other) noexcept
+        : std::tuple<Arguments...>(std::move(other))
+        , _settings(std::move(other._settings))
```
frankosterfeld, line 590: I get warnings here about initialization order

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -544,7 +585,39 @@ struct Block : protected std::tuple<Arguments...> {
...
+        // , state(std::move(other.state)) // atomic, not included in move operations
+        , msgIn(std::move(other.msgIn))
+        , msgOut(std::move(other.msgOut))
+        , ports_status(std::move(other.ports_status))
+        , _output_tags_changed(std::move(other._output_tags_changed))
+        , _mergedInputTag(std::move(other._mergedInputTag)) {}
+
+    Block &
+    operator=(Block &&other) noexcept {
+        auto tmp = std::move(other);
```
frankosterfeld, line 604: When I actually try to move-assign a graph, compilation fails in several places, including Port::operator=(Port &&other) (members wrongly prefixed with `_`), and below.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -544,7 +585,39 @@ struct Block : protected std::tuple<Arguments...> {
...
+
+        std::tuple<Arguments...>::swap(*this, *tmp);
```
frankosterfeld, line 606: `*tmp` -> `tmp`

blocks/testing/include/gnuradio-4.0/testing/FunctionBlocks.hpp
```diff
@@ -0,0 +1,108 @@
...
+#include <fmt/ranges.h>
+
+#include <gnuradio-4.0/Block.hpp>
+#include <gnuradio-4.0/reflection.hpp>
+#include <gnuradio-4.0/Tag.hpp>
+
+namespace gr::testing {
+
+template<typename T>
+struct FunctionSource : public gr::Block<FunctionSource<T>> {
```
RalphSteinhagen, line 14: These classes need some documentation w.r.t. what they do and what they are supposed to be used for.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -399,6 +413,19 @@ struct Block : protected std::tuple<Arguments...> {
     using StrideControl              = ArgumentsTypeList::template find_or_default<is_stride, Stride<0UZ, true>>;
     constexpr static bool blockingIO = std::disjunction_v<std::is_same<BlockingIO<true>, Arguments>...> || std::disjunction_v<std::is_same<BlockingIO<false>, Arguments>...>;
 
+    template<typename T>
+    auto &
+    getArgument() {
```
RalphSteinhagen, line 418: Do we need this? seems to to be an alias for the tupple get...

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -783,18 +843,18 @@ struct Block : protected std::tuple<Arguments...> {
                         };
                         (consume_port(input_port), ...);
                     },
-                    inputPorts(&self));
+                    inputPorts<traits::port::kind::Stream>(&self));
```
RalphSteinhagen, line 846: I like the trait concept.... the `traits::` and `::kind::` afterwards is a bit redundant.

What about naming this: `traits::port::Stream` (my preference) or `port::kind::Stream`

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -994,6 +1054,38 @@ struct Block : protected std::tuple<Arguments...> {
...
+        };
+        processPort(msgIn);
+        for_each_port(processPort, inputPorts<traits::port::kind::Message>(&self()));
+    }
+
+    template<typename TMessage>
+    void
+    emitMessage(TMessage &&message) {
+        emitMessage(msgOut, std::forward<TMessage>(message));
+    }
```
RalphSteinhagen, line 1081: Please remove this... can be derived from the one below which makes it also a bit more clear which port the message is being emitted.

core/test/qa_Messages.cpp
```diff
@@ -0,0 +1,286 @@
+#include "gnuradio-4.0/basic/common_blocks.hpp"
+#include <boost/ut.hpp>
+
+#include <gnuradio-4.0/Scheduler.hpp>
+#include <magic_enum.hpp>
+#include <magic_enum_utility.hpp>
```
RalphSteinhagen, line 6: These are some general comments regarding the unit tests:
  * It's good that you refactored and moved the blocks to the blocklib! :+1: 
  * The examples (or blocklib) should include some specific user-level examples how to use the MsgPort in a
    * **fan-in configuration `M->1`**: i.e. multiple `M` producer feeding a single message input port
    * **fan-out configuration `1->N`**: i.e. one message output port of one block connected to `N` message input ports of multiple  other blocks
    * **multicast (fan-in-out) `M -> N`**: where a group  of `M` message output ports/blocks is connected to `N` message input ports/blocks. 
  * brief example how the scheduler implicitly calls the block's process message callback functions as part of the work-loop (assume scheduler is in a 'running' state)
  * brief example how the scheduler could be manually triggered to process messages (assume scheduler is not in a 'running' state).

core/include/gnuradio-4.0/Block.hpp
ivan-cukic: The fact that the Block is a tuple is an internal thing (we should at some point see what should be public/protected/private in GR), this is API to access the arguments from outside of the block.

RalphSteinhagen: Could this be linked to the change of `struct Block` -> `class Block` in this PR?

core/test/qa_DynamicBlock.cpp
```diff
@@ -26,12 +26,12 @@ struct fixed_source : public gr::Block<fixed_source<T>, gr::PortOutNamed<T, "out
 };
 
 static_assert(gr::BlockLike<fixed_source<int>>);
-static_assert(gr::traits::block::input_ports<fixed_source<int>>::size() == 0);
-static_assert(gr::traits::block::output_ports<fixed_source<int>>::size() == 1);
+static_assert(gr::traits::block::stream_input_ports<fixed_source<int>>::size() == 0);
```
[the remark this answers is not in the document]
ivan-cukic: This is a trait - this is not a user function like inputPort. If you want to remove the traits namespace in general, I'd maybe be fine with that, but I'd keep that separate from this PR. (and it should be done for everything, not for one meta-function/group of meta-functions)

core/include/gnuradio-4.0/Block.hpp
ivan-cukic: I've added ANY to PortType like we have for Direction, and removed `kind`

core/include/gnuradio-4.0/Block.hpp
ivan-cukic: No, the inheritance from tuple was protected before this change. The struct -> class change just influences the default access if none specified.

In the case of block, it has public...private, so it should be a class and private by default.

For tuple-inheritance, it should even (IMO) be private as it is an implementation detail and not something that people should rely on.

RalphSteinhagen (approved):
Thanks for the tremendous amount of work! This looks good to me and we discussed most of the remaining things online and offline. Main thing is the Graph connect API defining the edges and the fan-in, fan-out, multicast mechanics that includes the port constraints for the circular buffers as part of the connection phase. 

However, these should be tackled in follow-up PRs as the focus of this PR was to make the message ports usable from an UI integration point-of-view -- which it does. 

Thanks again @ivan-cukic!

PR #256, opened 2024-01-30
Make the FFT block work again

giucam:


drslebedev (commented):
Thanks for the PR. I appreciate your effort in working on this. However, it appears that this part of the code includes known issues that are already on our TODO list for refactoring.

I've noticed that your changes are causing unit tests to fail. The root of the issue seems to be that isEOSTagPresent consistently is false within this section of the code. Consequently, the if (isEOSTagPresent) condition is never met, and the corresponding block of code is never executed.

I suggest that we (@RalphSteinhagen , @wirew0rm ) need to refactor or fix this part of the code.

RalphSteinhagen:
closing as stale/obsoleted.

PR #259, opened 2024-02-01
Use the internal port definition to detect messaging ports

ivan-cukic:


RalphSteinhagen (approved):
LGTM! :+1: 

As discussed during this morning's meeting. Can be merged as is if the CI passes.

PR #260, opened 2024-02-01
Scheduler: add function to test whether a MT scheduler is still running

frankosterfeld:
Makes it possible to check whether the scheduler finished, without blocking.


RalphSteinhagen (approved):
Ok

PR #244, opened 2024-01-11
Add an API to iterate the graph once

giucam:
This PR adds new API to the schedulers to iterate the graph and return the control to the caller, without waiting for it to finish. This is needed by opendigitizer to continuously update when drawin the UI. (https://github.com/fair-acc/opendigitizer/pull/141)

PR #262, opened 2024-02-05
Implement HttpBlock

vimpostor:
This implements the `HttpBlock` as described in #236.

The WebAssembly implementation took quite some jumping through hoops to get right.
There are a lot of quirks with regard to threading in emscripten, that make the implementation a minefield of risking getting stuck in infinite loops, because the emscripten thread for various reasons is unable to make any progress.
Unfortunately we cannot use `ASYNCIFY`, because that is incompatible with exceptions support ([that are already enabled](https://github.com/fair-acc/graph-prototype/blob/530ce6298a6138140de5ac6e05eae95703cdc47f/CMakeLists.txt#L84)).

There is also this fun [emscripten regression](https://github.com/emscripten-core/emscripten/issues/8234#issuecomment-1333548834), which means it's currently impossible to cancel asynchronously running fetch requests.

For the unit test the httplib dummy server obviously doesn't work with emscripten, so the [--pre-js](https://emscripten.org/docs/tools_reference/emcc.html#emcc-pre-js) mechanism is used to implement another one with node http, that is automatically injected into the runtime.
Note that in order for the unit test to run, `node` needs to be able to find `xhr2`, so:

```bash
npm install xhr2
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,8 @@ if (EMSCRIPTEN)
             -fwasm-exceptions
             -pthread
             "SHELL:-s PTHREAD_POOL_SIZE=30"
+            "SHELL:-s FETCH=1"
+            "SHELL:-s ASSERTIONS=1"
     )
 endif ()
 
@@ -136,7 +138,13 @@ FetchContent_Declare(
         GIT_TAG v0.2.0
 )
 
-FetchContent_MakeAvailable(fmt pmt ut yaml-cpp vir-simd)
+FetchContent_Declare(
+    httplib
+    GIT_REPOSITORY https://github.com/yhirose/cpp-httplib.git
+    GIT_TAG v0.15.1
+)
+
+FetchContent_MakeAvailable(fmt pmt ut yaml-cpp vir-simd httplib)
 
 add_library(pmtv INTERFACE)
 target_include_directories(pmtv INTERFACE ${pmt_SOURCE_DIR}/include/)
--- a/blocks/basic/include/gnuradio-4.0/basic/common_blocks.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/common_blocks.hpp
@@ -10,22 +10,6 @@
 #include <gnuradio-4.0/Graph.hpp>
 #include <gnuradio-4.0/reflection.hpp>
 
-// A convenience class to make writing unit tests easier.
-// This sink allows to inspect the input port values as a class member.
-template<typename T>
-class InspectSink : public gr::Block<InspectSink<T>> {
-public:
-    T             value{};
-    gr::PortIn<T> in;
-
-    constexpr void
-    processOne(T val) {
-        value = val;
-    }
-};
-
-ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T), (InspectSink<T>), value, in);
-
 template<typename T>
 class builtin_multiply : public gr::Block<builtin_multiply<T>> {
 public:
--- a/blocks/basic/include/gnuradio-4.0/basic/common_blocks.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/common_blocks.hpp
@@ -10,6 +10,22 @@
 #include <gnuradio-4.0/Graph.hpp>
 #include <gnuradio-4.0/reflection.hpp>
 
+// A convenience class to make writing unit tests easier.
+// This sink allows to inspect the input port values as a class member.
+template<typename T>
+class InspectSink : public gr::Block<InspectSink<T>> {
+public:
+    T             value{};
+    gr::PortIn<T> in;
+
+    constexpr void
+    processOne(T val) {
+        value = val;
+    }
+};
+
+ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T), (InspectSink<T>), value, in);
+
 template<typename T>
 class builtin_multiply : public gr::Block<builtin_multiply<T>> {
 public:
--- a/blocks/http/CMakeLists.txt
+++ b/blocks/http/CMakeLists.txt
@@ -1,0 +1,7 @@
+add_library(gr-http INTERFACE)
+target_link_libraries(gr-http INTERFACE gnuradio-core gnuradio-algorithm httplib::httplib)
+target_include_directories(gr-http INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/> $<INSTALL_INTERFACE:include/>)
+
+if (ENABLE_TESTING)
+    add_subdirectory(test)
+endif ()
--- a/blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
+++ b/blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
@@ -1,0 +1,287 @@
+#ifndef GNURADIO_HTTP_BLOCK_HPP
+#define GNURADIO_HTTP_BLOCK_HPP
+
+#include <gnuradio-4.0/Block.hpp>
+#include <gnuradio-4.0/Graph.hpp>
+#include <gnuradio-4.0/reflection.hpp>
+#include <pmtv/pmt.hpp>
+#include <semaphore>
+
+#ifdef __EMSCRIPTEN__
+#include <emscripten/emscripten.h>
+#include <emscripten/fetch.h>
+#include <emscripten/threading.h>
+#else
+#include <httplib.h>
+#endif
+
+using namespace gr;
+using namespace std::chrono_literals;
+
+namespace gr::http {
+
+enum class RequestType {
+    GET,
+    SUBSCRIBE,
+    POST,
+};
+
+using HttpBlockDoc = Doc<R""(
+The HttpBlock allows to use the responses from HTTP APIs (e.g. REST APIs) as the value for this block's output port.
+The block can be used either on-demand to do single requests, or can use long polling to subscribe to an event stream.
+The result is provided on a single output port as a map with the following keys:
+- status: The HTTP status code, usually 200 on success
+- raw-data: The data of the response
+- mime-type: The mime-type of the response
+)"">;
+
+template<typename T>
+class HttpBlock : public gr::Block<HttpBlock<T>, BlockingIO<false>, HttpBlockDoc> {
+private:
+    // used for queuing GET responses for the consumer
+    std::queue<pmtv::map_t> _backlog;
+    std::mutex              _backlog_mutex;
+
+    std::shared_ptr<std::thread> _thread;
+    std::atomic_size_t           _pendingRequests = 0;
+    std::atomic_bool             _shutdownThread  = false;
+    std::binary_semaphore        _ready{ 0 };
+
+#ifndef __EMSCRIPTEN__
+    std::unique_ptr<httplib::Client> _client;
+#endif
+
+#ifdef __EMSCRIPTEN__
+    void
+    queueWorkEmscripten(emscripten_fetch_t *fetch) {
+        pmtv::map_t result;
+        result["mime-type"] = "text/plain";
+        result["status"]    = static_cast<int>(fetch->status);
+        result["raw-data"]  = std::string(fetch->data, static_cast<std::size_t>(fetch->numBytes));
+
+        queueWork(result);
+    }
+
+    void
+    onSuccess(emscripten_fetch_t *fetch) {
+        queueWorkEmscripten(fetch);
+        emscripten_fetch_close(fetch);
+    }
+
+    void
+    onError(emscripten_fetch_t *fetch) {
+        // we still want to queue the response, the statusCode will just not be 200
+        queueWorkEmscripten(fetch);
+        emscripten_fetch_close(fetch);
+    }
+
+    void
+    doRequestEmscripten() {
+        emscripten_fetch_attr_t attr;
+        emscripten_fetch_attr_init(&attr);
+        if (type == RequestType::POST) {
+            strcpy(attr.requestMethod, "POST");
+            if (!parameters.empty()) {
+                attr.requestData     = parameters.c_str();
+                attr.requestDataSize = parameters.size();
+            }
+        } else {
+            strcpy(attr.requestMethod, "GET");
+        }
+
+        // this is needed so that we can call into member functions again, when we receive the Fetch callback
+        attr.userData = this;
+
+        attr.attributes = EMSCRIPTEN_FETCH_LOAD_TO_MEMORY;
+        attr.onsuccess  = [](emscripten_fetch_t *fetch) {
+            auto src = static_cast<HttpBlock<T> *>(fetch->userData);
+            src->onSuccess(fetch);
+        };
+        attr.onerror = [](emscripten_fetch_t *fetch) {
+            auto src = static_cast<HttpBlock<T> *>(fetch->userData);
+            src->onError(fetch);
+        };
+        const auto target = url + endpoint;
+        std::ignore       = emscripten_fetch(&attr, target.c_str());
+    }
+
+    void
+    runThreadEmscripten() {
+        if (type == RequestType::SUBSCRIBE) {
+            while (!_shutdownThread) {
+                // long polling, just keep doing requests
+                std::thread thread{ &HttpBlock::doRequestEmscripten, this };
+                thread.join();
+            }
+        } else {
+            while (!_shutdownThread) {
+                while (_pendingRequests > 0) {
+                    _pendingRequests--;
+                    std::thread thread{ &HttpBlock::doRequestEmscripten, this };
+                    thread.join();
+                }
+                _ready.acquire();
+            }
+        }
+    }
+#else
+    void
+    runThreadNative() {
+        _client = std::make_unique<httplib::Client>(url);
+        _client->set_follow_location(true);
+        if (type == RequestType::SUBSCRIBE) {
+            // it's long polling, be generous with timeouts
+            _client->set_read_timeout(1h);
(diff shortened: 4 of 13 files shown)


CMakeLists.txt
```diff
@@ -84,6 +84,9 @@ if (EMSCRIPTEN)
             -fwasm-exceptions
             -pthread
             "SHELL:-s PTHREAD_POOL_SIZE=30"
+            "SHELL:-s FETCH=1"
+            "SHELL:-s ASSERTIONS=1"
+            "SHELL:-s USE_PTHREADS"
```
frankosterfeld, line 89: I wonder what difference these make (Are they required for compilation? or correct behavior), and if these are something we should already have enabled independent of the HttpBlock?

blocks/basic/include/gnuradio-4.0/basic/common_blocks.hpp
```diff
@@ -10,6 +10,22 @@
 #include <gnuradio-4.0/Graph.hpp>
 #include <gnuradio-4.0/reflection.hpp>
 
+// A convenience class to make writing unit tests easier.
+// This sink allows to inspect the input port values as a class member.
+template<typename T>
+class InspectSink : public gr::Block<InspectSink<T>> {
```
frankosterfeld, line 16: This should be moved to gr::testing (and we should review the various test sinks at some point, there might be room for unification...)

blocks/http/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+FetchContent_Declare(
+    httplib
+    GIT_REPOSITORY https://github.com/yhirose/cpp-httplib.git
+    GIT_TAG v0.15.1
+)
+FetchContent_MakeAvailable(httplib)
```
frankosterfeld, line 6: I'd rather keep this in the top-level CMakeLists.txt so its harder to miss.
Also another dependency we have to keep in sync with opencmw-cpp (but I guess that cannot be avoided...)

blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
```diff
@@ -0,0 +1,295 @@
...
+        const auto target = url + endpoint;
+        std::ignore       = emscripten_fetch(&attr, target.c_str());
+    }
+
+    void
+    runThreadEmscripten() {
+        if (type == RequestType::SUBSCRIBE) {
+            while (!_shutdownThread) {
+                // long polling, just keep doing requests
+                std::thread thread{ &HttpBlock::doRequestEmscripten, this };
```
frankosterfeld, line 113: couldn't this reuse threads, instead of creating a new one each time?

blocks/http/test/qa_HttpBlock.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+#include <gnuradio-4.0/http/HttpBlock.hpp>
+
+template<typename T>
+struct fixed_source : public gr::Block<fixed_source<T>, gr::PortOutNamed<T, "out">> {
+    using super_t = gr::Block<fixed_source<T>, gr::PortOutNamed<T, "out">>;
+    T value = 1;
+
+    gr::work::Result
+    work(std::size_t requested_work) {
```
drslebedev, line 23: Can it be implemented using `processOne` or `processBulk` function?

blocks/http/test/qa_HttpBlock.cpp
```diff
@@ -0,0 +1,215 @@
...
+        source.trigger();
+        httpBlock.processScheduledMessages();
+        pmtv::map_t response;
+        while (response.empty()) {
+            std::this_thread::sleep_for(1ms);
+            std::ignore = source.work(1UZ);
+            auto work   = httpBlock.work(1UZ);
+            work        = sink.work(1UZ);
+            response    = sink.value;
+        }
```
drslebedev, line 80: I saw that you use while loops for all unit tests to run the graph. Will it work with the standard `Scheduler`? Why don't you use it?

CMakeLists.txt
vimpostor: - `FETCH`: This is needed so that the emscripten fetch API can be used, this is also already enabled [in opencmw](https://github.com/fair-acc/opencmw-cpp/blob/a7a7c5c319b93ddfaf160893665a011d2d88bff8/CMakeLists.txt#L56)
- `ASSERTIONS: This is not strictly needed, but it is useful for debugging. This causes the printed stacktraces to be more meaningful (still not anywhere close to the debugging capabilities with a native build)
- `USE_PTHREADS`: This enables PTHREAD support, but apparently the `-pthread` flag above is already enough. I have removed this again.

blocks/basic/include/gnuradio-4.0/basic/common_blocks.hpp
vimpostor: Done

blocks/http/CMakeLists.txt
vimpostor: Agreed and done.
Keeping them in sync should not be too much of a problem, from what I can tell cpp-httplib does not break API often.

blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
vimpostor: That's what I attempted first and unfortunately this is not possible. :( There is some strange interplay with the emscripten runtime doing the request, which will cause an infinite loop right here, unless we do it in a separate thread (that we explicitly `join()` on).

blocks/http/test/qa_HttpBlock.cpp
vimpostor: Done.

blocks/http/test/qa_HttpBlock.cpp
vimpostor: Yes it should just work with a scheduler.
The reason that I don't use it here, is because it allows me to have more control over the execution in the graph. 
Furthermore it means I don't have to rely on the graph execution to stop automatically due to forward propagation of blocks turning into the `STOPPED` state.
Once #244 is merged, I might reconsider using the scheduler for tests though ;)

blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
```diff
@@ -0,0 +1,295 @@
...
+    void
+    queueWork(const pmtv::map_t &item) {
+        {
+            std::lock_guard lg{ _backlog_mutex };
+            _backlog.push(item);
+        }
+        const auto work = this->invokeWork();
+        if (work == work::Status::DONE) {
+            std::atomic_store_explicit(&this->_state, lifecycle::State::REQUESTED_STOP, std::memory_order_release);
+            this->_state.notify_all();
```
drslebedev, line 179: Here you can use newly introduced `StateMachine`:
```cpp
bool ok = this->changeStateTo(gr::lifecycle::State::REQUESTED_STOP).has_value();
```

or just 
``cpp
this->requestStop()
``
See https://github.com/fair-acc/graph-prototype/pull/261 for more details.

blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
```diff
@@ -0,0 +1,295 @@
...
+            _client->stop();
+#endif
+            if (t->joinable()) {
+                t->join();
+            }
+            _shutdownThread = false;
+            delete t;
+
+            std::atomic_store_explicit(&this->_state, lifecycle::State::STOPPED, std::memory_order_release);
+            this->_state.notify_all();
```
drslebedev, line 214: Also here one can use `StateMachine` methods.
```cpp
bool ok = this->changeStateTo(gr::lifecycle::State::STOPPED).has_value();
```

The advantage is that it checks if transition from one state to another is possible.

blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
vimpostor: Thanks and done.

blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
vimpostor: Done

blocks/http/test/qa_HttpBlock.cpp
RalphSteinhagen: I strongly recommend and request to use the scheduler interface, as this ensures that the block is ready for production use, which invariably relies on a scheduler. We've already encountered scenarios, such as with the `Selector` block, where its bespoke `work()` iteration method in the unit-tests falters in real-world applications. Moreover, the re-implementation of a block's low-level `work()` function is, or ought to be, strongly discouraged, given that many users fail to implement other critical invariants such as the state machine, tag propagation, sampling ratio, etc.

It's pertinent to note that there isn't a singular scheduler implementation; rather, the aim is for users to feel empowered and encouraged to develop their own. Nonetheless, for unit tests, it's imperative that the blocks at least function with the `gr::scheduler::Simple` implementation, which stands as one of the two compliant implementations.

Given the reasons outlined above, it's improbable that we'll proceed with merging [#244](https://github.com/fair-acc/graph-prototype/pull/244), as it represents a workaround and overlooks many of the invariants that both the scheduler and block implementations are required to fulfil.

drslebedev: Do I understand it correctly that the main problem now which prevent you from usage of `Scheduler` is that you can not stop scheduler execution? The execution is stopped using `END_OF_STREAM` tag, which should be publish by source. In your case if you use `Scheduler`, it will run forever because `EOF` tag is not published.

Here is one of the simple example of source:
```cpp
template<typename T>
struct CountSource : public gr::Block<CountSource<T>> {
    gr::PortOut<T> out{};
    int            count{ 0 };
    int            nSamples{ 1024 };

    constexpr T
    processOne() {
        count++;
        if (count >= nSamples) {
            this->requestStop(); 
        }
        return static_cast<T>(count - 1); // -1 to start from 0
    }
};
```

blocks/http/test/qa_HttpBlock.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+        // change endpoint to something invalid
+        std::ignore = httpBlock.settings().set({ { "endpoint", "/does-not-exist" } });
+        std::ignore = httpBlock.settings().applyStagedParameters();
+        response.clear();
+        source.trigger();
+        httpBlock.processScheduledMessages();
+        while (response.empty()) {
+            std::this_thread::sleep_for(1ms);
+            std::ignore = source.work(1UZ);
```
drslebedev, line 85: Is `auto work = httpBlock.work(1UZ);` missing here?

blocks/http/test/qa_HttpBlock.cpp
drslebedev: Expanding on the previous discussion:

Regarding the unit tests: the current implementation involves the trigger mechanism using input port of `HttpBlock`. One should use `ClockSource` block or a block with similar functionality then as input. This would allow us to define a specific time frame, after which an `EOS` tag is published. Specifically, in the case of `ClockSource`, this could be effectively managed by setting parameters like `n_samples_max` and `sample_rate`. Integrating a `Scheduler` would ensure the process halts upon reaching the predefined time limit.

This leads to a broader question: Is an input port absolutely necessary for `HttpBlock`? Alternatively, could we design the `HttpBlock` to function autonomously as a source block, issuing the `EOS` tag independently? Start/Stop mechanism can be then done via message port.

blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
```diff
@@ -0,0 +1,293 @@
...
+#endif
+    }
+
+    void
+    startThread() {
+        lifecycle::State expectedThreadState = lifecycle::State::INITIALISED;
+        if (!this->_state.compare_exchange_strong(expectedThreadState, lifecycle::State::RUNNING, std::memory_order_acq_rel)) {
+            return;
+        }
+        this->_state.notify_all();
```
drslebedev, line 198: I think this if statement is not needed and always return when using `Scheduler`. `_state` is always `RUNNING` at this point.

blocks/http/test/qa_HttpBlock.cpp
drslebedev: @vimpostor, do you have any updates? After reviewing your pull request again and discussing it with Ralph, we have some suggestions on how it might be integrated with the Scheduler. Essentially, you could implement a sink that stops the scheduler under certain conditions. Here's an outline of the code approach:

Firstly, you'll need to create a custom sink block as follows:
```cpp
template<typename T>
class HttpTestSink : public gr::Block<HttpTestSink<T>> {
public:
    gr::PortIn<T>         in;
    T                     value{};
    std::function<void()> stopFunc;

    constexpr void
    processOne(T val) {
        if (val.empty()) {
            stopFunc();
        } else {
            value = val;
        };
    }
};

ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T), (HttpTestSink<T>), in);
```
Then, set up the Graph, Scheduler, and execute it as shown below:
```cpp
gr::Graph graph;
auto     &source    = graph.emplaceBlock<fixed_source<uint8_t>>();
auto     &httpBlock = graph.emplaceBlock<http::HttpBlock<uint8_t>>("http://localhost:8080", "/echo");
auto     &sink      = graph.emplaceBlock<HttpTestSink<pmtv::map_t>>();

expect(eq(ConnectionResult::SUCCESS, graph.connect<"out">(source).template to<"in">(httpBlock)));
expect(eq(ConnectionResult::SUCCESS, source.msgOut.connect(httpBlock.msgIn)));
expect(eq(ConnectionResult::SUCCESS, graph.connect<"out">(httpBlock).template to<"in">(sink)));

auto sched    = gr::scheduler::Simple<>(std::move(graph), threadPool);
sink.stopFunc = [&sched]() { sched.stop(); };

source.trigger();
httpBlock.processScheduledMessages();
sched.runAndWait();
pmtv::map_t response = sink.value;
expect(eq(std::get<std::string>(sink.value.at("raw-data")), "Hello world!"sv));
```


Also as I wrote in the comment to the code. You have to remove these code from `Http::startThread()` because `_state` is always `RUNNING` at this point:
```cpp
lifecycle::State expectedThreadState = lifecycle::State::INITIALISED;
if (!this->_state.compare_exchange_strong(expectedThreadState, lifecycle::State::RUNNING, 
   std::memory_order_acq_rel)) {
   return;
}
this->_state.notify_all();
```

I've tested this code on my machine, and the test passed successfully.

Alternative Approach. As previously mentioned, another option could involve creating a Source with a timeout mechanism, namely send EOS tag after timeout.

I hope this clarifies our suggestions. Looking forward to your thoughts and any further updates you might have.

vimpostor:
@drslebedev Thank you very much for the detailed response, that information really helped me understand your scheduler approach better.

I am trying to adapt the tests now and I have the parts working where the scheduler only runs once.
However, when I want to restart the scheduler after the `settingsChanged()` in the first test, I get a weird problem where `processOne()` of the `HttpBlock` never gets called again, despite the graph running and all connections still existing.

To restart the graph I simply call `Scheduler::reset()` to bring back all blocks into the initialized state, then I call `Scheduler::runAndWait()` again like before. Is there anything else one needs to call to allow the scheduler to run the graph another time after it is finished?

drslebedev:
It looks like `Scheduler::reset()` might not be fully resetting all necessary components. My recommendation is to instantiate a new `Scheduler` for each test case. Given that the graphs in your tests are nearly identical, encapsulating the graph creation and related setup into a function that returns either a `Graph` or` Scheduler` (just see what fits better). This approach would help ensure a clean state for each test.

blocks/http/test/qa_HttpBlock.cpp
vimpostor: Yes, but N/A after the scheduler rewrite.

blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
vimpostor: Indeed, I removed it.

vimpostor:
@drslebedev I finished the tests rewrite using the scheduler. There were again some emscripten peculiarities (for some reason, the pre.js mechanism was not compatible with the setup of starting and stopping the server in each unit test), but everything should work now.

If the CI agrees, this should be ready for review again.

> Given that the graphs in your tests are nearly identical, encapsulating the graph creation and related setup into a function that returns either a Graph or Scheduler

I tried that, but except for the graph connections there is no boilerplate to be saved: The sink has to be accessible from within the test, so we cannot just encapsulate it in a `createGraph()` function. The `HttpBlock` instantiation differs in each unit test, so we can also not encapsulate that in such a function. The source also has to be accessible from the first unit test, so the only resolution would be to perfectly forward the `HttpBlock` instantiation parameters to the `createGraph()` function, then return a tuple of `[source, httpBlock, sink]` from that function which can then be destructured from the calling site. But I encountered some move issues with this and discovered that some template parameters cannot longer be automatically derived, resulting in more boilerplate than we save from wrapping the three graph connections in a function, so in the end I opted to just leave it as is (unless you know a more elegant solution).

drslebedev:
@vimpostor 
Thank you for making those updates! The code looks good to me, and it can be merged once all the tests passed. 

Could you please squash the commits into a single commit or possibly two? Given the presence of some work-in-progress commits, consolidating them will contribute to a cleaner commit history.

vimpostor:
Squashed it into two commits.

drslebedev (approved):
@vimpostor
Thank you for making those updates! The code looks good to me, and it can be merged once all the tests passed.

Could you please squash the commits into a single commit or possibly two? Given the presence of some work-in-progress commits, consolidating them will contribute to a cleaner commit history.

PR #263, opened 2024-02-06
Fix runaway multiplication of messages

giucam:
This commit fixes runaway spamming of messages as soon as one is sent to/from the scheduler.


RalphSteinhagen (approved):

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -161,6 +161,7 @@ class SchedulerBase {
             if (const auto available = reader.available(); available > 0) {
                 const auto &input = reader.get(available);
                 outPort.streamWriter().publish([&](auto &output) { std::ranges::copy(input, output.begin()); }, available);
+                inPort.streamReader().consume(available);
```
frankosterfeld, line 164: I would reuse `reader`, for clarity.

wirew0rm: Since the function returns exactly the same reference, i'll change it and merge it as soon as the ci has passed again.

PR #264, opened 2024-02-06
messaging update

frankosterfeld:
Ivan's WIP, so I have a place to put comments.

### DIFF
--- a/core/include/gnuradio-4.0/Message.hpp
+++ b/core/include/gnuradio-4.0/Message.hpp
@@ -20,10 +20,40 @@ const std::string ErrorInfo = "ERROR_INFO_KEY"; // optional: if a message has an
 
 namespace message::kind {
 const std::string Error           = "ERROR_KIND";
-const std::string Graph_update    = "GRAPH_UPDATE_KIND";
+const std::string GraphUpdate     = "GRAPH_UPDATE_KIND";
+const std::string SchedulerUpdate = "SCHEDULER_UPDATE_KIND";
 const std::string UpdateSettings  = "UPDATE_SETTINGS_KIND";
 const std::string SettingsChanged = "SETTINGS_CHANGED_KIND";
 } // namespace message::kind
+
+namespace message::graph_update {
+namespace what {
+const std::string BlockAdded         = "ADDED_BLOCK_GRAPH_UPDATE";
+const std::string BlockRemoved       = "REMOVED_BLOCK_GRAPH_UPDATE";
+const std::string BlockExchanged     = "EXCHANGED_BLOCK_GRAPH_UPDATE";
+const std::string BlocksConnected    = "CONNECTED_BLOCKS_GRAPH_UPDATE";
+const std::string BlocksDisconnected = "DISCONNECTED_BLOCKS_GRAPH_UPDATE";
+} // namespace what
+
+namespace data {
+const std::string AddedBlockId   = "ADDED_BLOCK_ID";  // for BlockAdded, BlockExchanged
+const std::string RemovedBlockId = "REMOVE_BLOCK_ID"; // for BlockRemoved, BlockExchanged
+
+const std::string SourceBlock      = "SOURCE_BLOCK";
+const std::string SourcePort       = "SOURCE_PORT";
+const std::string DestinationBlock = "DESTINATION_BLOCK";
+const std::string DestinationPort  = "DESTINATION_PORT";
+} // namespace data
+} // namespace message::graph_update
+
+namespace message::scheduler_update {
+const std::string SchedulerStarted  = "STARTED_SCHEDULER_UPDATE";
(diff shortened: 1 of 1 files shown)


core/include/gnuradio-4.0/Message.hpp
```diff
@@ -20,11 +20,41 @@ const std::string ErrorInfo = "ERROR_INFO_KEY"; // optional: if a message has an
...
-const std::string Graph_update    = "GRAPH_UPDATE_KIND";
+const std::string GraphUpdate     = "GRAPH_UPDATE_KIND";
+const std::string SchedulerUpdate = "SCHEDULER_UPDATE_KIND";
 const std::string UpdateSettings  = "UPDATE_SETTINGS_KIND";
 const std::string SettingsChanged = "SETTINGS_CHANGED_KIND";
 } // namespace message::kind
 
+namespace message::graph_update {
+namespace what {
+const std::string BlockAdded         = "ADDED_BLOCK_GRAPH_UPDATE";
```
frankosterfeld, line 31: I could use some example messages how these would look like in practice. I assume they would also contain the initial block properties?

core/include/gnuradio-4.0/Message.hpp
```diff
@@ -20,11 +20,41 @@ const std::string ErrorInfo = "ERROR_INFO_KEY"; // optional: if a message has an
...
+const std::string BlockRemoved       = "REMOVED_BLOCK_GRAPH_UPDATE";
+const std::string BlockExchanged     = "EXCHANGED_BLOCK_GRAPH_UPDATE";
+const std::string BlocksConnected    = "CONNECTED_BLOCKS_GRAPH_UPDATE";
+const std::string BlocksDisconnected = "DISCONNECTED_BLOCKS_GRAPH_UPDATE";
```
frankosterfeld, line 35: There's the initial creation of the UI elements, which could be done by the scheduler sending an array of BlockAdded/BlocksConnected messages, or a single GraphDescription message (that might require a quite complex pmt structure, though). Alternatively, the UI could read the structure from the graph before passing it to the scheduler. I'd prefer the single channel (messages) though.

core/include/gnuradio-4.0/Message.hpp
```diff
@@ -20,11 +20,41 @@ const std::string ErrorInfo = "ERROR_INFO_KEY"; // optional: if a message has an
...
+
+namespace data {
+const std::string AddedBlockId   = "ADDED_BLOCK_ID";  // for BlockAdded, BlockExchanged
+const std::string RemovedBlockId = "REMOVE_BLOCK_ID"; // for BlockRemoved, BlockExchanged
+
+const std::string SourceBlock      = "SOURCE_BLOCK";
+const std::string SourcePort       = "SOURCE_PORT";
+const std::string DestinationBlock = "DESTINATION_BLOCK";
+const std::string DestinationPort  = "DESTINATION_PORT";
+} // namespace data
```
frankosterfeld, line 46: For some workflows it would be good to have some kind of loose request/response pattern: The user creates a new block at a certain position. The UI sends a CreateBlock message to the scheduler. The scheduler creates the block and sends back BlockCreated with the actual unique_name. To then link the new block to the previous request (and UI action), some kind of USER_CONTEXT could be passed with CreateBlock, that would then be passed back with BlockCreated.

core/include/gnuradio-4.0/Message.hpp
frankosterfeld: Would BlockAdded/BlockRemoved/ConnectionAdded/BlocksConnected/BlocksDisconnected be accompanied by AddBlock/RemoveBlock/ConnectBlocks/DisconnectBlocks commands, to trigger these from the UI? Or should it be the same command code for both directions?

core/include/gnuradio-4.0/Message.hpp
```diff
@@ -20,11 +20,41 @@ const std::string ErrorInfo = "ERROR_INFO_KEY"; // optional: if a message has an
...
+const std::string BlocksConnected    = "CONNECTED_BLOCKS_GRAPH_UPDATE";
+const std::string BlocksDisconnected = "DISCONNECTED_BLOCKS_GRAPH_UPDATE";
+} // namespace what
+
+namespace data {
+const std::string AddedBlockId   = "ADDED_BLOCK_ID";  // for BlockAdded, BlockExchanged
+const std::string RemovedBlockId = "REMOVE_BLOCK_ID"; // for BlockRemoved, BlockExchanged
+
+const std::string SourceBlock      = "SOURCE_BLOCK";
+const std::string SourcePort       = "SOURCE_PORT";
```
frankosterfeld, line 43: how do we describe the port (msg port vs. stream port) in the pmt?

frankosterfeld:
obsolete.

PR #268, opened 2024-02-08
Frank/message threading

frankosterfeld:


giucam:
With this Scheduler::processScheduledMessages() never gets called in the multi threaded scenario. I can call it from the outside every frame and that works, but i'm not sure if we want it to be manual like that and anyway it should be consistent between single and multi threaded.

frankosterfeld:
> With this Scheduler::processScheduledMessages() never gets called in the multi threaded scenario. I can call it from the outside every frame and that works, but i'm not sure if we want it to be manual like that and anyway it should be consistent between single and multi threaded.

We discussed that yesterday and said that the messages should be processed in the schedulers main thread. One would use runAndWait() or call processScheduledMessages manually (in the loop i used in opendigitizer, eg). Actually runAndWait is now more useful again, as one will be able to stop the scheduler from the UI thread via message.

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -161,7 +163,9 @@ class SchedulerBase {
             if (const auto available = reader.available(); available > 0) {
                 const auto &input = reader.get(available);
                 outPort.streamWriter().publish([&](auto &output) { std::ranges::copy(input, output.begin()); }, available);
-                reader.consume(available);
+                if (!reader.consume(available)) {
+                    throw std::runtime_error("Failed to consume messages from message port");
```
RalphSteinhagen, line 167: Since we now have the message ports in place, please avoid and replace throwing exceptions by using `emitMessage(...)` to emit error messages via the `msgOut` port.

Example code (in case of state errors) is:
```cpp
if (auto e = block.changeState(lifecycle::State::STOPPED); !e) {
    using namespace gr::message;
    this->emitMessage(this->msgOut, { { key::Kind, kind::Error }, { key::ErrorInfo, e.error().message }, { key::Location, e.error().srcLoc() } });
}
```

RalphSteinhagen (approved):
Looks OK and unit-tests pass. :+1: 

Please replace the new  (and future) exceptions with emitting error messages via the `msgOut'. Will will/shall/must gradually fade out non-user-thrown exceptions in the library. Added a small snippet that should help.

Otherwise, this can be merged as is.

core/include/gnuradio-4.0/Scheduler.hpp
frankosterfeld: Replaced the exceptions in Scheduler now with static_asserts/error messages

PR #269, opened 2024-02-08
Send out the staged settings when emitting a settings changed message

giucam:


core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1439,7 +1439,11 @@ class Block : public lifecycle::StateMachine<Derived>, protected std::tuple<Argu
 
                 Message settingsUpdated;
                 settingsUpdated[gr::message::key::Kind] = gr::message::kind::SettingsChanged;
-                settingsUpdated[gr::message::key::Data] = settings().get();
+                auto set = settings().get();
+                for (auto &&staged: settings().stagedParameters()) {
+                    set[staged.first] = staged.second;
+                }
+                settingsUpdated[gr::message::key::Data] = set;
```
wirew0rm, line 1446: The settings here are only staged and not yet applied, it would be more logical to only report the changed settings once they are applied here:

https://github.com/fair-acc/graph-prototype/blob/9c0d83b6f6bb0ac661a7b3ee49b8affdc9dcc0b0/core/include/gnuradio-4.0/Block.hpp#L847-L855

https://github.com/fair-acc/graph-prototype/blob/9c0d83b6f6bb0ac661a7b3ee49b8affdc9dcc0b0/core/include/gnuradio-4.0/Block.hpp#L860-L865

https://github.com/fair-acc/graph-prototype/blob/9c0d83b6f6bb0ac661a7b3ee49b8affdc9dcc0b0/core/include/gnuradio-4.0/Block.hpp#L591-L595

RalphSteinhagen:
closing this as this seems to be superseeded by PR https://github.com/fair-acc/graph-prototype/pull/281

PR #270, opened 2024-02-08
Create the maximum number of threads when the task type is CPU_BOUND

giucam:
This fixes the scheduler not working properly because it divides the blocks among the maximum number of threads but the thread pool was only creating the minimum amount of threads.
When minimum and maximum are not the same that resulted in only part of the blocks being executed, in case of a flowgraph that keeps running.

This makes the flowgraph in opendigitizer work again.

### DIFF
--- a/core/include/gnuradio-4.0/thread/thread_pool.hpp
+++ b/core/include/gnuradio-4.0/thread/thread_pool.hpp
@@ -405,7 +405,8 @@ public:
                     uint32_t max = std::thread::hardware_concurrency())
         : _poolName(name), _taskType(taskType), _minThreads(std::min(min, max)), _maxThreads(max) {
         assert(min > 0 && "minimum number of threads must be > 0");
-        for (uint32_t i = 0; i < _minThreads; ++i) {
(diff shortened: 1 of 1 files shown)


wirew0rm:
Some Questions:
- Where in opendigitizer does this occur? From my short investgation the opencmw worker uses `min=max=1` and the ui uses the default pool which sets both to `std::thread::hardware_concurrency()` -> can the immediate problem be worked around in opendigitizer and this be solved and documented properly with a little bit more time?
- I tried to compare the behavior of the implementation with and without the changes to the documentation and this is basically undocumented and the only example has min=max. I think the meaning of the min and max parameter for CPU_BOUND has to be clarified in the documentation
- `_minThreads` is used to set the cpuAffinity mask and otherwise unused. For me it looks like the CPU_BOUND policy only used min and this should also be changed? In general, should min = max be enforced for CPU_BOUND?

wirew0rm:
Discussed this online and determined that this has to be properly fixed and documented, the current iteration of the PR would only shift one problem to another.
For the immediate use in opendigitizer, there is the workaround of just not instantiating thread pools with min != max.

RalphSteinhagen:
stale PR.

PR #272, opened 2024-02-11
Fix race condition in multi-producer CircularBuffer

frankosterfeld:
Fix race condition in multi-producer CircularBuffer

In MultithreadedStrategy, the _cursor is pushed by the writers on reserve, i.e. points to the next's write's start position. Do not use by the readers to determine the number slots available for reading, but track that as a separate readCursor.
    
To update the readCursor safely when publishing a range, let a writer wait for any previous ranges to be published first. This trades correctness and simpler implementation for some performance.
    
These issues somehow surface with T=std::map (gr::Message), so add tests for that.

### DIFF
--- a/core/include/gnuradio-4.0/ClaimStrategy.hpp
+++ b/core/include/gnuradio-4.0/ClaimStrategy.hpp
@@ -24,12 +24,12 @@ struct NoCapacityException : public std::runtime_error {
 
 template<typename T>
 concept ClaimStrategy = requires(T /*const*/ t, const std::vector<std::shared_ptr<Sequence>> &dependents, const std::size_t requiredCapacity,
-        const std::make_signed_t<std::size_t> cursorValue, const std::make_signed_t<std::size_t> sequence, const std::make_signed_t<std::size_t> availableSequence, const std::size_t n_slots_to_claim) {
+        const std::make_signed_t<std::size_t> cursorValue, const std::make_signed_t<std::size_t> sequence, const std::make_signed_t<std::size_t> offset, const std::make_signed_t<std::size_t> availableSequence, const std::size_t n_slots_to_claim) {
     { t.hasAvailableCapacity(dependents, requiredCapacity, cursorValue) } -> std::same_as<bool>;
     { t.next(dependents, n_slots_to_claim) } -> std::same_as<std::make_signed_t<std::size_t>>;
     { t.tryNext(dependents, n_slots_to_claim) } -> std::same_as<std::make_signed_t<std::size_t>>;
     { t.getRemainingCapacity(dependents) } -> std::same_as<std::make_signed_t<std::size_t>>;
-    { t.publish(sequence) } -> std::same_as<void>;
+    { t.publish(offset, n_slots_to_claim) } -> std::same_as<void>;
     { t.isAvailable(sequence) } -> std::same_as<bool>;
     { t.getHighestPublishedSequence(sequence, availableSequence) } -> std::same_as<std::make_signed_t<std::size_t>>;
 };
@@ -108,7 +108,8 @@ public:
         return static_cast<signed_index_type>(_size) - (produced - consumed);
     }
 
-    void publish(signed_index_type sequence) {
+    void publish(signed_index_type offset, std::size_t n_slots_to_claim) {
+        const auto sequence = offset + static_cast<signed_index_type>(n_slots_to_claim);
         _cursor.setValue(sequence);
         _nextValue = sequence;
         if constexpr (hasSignalAllWhenBlocking<WAIT_STRATEGY>) {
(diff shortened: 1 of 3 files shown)


core/include/gnuradio-4.0/ClaimStrategy.hpp
```diff
@@ -283,7 +284,16 @@ class alignas(hardware_constructive_interference_size) MultiThreadedStrategy
     }
 
 private:
-    void                      setAvailable(signed_index_type sequence) noexcept { setAvailableBufferValue(calculateIndex(sequence), calculateAvailabilityFlag(sequence)); }
+    void                      setAvailable(signed_index_type sequence) noexcept {
+        setAvailableBufferValue(calculateIndex(sequence), calculateAvailabilityFlag(sequence));
+        while (true) {
+            const auto oldReadCursor = _readCursor->value();
+            const auto newReadCursor = getHighestPublishedSequence(oldReadCursor + 1, _writeCursor.value());
```
frankosterfeld, line 291: I think this is still not correct when actually using multiple writers, as getHighestPublishedSequence is operating on data shared between writers.

frankosterfeld:
Not the correct fix, should also test actual multi-writer test cases. Keeping this around to collect tests at least.

frankosterfeld:
> Not the correct fix, should also test actual multi-writer test cases. Keeping this around to collect tests at least.

Added that test case, and a hopefully correct solution that might have worse performance (practical implications to be tested), but are maybe good enough for the MVP, and the MsgPort use case.

core/include/gnuradio-4.0/ClaimStrategy.hpp
```diff
@@ -24,14 +24,13 @@ struct NoCapacityException : public std::runtime_error {
...
+        const std::make_signed_t<std::size_t> cursorValue, const std::make_signed_t<std::size_t> sequence, const std::make_signed_t<std::size_t> availableSequence, const std::size_t n_slots_to_claim, const std::size_t n_slots) {
     { t.hasAvailableCapacity(dependents, requiredCapacity, cursorValue) } -> std::same_as<bool>;
     { t.next(dependents, n_slots_to_claim) } -> std::same_as<std::make_signed_t<std::size_t>>;
     { t.tryNext(dependents, n_slots_to_claim) } -> std::same_as<std::make_signed_t<std::size_t>>;
     { t.getRemainingCapacity(dependents) } -> std::same_as<std::make_signed_t<std::size_t>>;
-    { t.publish(sequence) } -> std::same_as<void>;
-    { t.isAvailable(sequence) } -> std::same_as<bool>;
-    { t.getHighestPublishedSequence(sequence, availableSequence) } -> std::same_as<std::make_signed_t<std::size_t>>;
+    { t.publish(sequence, n_slots) } -> std::same_as<void>;
+    { t.readCursorPosition() } -> std::same_as<std::make_signed_t<std::size_t>>;
```
wirew0rm, line 33: Trying to find out whether it is possible to fix this without chaning the Circular Buffer API. I made an separate Branch with the unit tests and a yet incomplete fix on the Reader side here: https://github.com/fair-acc/graph-prototype/tree/fixCircularBufferNoApiChange

Unfortunately this does not seem to be enough to pass the new multiProducerMap unit tests.

core/include/gnuradio-4.0/ClaimStrategy.hpp
```diff
@@ -24,12 +24,12 @@ struct NoCapacityException : public std::runtime_error {
...
 template<typename T>
 concept ClaimStrategy = requires(T /*const*/ t, const std::vector<std::shared_ptr<Sequence>> &dependents, const std::size_t requiredCapacity,
-        const std::make_signed_t<std::size_t> cursorValue, const std::make_signed_t<std::size_t> sequence, const std::make_signed_t<std::size_t> availableSequence, const std::size_t n_slots_to_claim) {
+        const std::make_signed_t<std::size_t> cursorValue, const std::make_signed_t<std::size_t> sequence, const std::make_signed_t<std::size_t> offset, const std::make_signed_t<std::size_t> availableSequence, const std::size_t n_slots_to_claim) {
     { t.hasAvailableCapacity(dependents, requiredCapacity, cursorValue) } -> std::same_as<bool>;
     { t.next(dependents, n_slots_to_claim) } -> std::same_as<std::make_signed_t<std::size_t>>;
     { t.tryNext(dependents, n_slots_to_claim) } -> std::same_as<std::make_signed_t<std::size_t>>;
     { t.getRemainingCapacity(dependents) } -> std::same_as<std::make_signed_t<std::size_t>>;
-    { t.publish(sequence) } -> std::same_as<void>;
+    { t.publish(offset, n_slots_to_claim) } -> std::same_as<void>;
```
wirew0rm, line 32: note: while this is an API change, it corresponds to the [API from disruptor-cpp](https://github.com/Abc-Arbitrage/Disruptor-cpp/blob/master/Disruptor/MultiProducerSequencer.h#L186) and with the current API i don't see a way to reliably publish more than one slot (apart from introducing additional state or other different API).

RalphSteinhagen: OK.

RalphSteinhagen (approved):
Thanks @frankosterfeld and @wirew0rm for tackling this issue and following-up the deviation from the original Disruptor pattern. This is certainly the better/sound solution.

PR #273, opened 2024-02-13
Frank/messages scheduler control

frankosterfeld:
Scheduler: Allow lifecycle control via messages
    
Allow the remote control (start, stop, pause, resume) via messages.
    
For that, implement pause/resume, refactor the schedulers to use CRTP (so all lifecycle methods can be called from the base class).

### DIFF
--- a/core/include/gnuradio-4.0/Scheduler.hpp
+++ b/core/include/gnuradio-4.0/Scheduler.hpp
@@ -9,6 +9,7 @@
 #include <queue>
 
 #include <gnuradio-4.0/Graph.hpp>
+#include <gnuradio-4.0/LifeCycle.hpp>
 #include <gnuradio-4.0/Message.hpp>
 #include <gnuradio-4.0/Port.hpp>
 #include <gnuradio-4.0/Profiler.hpp>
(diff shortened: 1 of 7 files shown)


core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -17,24 +19,29 @@ using gr::thread_pool::BasicThreadPool;
 
 enum ExecutionPolicy { singleThreaded, multiThreaded };
 
-template<profiling::ProfilerLike TProfiler = profiling::null::Profiler>
+constexpr std::chrono::milliseconds kMessagePollInterval{ 10 };
+
+template<typename Derived, profiling::ProfilerLike TProfiler = profiling::null::Profiler>
 class SchedulerBase {
-public:
```
RalphSteinhagen, line 22: Since you made already larger modification to this class, please replace the custom state handling with `SchedulerBase` publically deriving from the  `lifecycle::StateMachine`. This should simplify also the callback handle w.r.t. `start()`, `stop()` and others and make this more consistent with `Block<T>`.

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -17,24 +19,29 @@ using gr::thread_pool::BasicThreadPool;
...
     std::atomic_bool                    _stop_requested;
 
     MsgPortOutNamed<"__ForChildren"> _toChildMessagePort;
     MsgPortInNamed<"__FromChildren"> _fromChildMessagePort;
 
-    std::string unique_name = gr::meta::type_name<SchedulerBase<TProfiler>>(); // TODO: to be replaced if scheduler derives from Block<T>
+    std::string unique_name = gr::meta::type_name<Derived>(); // TODO: to be replaced if scheduler derives from Block<T>
+
+    [[nodiscard]] constexpr lifecycle::State
+    state() const {
```
RalphSteinhagen, line 42: -> should become obsolete through using `lifecycle::StateMachine`.

core/include/gnuradio-4.0/Scheduler.hpp
RalphSteinhagen: This causes a heavy amount of untested code-duplication (see below) and must be addressed/simplified before being merged.

RalphSteinhagen (changes requested):
There is some untested code duplication compared to the existing `lifecycle::StateMachine` handling that must be addressed. This should make this PR also quite a bit smaller and easier to read/review.

Follow-up review by @drslebedev or @wirew0rm

core/include/gnuradio-4.0/Scheduler.hpp
frankosterfeld: Done, refactored the schedulers accordingly.

core/include/gnuradio-4.0/Scheduler.hpp
frankosterfeld: indeed

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -20,212 +24,283 @@ using namespace std::string_literals;
...
     }
 
     void
-    connectBlockMessagePorts() {
-        _graph.forEachBlock([this](auto &block) {
-            if (ConnectionResult::SUCCESS != _toChildMessagePort.connect(*block.msgIn)) {
-                this->emitMessage(msgOut, { { key::Kind, kind::Error },
-                                      { key::ErrorInfo, fmt::format("Failed to connect scheduler input message port to child '{}'", block.uniqueName()) },
-                                      { key::Location, fmt::format("{}", std::source_location::current()) } });
+    setBlocksState(lifecycle::State state) {
```
RalphSteinhagen, line 189: I spent quite some time on this and I think -- similiar to the emitError -- this function should not be explicitely defined but inlined for the time being. Similarly for the ` _graph.forEachBlock(...)`.

The rationale is simplicity, lean and clean design with least surprises or custom API that users and/or maintainers need to learn. 

Both functions combined are basic several onion layers of standard C++ functions that by renaming them obfuscates a bit what happens under the hood. In this case the `_graph.forEachBlock` is basically -- for example/illustration -- a
```cpp
std::ranges::for_each(_blocks, [f](const auto &block_ptr) { std::invoke(f, *block_ptr.get()); });
```
The emitError is basically a 
```cpp
 emitMessage(port, { { key::Kind, kind::Error }, { key::ErrorInfo, error.message }, { key::Location, error.srcLoc() } });
```
which hides the keys/fields that are being set into the method name (and creates an issue that the source location isn't the calling side but the location where emitError is being defined).

I think creating new public API here and at stage -- I know this is/was well meaning -- is PMO and we should introduce only new API when it's actually needed and all the other uses cases are clear. Especially w.r.t. the `emitError` case, it's mentioned already above that the `emitMessage` should be irradicated in favour of the scheduler being based on the `Block<T>` API  where these type of definitions and/or parallel CRTP (similar to the `lifecycle::StateMachine<T>`) are better placed.

Until then I think it's better to keep these definitions (especially there are only a few of them) explicited and unwrapped to the minimum API that is presently defined.

During the next iterations/PR, we should also remove `forEachBlock` from `Graph` since this is just a very thin wrapper that adds more costs in terms of cognitive complexity than benefit (at least at the moment).

RalphSteinhagen (dismissed):
@frankosterfeld, thanks for the changes and your efforts in addressing this.

I believe there's an opportunity to further simplify the externally facing (and internal) API. These basic schedulers serve as the reference implementation for more complex derivatives, such as work-stealing algorithms, topological considerations, and performance optimizations and as such should be as brief, concise, and with least-surprise/non-standard-C++ API as possible . Given their critical function and role, optimising for quality, performance, and readability is paramount.

In this context, I encourage us to streamline the codebase by leveraging existing STL features wherever possible. Additionally, let's aim to minimise the creation of new API/methods unless absolutely necessary. The focus should be on reducing duplication and avoiding unnecessary layers of indirection, which can introduce cognitive complexity.

While the changes made represent progress, I see room for refinement. I believe our next steps should prioritize further simplification.

For the follow-up PR review, I suggest involving @wirew0rm and/or @drslebedev for further insights.

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -20,212 +24,283 @@ using namespace std::string_literals;
...
+    executionPolicy() {
+        return execution;
+    }
+
     explicit SchedulerBase(gr::Graph &&graph, std::shared_ptr<BasicThreadPool> thread_pool = std::make_shared<BasicThreadPool>("simple-scheduler-pool", thread_pool::CPU_BOUND),
                            const profiling::Options &profiling_options = {})
         : _graph(std::move(graph)), _profiler{ profiling_options }, _profiler_handler{ _profiler.forThisThread() }, _pool(std::move(thread_pool)) {}
 
-    ~SchedulerBase() { stop(); }
+    ~SchedulerBase() { std::ignore = this->changeStateTo(lifecycle::State::REQUESTED_STOP); }
```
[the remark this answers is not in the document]
frankosterfeld: Prints error now and aborts, as that would be a logic error that should never happen.

core/include/gnuradio-4.0/Scheduler.hpp
frankosterfeld: Done, removed both emitError() and setBlocksState() abstractions.

RalphSteinhagen (approved):
@frankosterfeld thanks for your patience with my perhaps-perceived-nitpicking and making the changes. The code became much more succint and concise which should help further abstraction and maintenance.

Once the flaky QA passed, I'd merge this.

PR #277, opened 2024-02-19
Port HttpBlock subscription management to messaging ports

vimpostor:
This ports the subscription management of `HttpBlock` to a messaging port based implementation, which was the last missing piece to remove the input streaming port, which should fix the flaky CI test, when building with coverage enabled.

The triggering of `GET`-style requests was already ported to messaging ports in a later commit of the initial PR.

### DIFF
--- a/blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
+++ b/blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
@@ -213,7 +213,6 @@ private:
     }
 
 public:
-    PortIn<T>            in;
     PortOut<pmtv::map_t> out;
 
     std::string           url;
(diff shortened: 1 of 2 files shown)


blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
```diff
@@ -245,18 +244,7 @@ class HttpBlock : public gr::Block<HttpBlock<T>, BlockingIO<false>, HttpBlockDoc
     }
 
     [[nodiscard]] constexpr auto
-    processOne(T value) noexcept {
```
frankosterfeld, line 248: I wonder if `void processBulk(PublishableSpan auto &output)` would be a bit better here? 
Could the output port become a message port, too? Then the helper thread could probably write the messages directly into the port's writer?

RalphSteinhagen (approved):
OK

PR #279, opened 2024-02-20
added Drawable<UICategory, ToolKit> block annotation

RalphSteinhagen:
This adds default `Drawable` annotation in view of issue https://github.com/fair-acc/opendigitizer/issues/131 to declare whether a block supports rendering UI content that -- if declared -- must be drawn/processed through a mandatory `void draw() {...}` function, e.g.: 

```cpp
template<typename T>
struct ImChartMonitor : public Block<ImChartMonitor<T>, BlockingIO<false>, Drawable<UICategory::ChartPane, "console">> {
    PortIn<T>   in;
    float       sample_rate = 1000.0f;
    std::string signal_name = "unknown signal";

    constexpr void processOne(const T &input) noexcept { /*  ... as before ... */    }

    work::Status draw() noexcept {
        [[maybe_unused]] const work::Status status = this->invokeWork(); // calls work(...) -> processOne(...) (all in the same thread as this 'draw()'
        // handle drawing code
        return status;
    }
};
```

From the unit test:
```cpp
"drawable"_test = [] {
        struct TestBlock0 : gr::Block<TestBlock0> {
        } testBlock0;
        expect(!testBlock0.meta_information.value.contains("Drawable")) << "not drawable";

        struct TestBlock1 : gr::Block<TestBlock1, gr::Drawable<gr::UICategory::Toolbar, "console">> {
            void draw() {}
        } testBlock1;
        expect(testBlock1.meta_information.value.contains("Drawable")) << "drawable";
        const auto &drawableConfigMap = std::get<gr::property_map>(testBlock1.meta_information.value.at("Drawable"s));
        expect(drawableConfigMap.contains("Category"));
        expect(eq(std::get<std::string>(drawableConfigMap.at("Category")), "Toolbar"s));
        expect(drawableConfigMap.contains("Toolkit"));
        expect(eq(std::get<std::string>(drawableConfigMap.at("Toolkit")), "console"s));
    };
```

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/common_blocks.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/common_blocks.hpp
@@ -187,6 +187,9 @@ public:
         return { requested_work, available_samples, gr::work::Status::OK };
     }
 
+    gr::work::Status
+    draw() {}
+
     void
     processScheduledMessages() override {}
 
--- a/blocks/testing/include/gnuradio-4.0/testing/ImChartMonitor.hpp
+++ b/blocks/testing/include/gnuradio-4.0/testing/ImChartMonitor.hpp
@@ -13,7 +13,7 @@
 namespace gr::testing {
 
 template<typename T>
-struct ImChartMonitor : public Block<ImChartMonitor<T>, BlockingIO<false>> {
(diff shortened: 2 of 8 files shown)


drslebedev (approved):
Thanks for the PR!
The implementation of the new `Drawable` block annotation looks good.
The provided unit test examples demonstrates the common use cases. 

The PR can be merged.

PR #280, opened 2024-02-20
Fix httplib name

Waqar144:
It should be cpp-httplib otherwise it conflicts with httplib FetchContent in another project

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,12 +139,12 @@ FetchContent_Declare(
 )
 
 FetchContent_Declare(
-    httplib
+    cpp-httplib
     GIT_REPOSITORY https://github.com/yhirose/cpp-httplib.git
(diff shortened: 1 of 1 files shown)


wirew0rm (approved):
Additional explanation, hope this is what you mean: With fetchContent all requests for the same dependency have to use the same name, otherwise they will be included multiple time which will then lead to clashing targets. I first got confused by the description and assumed that this should prevent a collision of the first fetchContent argument.

This problem occurs when depending on opencmw-cpp and graph-prototype from the same project. The other declaration of the httplib dependency is [here](https://github.com/fair-acc/opencmw-cpp/blob/a7a7c5c319b93ddfaf160893665a011d2d88bff8/cmake/DependenciesNative.cmake#L18-L22). Using the github project name seems like the better and less ambiguous choice of the two.

PR #281, opened 2024-02-21
Send a message with updated settings whenever settings update

ivan-cukic:


core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1458,22 +1476,7 @@ class Block : public lifecycle::StateMachine<Derived>, protected std::tuple<Argu
...
-                settingsUpdated[gr::message::key::Data] = settings().get();
-
-                if (!notSet.empty()) {
-                    Message errorMessage;
-                    errorMessage[gr::message::key::Kind]         = gr::message::kind::Error;
-                    errorMessage[gr::message::key::Data]         = notSet;
-                    settingsUpdated[gr::message::key::ErrorInfo] = std::move(errorMessage);
-                }
-                emitMessage(msgOut, std::move(settingsUpdated));
+                applyChangedSettings(notSet);
```
RalphSteinhagen, line 1479: The pattern of staging parameters in non-real-time and applying them from the real-time thread is a common pattern in embedded real-time computing.

As discussed in the chat `applyChangedSettings(...)` may not be called from `processMessages(...)` for thread-safety and real-time latency constraints reasons. 

The other uses that are invoked from the `work(...)` or `process[Bulk, One](...)` contexts are OK though. I.e. the processMessage function may add/stage new settings but they can only be applied from a real-time processing thread point of view. 

Thus the UI will and can reliably only get a delayed reply of when the settings are actually being applied.

In most cases (continuous processing) the delay should be rather small though (few ms upt to at most <~100 ms).

RalphSteinhagen (changes requested):
Largely OK. 

The main critique is that `applyChangedSettings(...)` may not be called from `processMessages(...)` for thread-safety and notably real-time latency constraints reasons.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -882,14 +882,32 @@ class Block : public lifecycle::StateMachine<Derived>, protected std::tuple<Argu
...
+        Message settingsUpdated;
+        settingsUpdated[gr::message::key::Kind] = gr::message::kind::SettingsChanged;
+
         if (settings().changed()) {
-            if (const auto forward_parameters = settings().applyStagedParameters(); !forward_parameters.empty()) {
-                for_each_port([&forward_parameters](PortLike auto &outPort) { outPort.publishTag(forward_parameters, 0); }, outputPorts<PortType::STREAM>(&self()));
+            auto applyResult = settings().applyStagedParameters();
+
+            if (!applyResult.forwardParameters.empty()) {
+                for_each_port([&applyResult](PortLike auto &outPort) { outPort.publishTag(applyResult.forwardParameters, 0); }, outputPorts<PortType::STREAM>(&self()));
```
frankosterfeld, line 894: Couldn't this for_each_port loop be just a call to Block::publishTag()?

PR #282, opened 2024-02-22
Add size() member to dummy span types

Waqar144:
opendigitizer's fails to compile otherwise


RalphSteinhagen (approved):
OK ... @Waqar144  thanks for the changes. This is benine but it's a bit unclear why we need this because the `DummyConsumableSpan` only is used to check the concept whether the user-provided `processBulk(...)` function passes the concept... :thinking:

PR #283, opened 2024-02-22
Frank/datasink metadata

frankosterfeld:
Fix issues in DataSink:

Make sure lookup by signal_name is thread-safe and is not accessing the
signal_name member directly, which might be updated in another thread.

Simplify and fix the signal metadata propagation to work with the automatic forwarding of source properties matching default tags.

Fix the metadata propagation for GRC-loaded graphs.

Needed for https://github.com/fair-acc/opendigitizer/issues/159 and https://github.com/fair-acc/opendigitizer/issues/160

### DIFF
--- a/core/include/gnuradio-4.0/Block.hpp
+++ b/core/include/gnuradio-4.0/Block.hpp
@@ -318,10 +318,11 @@ struct isBlockDependent {
  *     * case sinks: HW triggered vs. fixed-size consumer (may block/never finish for insufficient input data and fixed Port::MIN>0)
  * <ul>
  *
- * In addition derived classes can optionally implement any subset of the lifecycle methods ( `start()`, `stop()`, `reset()`, `pause()`, `resume()`).
+ * In addition derived classes can optionally implement any subset of the lifecycle methods ( `init()`, `start()`, `stop()`, `reset()`, `pause()`, `resume()`).
  * The Scheduler invokes these methods on each Block instance, if they are implemented, just before invoking its corresponding method of the same name.
  * @code
  * struct userBlock : public Block<userBlock> {
+ * void init() {...} // Implement any initialisation logic required for the block within this method.
  * void start() {...} // Implement any startup logic required for the block within this method.
  * void stop() {...} // Use this method for handling any clean-up procedures.
  * void pause() {...} // Implement logic to temporarily halt the block's operation, maintaining its current state.
@@ -593,7 +594,7 @@ public:
     }
 
     void
-    init(std::shared_ptr<gr::Sequence> progress_, std::shared_ptr<gr::thread_pool::BasicThreadPool> ioThreadPool_) {
+    performInit(std::shared_ptr<gr::Sequence> progress_, std::shared_ptr<gr::thread_pool::BasicThreadPool> ioThreadPool_) {
         progress     = std::move(progress_);
         ioThreadPool = std::move(ioThreadPool_);
 
--- a/core/include/gnuradio-4.0/Block.hpp
+++ b/core/include/gnuradio-4.0/Block.hpp
@@ -318,11 +318,10 @@ struct isBlockDependent {
  *     * case sinks: HW triggered vs. fixed-size consumer (may block/never finish for insufficient input data and fixed Port::MIN>0)
  * <ul>
  *
- * In addition derived classes can optionally implement any subset of the lifecycle methods ( `init()`, `start()`, `stop()`, `reset()`, `pause()`, `resume()`).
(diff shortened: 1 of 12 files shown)


core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -85,7 +85,7 @@ class SchedulerBase : public lifecycle::StateMachine<Derived> {
 
     void
     stateChanged(lifecycle::State newState) {
-        emitMessage(msgOut, { { kind::SchedulerStateUpdate, std::string(magic_enum::enum_name(newState)) } });
+        emitMessage(msgOut, { { key::Kind, kind::SchedulerStateUpdate }, { key::What, std::string(magic_enum::enum_name(newState)) } });
```
frankosterfeld, line 88: Thinking about it, maybe `key::Data` instead of `key::What` should be used? @ivan-cukic do you have an opinion?

frankosterfeld:
For the question of how to handle the old init(...) (now performInit() in this PR) method vs. the state machine callback also called `init()` atm, also see https://github.com/fair-acc/graph-prototype/issues/289

core/include/gnuradio-4.0/Scheduler.hpp
frankosterfeld: done

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -318,10 +318,11 @@ struct isBlockDependent {
  *     * case sinks: HW triggered vs. fixed-size consumer (may block/never finish for insufficient input data and fixed Port::MIN>0)
  * <ul>
  *
- * In addition derived classes can optionally implement any subset of the lifecycle methods ( `start()`, `stop()`, `reset()`, `pause()`, `resume()`).
+ * In addition derived classes can optionally implement any subset of the lifecycle methods ( `init()`, `start()`, `stop()`, `reset()`, `pause()`, `resume()`).
```
RalphSteinhagen, line 321: please revert

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -318,10 +318,11 @@ struct isBlockDependent {
...
  * The Scheduler invokes these methods on each Block instance, if they are implemented, just before invoking its corresponding method of the same name.
  * @code
  * struct userBlock : public Block<userBlock> {
+ * void init() {...} // Implement any initialisation logic required for the block within this method.
```
RalphSteinhagen, line 325: and this.

RalphSteinhagen (changes requested):
@frankosterfeld , thanks for the update. :+1: 

I like the `std::optional<U> getProperty(const property_map &m, std::string_view key);`. We should (separate PR) elevate this to using `std::expected<..>` and make this a core function or perhaps for the pmt-library itself as this is a common use case that I also saw @ivan-cukic introduced for the message handling. 

I also like the GRC-based loader-based unit-test in qa_Settings. Why is this disabled for Emscripten?
As discussed in the GR Architecture channel, this grc-file-based definition is good both for unit-tests as well as basic-usage documentation. We had a quick iteration over here, and maybe we will add another optional `using ... = fixed_string("grc definition");` to the blocks to allow for documenting these (at least the basic) use cases. Unit-tests and documentation aside, this could also be used on the GR website to launch a WASM GR UI for the block documentation where users could directly play/experiment with a give block.

Regarding the modifications to core library functions in Scheduler.hpp and Block.hpp (comments only), unless these changes address specific bugs or are critically needed, I suggest reverting them due to concurrent modifications being made in other PRs.

Your contributions are valuable, and with these minor adjustments, this PR could be merged. Thanks in advance.

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -85,7 +85,7 @@ class SchedulerBase : public lifecycle::StateMachine<Derived> {
 
     void
     stateChanged(lifecycle::State newState) {
-        emitMessage(msgOut, { { kind::SchedulerStateUpdate, std::string(magic_enum::enum_name(newState)) } });
+        emitMessage(msgOut, { { key::Kind, kind::SchedulerStateUpdate }, { key::Data, std::string(magic_enum::enum_name(newState)) } });
```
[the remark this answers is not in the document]
frankosterfeld: Well, it's not needed, I just think these messages are malformed (my fault, from the previous PR), and do not follow the intended pattern of key/value pairs, where the keys from the key:: namespace.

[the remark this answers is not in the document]
frankosterfeld: reverted here, moved to https://github.com/fair-acc/graph-prototype/pull/294

RalphSteinhagen (approved):
@frankosterfeld thanks for the changes. Will squash-merge it once the CI passes.

PR #286, opened 2024-02-23
Fix compiler warnings.

drslebedev:
I have fixed nearly all of the warnings present. However, a few remain unfixed for specific reasons:

1) For warnings originating from `qa_port_array.cpp`, I have implemented fixes in a separate PR, as these are closely related to changes being made there.
2) Regarding the warnings in `ImChart`, I believe @RalphSteinhagen  would be the best person to tackle these.


RalphSteinhagen (approved):
Thanks for the changes and clean-up. Can be merged once the CI passes.

PR #288, opened 2024-02-25
Add blocks defined in the blocklib to the blocks registry

ivan-cukic:
- Blocks that can be added to the blocks registry have been added
- Blocks that can not be added because of NTTPs or template template parameters have been marked as such
- Added notes to the blocks defined in the tests that they should be added to the block registry if they are moved to blocklib
- Small block registry refactoring
- Replaced `ENABLE_REFLECTION_FOR_TEMPLATE_FULL` with `ENABLE_REFLECTION_FOR_TEMPLATE` where possible


RalphSteinhagen:
> Blocks that can not be added because of NTTPs or template template parameters have been marked as such

Is this a fundamental limit? Would be a pity if it is.

We need/use NTTPs to minimise the amount of external code generation and code duplication. Notably, for example, basic math operations where the block structure is identically apart from the underlying `+`, `-`, `*`, ... operators and type multiplicity (i.e. `int8_t`, ..., `float32_t`, ..., `std::complex<T>`, `UncertainValue<T>`, `DataSet<T>`, etc.).

Also, NTTP aside, we used the `ENABLE_REFLECTION_FOR_TEMPLATE_FULL` reflection macro to express blocks with more than one type parameter. Notably, for example, those conversion blocks that translate between `T->U` (e.g. `uint16_t -> float32_t`) and/or have multiple output options such as the FFT (`T`=`uint8_t`, ..., `std::complex<float>`  to `U`=`float`, `std::complex<float>`, `std::vector<std::complex<float>>`  or `DataSet<float>` )

We must find a solution w/o spelling out all the different combinations.

RalphSteinhagen:
> Added notes to the blocks defined in the tests that they should be added to the block registry if they are moved to blocklib

Yoda's principles: either do or don't, don't try. Either these are being moved or drop the superfluous in-code documentation (creates noise and the meaning will soon be lost to history).

blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
```diff
@@ -262,6 +262,8 @@ struct ClockSource : public gr::Block<ClockSource<T, useIoThread, ClockSourceTyp
 ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T, bool useIoThread, typename ClockSourceType), (gr::basic::ClockSource<T, useIoThread, ClockSourceType>), out, n_samples_max, chunk_size, sample_rate,
                                     verbose_console);
 
+// NOTE: ClockSource can not be registered to the block registry (GR_REGISTER_BLOCK) as it has NTTPs
```
RalphSteinhagen, line 265: See my comment above.

blocks/basic/test/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
 add_ut_test(qa_Selector)
 add_ut_test(qa_sources)
 add_ut_test(qa_DataSink)
+
```
RalphSteinhagen, line 4: Maybe we should use the described workaround in the issue:

> Execute calling boost::ut::cfg<boost::ut::override>.run(); from main and the test passes, but arguments such --help do nothing.

This would also help with the UT command line options.

blocks/basic/test/qa_BasicKnownBlocks.cpp
```diff
@@ -0,0 +1,27 @@
...
+    std::vector<std::string> desired{
+        //
+        "builtin_counter"s,              //
+        "builtin_multiply"s,             //
+        "gr::basic::DataSink"s,          //
+        "gr::basic::FunctionGenerator"s, //
+        "gr::basic::Selector"s,          //
+        "gr::basic::SignalGenerator"s    //
+    };
+    std::ranges::sort(desired);
```
RalphSteinhagen, line 24: This is a brittle test and will fail whenever someone adds a new Block to `gr::globalBlockRegistry()`. Must be changed to something more robust and/or test a subset/intersection.

blocks/basic/test/qa_Selector.cpp
```diff
@@ -55,7 +55,9 @@ struct repeated_source : public gr::Block<repeated_source<T>> {
     }
 };
 
-ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T), (repeated_source<T>), identifier, remaining_events_count, values, out);
+ENABLE_REFLECTION_FOR_TEMPLATE(repeated_source, identifier, remaining_events_count, values, out);
+
+// NOTE: If you move repeated_source to blocklib, call GR_REGISTER_BLOCK
```
RalphSteinhagen, line 60: please remove these and other related comments. Either refactor this specific blocks so that they can be part of the official blocklib or drop the comment because it's a superfluous documentation with its meaning soon being forgotten.

N.B. we agreed that unit-tests should -- in large -- not rely on custom block definitions that are only available in the unit-tests but rely on the official curated ones from the blocklib. This makes maintenance, refactoring, and co-use of common functionality so much easier.

Also, many of this unit-test-only blocks have redundant functions to the existing blocklibs ones.

blocks/filter/include/gnuradio-4.0/filter/all.hpp
```diff
@@ -0,0 +1,6 @@
+#ifndef GNURADIO_BLOCKS_FILTER_ALL_HPP
```
RalphSteinhagen, line 1: Why this header definition?

blocks/filter/include/gnuradio-4.0/filter/time_domain_filter.hpp
```diff
@@ -36,14 +37,17 @@ H(z) = b[0] + b[1]*z^-1 + b[2]*z^-2 + ... + b[N]*z^-N
     }
 };
 
-enum class IIRForm {
```
RalphSteinhagen, line 39: What's wrong with this enum class definition? I thought this would be a perfectly valid use-case to multiplex on the specific IIR implementation?

blocks/filter/test/qa_FilterKnownBlocks.cpp
```diff
@@ -0,0 +1,23 @@
...
+    std::ranges::sort(known);
+
+    std::vector<std::string> desired{
+        //
+        "gr::filter::fir_filter"s, //
+        "gr::filter::iir_filter"s, //
+    };
+    std::ranges::sort(desired);
+
+    assert((std::ranges::includes(known, desired)));
```
RalphSteinhagen, line 22: brittle unit-test. Please refactor for testing on intersection.

blocks/fourier/include/gnuradio-4.0/fourier/fft.hpp
```diff
@@ -228,5 +228,6 @@ On the choice of window (mathematically aka. apodisation) functions
 
 ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T, typename U, template<typename, typename> typename FourierAlgoImpl), (gr::blocks::fft::FFT<T, U, FourierAlgoImpl>), //
                                     in, out, algorithm, fftSize, window, outputInDb, outputInDeg, unwrapPhase, sample_rate, signal_name, signal_unit, signal_min, signal_max);
+// NOTE: FFT can not be registered to the block registry (GR_REGISTER_BLOCK) as it has a template template parameter
```
RalphSteinhagen, line 231: As described above, we need to make this use-case work ... otherwise we fall back to either external code-generation or spelling-out all type/NTTP combinations.

blocks/fourier/test/qa_fourier.cpp
```diff
@@ -35,7 +35,9 @@ struct CountSource : public gr::Block<CountSource<T>> {
     }
 };
 
-ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T), (CountSource<T>), out, count, nSamples);
+ENABLE_REFLECTION_FOR_TEMPLATE(CountSource, out, count, nSamples);
+
+// NOTE: If you move CountSource to blocklib, call GR_REGISTER_BLOCK
```
RalphSteinhagen, line 40: remove. `CountSource` is anyway a redundant block.

blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
```diff
@@ -220,6 +223,8 @@ class HttpBlock : public gr::Block<HttpBlock<T>, BlockingIO<false>, HttpBlockDoc
     gr::http::RequestType type = gr::http::RequestType::GET;
     std::string           parameters; // x-www-form-urlencoded encoded POST parameters
 
+    explicit HttpBlock() {}
```
RalphSteinhagen, line 226: This slipped through in an earlier review. Blocks should not define their own constructor but rely on either the life-cycle callback methods and/or `settingsChanged(...)` for initialising. 

These constructors are not invoked by the default `emplaceBlock(...)` construction method and notably also do not invoke the constructors defined in `Block<T>` which causes issues w.r.t. settings handling, resetting blocks, and default initialisation.

If you could fix this and the corresponding unit-test please.

blocks/http/include/gnuradio-4.0/http/all.hpp
```diff
@@ -0,0 +1,6 @@
+#ifndef GNURADIO_BLOCKS_HTTP_ALL_HPP
```
RalphSteinhagen, line 1: Why this header file? Doesn't one usually 'only-include-what-you-need'?!?

blocks/http/test/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 add_ut_test(qa_HttpBlock)
 target_link_libraries(qa_HttpBlock PRIVATE gr-http)
 
+# Not using ut as it has issues with static objects
+# https://github.com/boost-ext/ut/issues/611
```
RalphSteinhagen, line 5: As above, this should be an easy fix with adding the `boost::ut::cfg<boost::ut::override>.run();` in the unit-tests `main()` method.

blocks/http/test/qa_HttpBlock.cpp
```diff
@@ -58,7 +60,8 @@ class HttpTestSink : public gr::Block<HttpTestSink<T>> {
     }
 };
 
-ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T), (HttpTestSink<T>), in);
+ENABLE_REFLECTION_FOR_TEMPLATE(HttpTestSink, in);
+// NOTE: If you move HttpTestSink to blocklib, call GR_REGISTER_BLOCK
```
RalphSteinhagen, line 64: as above.

blocks/http/test/qa_HttpKnownBlocks.cpp
```diff
@@ -0,0 +1,22 @@
...
+    auto known = gr::globalBlockRegistry().knownBlocks();
+    std::ranges::sort(known);
+
+    std::vector<std::string> desired{
+        //
+        "gr::http::HttpBlock"s, //
+    };
+    std::ranges::sort(desired);
+
+    assert((std::ranges::includes(known, desired)));
```
RalphSteinhagen, line 21: brittle test depends on sorting. Intersection might be better. Also, this test is nearly identical to others and should probably be grouped to one for simplicity.

blocks/testing/include/gnuradio-4.0/testing/TagMonitors.hpp
```diff
@@ -346,11 +346,15 @@ struct TagSink : public Block<TagSink<T, UseProcessVariant>> {
 } // namespace gr::testing
 
 ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T, gr::testing::ProcessFunction b), (gr::testing::TagSource<T, b>), out, n_samples_max, sample_rate, signal_name, verbose_console, mark_tag);
+// NOTE: TagSource can not be registered to the block registry (GR_REGISTER_BLOCK) as it has NTTPs
```
RalphSteinhagen, line 349: Critical: we need to find a solution to handle NTTPs via the plugin system.

blocks/testing/include/gnuradio-4.0/testing/all.hpp
```diff
@@ -0,0 +1,11 @@
+#ifndef GNURADIO_BLOCKS_TESTING_ALL_HPP
```
RalphSteinhagen, line 1: do we need this? Only-include-what-you-need paradigm.

blocks/testing/test/qa_TestingKnownBlocks.cpp
```diff
@@ -0,0 +1,26 @@
...
+        //
+        "gr::testing::FunctionSource"s,  //
+        "gr::testing::FunctionProcess"s, //
+        "gr::testing::FunctionSink"s,    //
+        "gr::testing::MessageSender"s,   //
+        "gr::testing::InspectSink"s      //
+    };
+    std::ranges::sort(desired);
+
+    assert((std::ranges::includes(known, desired)));
```
RalphSteinhagen, line 25: I think this could/should be made a UT-based unit-test.

core/benchmarks/bm-nosonar_node_api.cpp
```diff
@@ -96,6 +97,8 @@ class math_bulk_op : public gr::Block<math_bulk_op<T, op>, gr::PortInNamed<T, "i
     }
 };
 
+// NOTE: math_bulk_op can not be registered to the block registry (GR_REGISTER_BLOCK) as it has NTTPs
```
RalphSteinhagen, line 100: This is one of the prominent use-cases why we use C++: minimising code-duplication through templating.

We must find an acceptable solution for this.

core/test/qa_Block.cpp
```diff
@@ -22,6 +22,7 @@ struct BlockSignaturesNone : public gr::Block<BlockSignaturesNone<T>> {
 };
 
 ENABLE_REFLECTION_FOR_TEMPLATE(BlockSignaturesNone, in, out);
+// NOTE: If you move BlockSignaturesNone to blocklib, call GR_REGISTER_BLOCK
```
RalphSteinhagen, line 25: Superfluous comment in this case, as these prototype blocks are only empty implementations needed for their `process[Bulk, One](...)` function signatures. Comments can be removed as these blocks will never make it into the blocklib.

core/test/qa_grc.cpp
```diff
@@ -9,20 +9,28 @@
 
 #include <boost/ut.hpp>
 
+#if defined(__clang__) && __clang_major__ >= 16
+// clang 16 does not like ut's default reporter_junit due to some issues with stream buffers and output redirection
+template<>
+auto boost::ut::cfg<boost::ut::override> = boost::ut::runner<boost::ut::reporter<>>{};
+#endif
+
```
RalphSteinhagen, line 17: Is this still relevant since we use clang17?

RalphSteinhagen (changes requested):
Largely OK. 

The main comments/concerns that need to be addressed are:

  * support blocks with two typed parameters and NTTPs -- we need to find a viable solution without having to rely on external code-generation and spelling-out all different combinations.
  * > Added notes to the blocks defined
  
    Yoda's principles: either do or don't, don't try. Either these are being moved or drop the superfluous in-code documentation (creates noise and the meaning will soon be lost to history). In many cases these are redundant unit-test-only block definitions that should be removed/superseeded by the canonical ones in the blocklib.
  * > Execute calling boost::ut::cfgboost::ut::override.run(); from main and the test passes, but arguments such --help do nothing.
    
    Rather than frequently commenting/repeating on the issue (creates noise that will likely soon be ignored), we should probably use the proposed workaround of calling `boost::ut::cfg<boost::ut::override>.run();` in the unit-test's `main(...)` function.

blocks/http/include/gnuradio-4.0/http/all.hpp
ivan-cukic: If we want the blocks to be visible to the registry, they are all needed.

Applications that don't need GRC (or any runtime block instantiation), but only use fixed C++-defined flows are fine to include only the blocks they need. Applications that need to be able to instantiate any known block have to include them all.

core/benchmarks/bm-nosonar_node_api.cpp
ivan-cukic: As written in the general comment, we could convert this to a type param.

blocks/http/test/qa_HttpKnownBlocks.cpp
ivan-cukic: See the earlier comment.

I'd like to group them, but they are in different components.

blocks/filter/include/gnuradio-4.0/filter/time_domain_filter.hpp
ivan-cukic: See the general comment

core/test/qa_grc.cpp
ivan-cukic: Yes, crashed on clang 17 for me

blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
ivan-cukic: There are quite a few blocks that do the same. I'd propose creating a task to remove the `emplaceBlock` version that is not settings-based (the one that is marked for removal) and to port all blocks and tests to it.

blocks/basic/test/qa_BasicKnownBlocks.cpp
RalphSteinhagen: OK. mea culpa. Assumed that `std::ranges::includes` would required both sorted and contiguous, which it doesn't

https://compiler-explorer.com/z/Gz89frx8n

RalphSteinhagen:
Think I found a way to support registering blocks (w/ automatic supported type and NTTP deduction) for the following limited `Block<T>` cases:

* non-templated -- rare but possible
* `template <auto>` - one NTTP-only (skipped for brevity)
* `template <T>` -- assumed to be the most common case
* `template <T, auto>` -- assumed to be the second most common case
* `template <T, U>` -- two types  (skipped for brevity, example: conversion blocks)
* `template <T, U, auto>` -- two types 1 2 NTTPs (skipped for brevity)
* `template <T, U, auto, auto>` -- two types + 2 NTTPs (skipped for brevity)

The limits are 'by-convention' and to minimise the library boiler-plate code. If we want to cover more complex permutations, we probably should start looking more into JIT compilation for these special cases.

Registration could be done during run-time or statically via something like:
```cpp
// user-space definition of blocks, for example:

enum class Enum0 : int { V0, V1, V2 };

template<typename T, Enum0 e = Enum0::V0>
struct MyBlock0 : Block<MyBlock0<T, e>> {
    static constexpr const char* name = "MyBlock0"; // derived via reflection
    using supported_type = std::tuple<int, std::string>; // defined by user or using default
    using value_type = T;

    void printImpl() {
        fmt::println("  type parameter: {} enum value: {}", type_name<value_type>(), static_cast<int>(e));
    }
};

enum class Enum1 : int { V0, V1, V2 };

template<typename T, Enum1 e = Enum1::V1>
struct MyBlock1 : Block<MyBlock1<T, e>> {
    static constexpr const char* name = "MyBlock1"; // derived via reflection
    using supported_type = std::tuple<float, double>; // defined by user or using default
    using value_type = T;

    void printImpl() {
        fmt::println("  type parameter: {} enum value: {}", type_name<value_type>(), static_cast<int>(e));
    }
};

enum class Enum2 : int { V0, V1, V2 };

template<typename T, Enum2 e = Enum2::V2>
struct MyBlock2 : Block<MyBlock2<T, e>> {
    static constexpr const char* name = "MyBlock2"; // derived via reflection
    using supported_type = std::tuple<std::string>; // defined by user or using default
    using value_type = T;

    void printImpl() {
        fmt::println("  type parameter: {} enum value: {}", type_name<value_type>(), static_cast<int>(e));
    }
};

template<typename T>
struct MyBlock3 : Block<MyBlock3<T>> {
    static constexpr const char* name = "MyBlock3"; // derived via reflection
    using supported_type = std::tuple<float>; // defined by user or using default
    using value_type = T;

    void printImpl() {
        fmt::println("  type parameter: {}", type_name<value_type>());
    }
};

struct MyBlock4 : Block<MyBlock4> {
    static constexpr const char* name = "MyBlock4"; // derived via reflection
    using value_type = void;

    void printImpl() {
        fmt::println("  non-templated block");
    }
};

// registering blocks, here statically:

[[maybe_unused]]static const auto _register = [](){ // could find a prettier wrapper/definition
    // static registration
    registerBlockImpl<MyBlock0, Enum0::V0>();
    registerBlockImpl<MyBlock1, Enum1::V1>();
    registerBlockImpl<MyBlock2, Enum2::V2>();
    registerBlockImpl<MyBlock3>();
    registerBlockImpl<MyBlock4>();
    return 0;
}();
```
The `supported_types` could be defined within the user template/class or (perhaps, tbd) provided externally. Internally would have the advantage of being able to re-use this for compile-time derived block documentation. All code subject to review/simplification etc. (notably the use of tuples could probably benefit from using type_lists instead).

The concept is here: https://compiler-explorer.com/z/zbdxK8Gxn

RalphSteinhagen (approved):
Thanks for the changes! LGTM now!

There is at least one unrelated/known flakey test w.r.t. benchmark and getting perf data.

Need to see that the CI passes also for the debug target (likely another issue) before this PR can be merged.

blocks/http/include/gnuradio-4.0/http/HttpBlock.hpp
```diff
@@ -184,56 +197,63 @@ class HttpBlock : public gr::Block<HttpBlock<T>, BlockingIO<false>, HttpBlockDoc
     }
 
     void
-    runThread() {
+    startThread() {
+        if (_thread) {
+            _thread.reset();
```
vimpostor, line 202: @RalphSteinhagen What is this new check and reset for? This is effectively a noop, as the line immediately below it assigns a new `shared_ptr`, which includes implicitly calling `reset()` on the old one, or am I missing an edge case here?

RalphSteinhagen: Yes, there is a data-race in the reset/thread destruction. The new thread creation and deletion of the old access shared resources which isn't safe. Forcing the reset mitigated this.

PR #290, opened 2024-02-27
Support processBulk() function with vector<ConsumableSpan/PublishableSpan/std::span>

drslebedev:
Key Features of this PR:

1) Enhance the `processBulk()` function signature to support vectors of `ConsumableSpan` and `PublishableSpan` as well as  vector of `std::span<T>`. Here are some examples:
```cpp
gr::work::Status
    processBulk(const std::vector<std::span<const T>> &, std::vector<std::span<T>> &)
```
```cpp
template<gr::ConsumableSpan TInput, gr::PublishableSpan TOutput>
    gr::work::Status
    processBulk(const std::vector<TInput> &, std::vector<TOutput> &)
```

2) Removed strict checks and the default value for `BufferReader::get()`, where calling `get(0)` now returns 0 samples.

3) Introduce `SpanReleasePolicy`, which determines how samples are handled if `consume()` or `publish()` were not explicitly called by the user in `processBulk()`. 
The default policy for `Sync` ports are consume/publish `All` samples.
The default policy for `Async` ports are consume/publish `Zero` samples.

4) Added support for repeated values in `TagSource`, remove obsolete `RepeatedSource` and `SinkValidator` blocks.

The PR consists of 5 commits. They are rather independent and can be checked one-by-one to simplify review process.


blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
```diff
@@ -325,7 +325,8 @@ class DataSink : public Block<DataSink<T>> {
                 fnc(readData, std::span<const Tag>(relevantTags));
                 std::ignore = tag_reader.consume(relevantTags.size());
             } else {
-                std::ignore = tag_reader.consume(tag_reader.available());
+                const auto tags = tag_reader.get();
+                std::ignore     = tag_reader.consume(tag_reader.available());
```
RalphSteinhagen, line 329: should probably be 

```cpp
if (tag_reader.consume(tags.size())) { /* handle error*/};
```

since `available()` may change between successive calls.

RalphSteinhagen (approved):
We discussed this already quite a bit off-line and the final result looks good to me... I love `std::span`! :+1: 

Thanks also for clean-up/refactoring of the unit-tests and merging of helper blocks with Tag[Source,Sink]!

PR #297, opened 2024-03-14
Refactor workInternal

wirew0rm:
This PR aims to restructure the core work function dispatch mechanism implemented in `Block::workInternal`.

Noteworthy changes:
- breaks up the code into determining available samples and afterwards dispatching to the correct work function
- restructure and partially reimplement decimate and stride functionalities:
  - the decimation settings (numerator/denominator) now determine the minimum atomic amount of work that can be processed (previously this was determined by the ports' minimum samples)
  - stride is applied after denominator samples. (Previously controlled by the ports' maxium samples)
- cleanup
  - reduce (gcc) warnings to 0

Open issues:
- [x] clang/emscirpten test failures
- [ ] treatment of incomplete last chunk
- [ ] support for moving tags to the first tag of this or the next tag if port constraints don't allow chunking.
- [x] remove debug printouts
- [x] squash git history
- [x] investigate and reenable asan

### DIFF
--- a/core/include/gnuradio-4.0/Block.hpp
+++ b/core/include/gnuradio-4.0/Block.hpp
@@ -385,7 +385,7 @@ class Block : public lifecycle::StateMachine<Derived>, //
 //              protected std::tuple<typename gr::meta::typelist<Arguments...>::template filter<gr::isBlockDependent>> // only add port types to the tuple, the other info are kept in the using
 //              statements below
 {
-    static std::atomic_size_t _unique_id_counter;
+    static std::atomic_size_t _uniqueIdCounter;
     template<typename T, gr::meta::fixed_string description = "", typename... Args>
     using A = Annotated<T, description, Args...>;
 
@@ -396,6 +396,7 @@ public:
     using block_template_parameters  = meta::typelist<Arguments...>;
     using Resampling                 = ArgumentsTypeList::template find_or_default<is_resampling_ratio, ResamplingRatio<1UL, 1UL, true>>;
     using StrideControl              = ArgumentsTypeList::template find_or_default<is_stride, Stride<0UL, true>>;
+    using AllowIncompleteFinalUpdate = ArgumentsTypeList::template find_or_default<is_incompleteFinalUpdatePolicy, IncompleteFinalUpdatePolicy<IncompleteFinalUpdateEnum::DROP>>;
     using DrawableControl            = ArgumentsTypeList::template find_or_default<is_drawable, Drawable<UICategory::None, "">>;
     constexpr static bool blockingIO = std::disjunction_v<std::is_same<BlockingIO<true>, Arguments>...> || std::disjunction_v<std::is_same<BlockingIO<false>, Arguments>...>;
 
@@ -422,7 +423,6 @@ public:
 
     constexpr static TagPropagationPolicy tag_policy = TagPropagationPolicy::TPP_ALL_TO_ALL;
 
-    //
     using RatioValue = std::conditional_t<Resampling::kIsConst, const gr::Size_t, gr::Size_t>;
     A<RatioValue, "numerator", Doc<"Top of resampling ratio (<1: Decimate, >1: Interpolate, =1: No change)">, Limits<1UL, std::numeric_limits<RatioValue>::max()>> numerator = Resampling::kNumerator;
     A<RatioValue, "denominator", Doc<"Bottom of resampling ratio (<1: Decimate, >1: Interpolate, =1: No change)">, Limits<1UL, std::numeric_limits<RatioValue>::max()>> denominator
@@ -430,11 +430,10 @@ public:
     using StrideValue = std::conditional_t<StrideControl::kIsConst, const gr::Size_t, gr::Size_t>;
     A<StrideValue, "stride", Doc<"samples between data processing. <N for overlap, >N for skip, =0 for back-to-back.">> stride = StrideControl::kStride;
 
-    //
-    std::size_t stride_counter = 0UZ;
+    gr::Size_t strideCounter = 0UL; // leftover stride from previous calls
 
     // TODO: These are not involved in move operations, might be a problem later
-    const std::size_t unique_id   = _unique_id_counter++;
+    const std::size_t unique_id   = _uniqueIdCounter++;
     const std::string unique_name = fmt::format("{}#{}", gr::meta::type_name<Derived>(), unique_id);
 
     //
@@ -484,42 +483,8 @@ public:
     };
     std::map<std::string, std::set<std::string>> propertySubscriptions;
 
-    struct PortsStatus {
-        std::size_t in_min_samples{ 1UZ };                                             // max of `port.min_samples()` of all input ports
-        std::size_t in_max_samples{ std::numeric_limits<std::size_t>::max() };         // min of `port.max_samples()` of all input ports
-        std::size_t in_available{ std::numeric_limits<std::size_t>::max() };           // min of `port.streamReader().available()` of all input ports
-        std::size_t nSamplesToNextTag{ std::numeric_limits<std::size_t>::max() };      // min distance to next Tag
-        std::size_t nSamplesToNextTagAfter{ std::numeric_limits<std::size_t>::max() }; // min distance after the next Tag
-        std::size_t nSamplesToEosTag{ std::numeric_limits<std::size_t>::max() };       // min of `port.samples_to_eos_tag` of all input ports
-
-        std::size_t out_min_samples{ 1UZ };                                     // max of `port.min_samples()` of all output ports
-        std::size_t out_max_samples{ std::numeric_limits<std::size_t>::max() }; // min of `port.max_samples()` of all output ports
-        std::size_t out_available{ std::numeric_limits<std::size_t>::max() };   // min of `port.streamWriter().available()` of all input ports
-
-        std::size_t in_samples{ 0UZ };  // number of input samples to process
-        std::size_t out_samples{ 0UZ }; // number of output samples, calculated based on `numerator` and `denominator`
-
-        bool in_at_least_one_port_has_data{ false }; // at least one port has data
-        bool in_at_least_one_tag_available{ false }; // at least one port has a tag
-
-        bool has_sync_input_ports{ false };  // if all ports are async, status is not important
-        bool has_sync_output_ports{ false }; // if all ports are async, status is not important
-
-        constexpr bool
-        enoughSamplesForOutputPorts(std::size_t n) {
-            return !has_sync_output_ports || n >= out_min_samples;
-        }
-
-        constexpr bool
-        spaceAvailableOnOutputPorts(std::size_t n) {
-            return !has_sync_output_ports || n <= out_available;
-        }
-    };
-
-    PortsStatus ports_status{};
-
 protected:
-    bool _output_tags_changed = false;
+    bool _outputTagsChanged = false;
     Tag  _mergedInputTag{};
 
     // intermediate non-real-time<->real-time setting states
@@ -533,66 +498,6 @@ protected:
     [[nodiscard]] constexpr const auto &
     self() const noexcept {
         return *static_cast<const Derived *>(this);
-    }
-
-    void
-    updatePortsStatus() {
-        ports_status = PortsStatus();
-        // TODO: recheck definition of denominator vs. numerator w.r.t. up-/down-sampling
-        //  ports_status.in_min_samples  = denominator;
-        //  ports_status.out_min_samples = numerator;
-
-        auto adjust_for_input_port = [&ps = ports_status]<PortLike Port>(Port &port) {
-            if constexpr (std::remove_cvref_t<Port>::kIsSynch) {
-                if (port.isConnected()) {
-                    ps.has_sync_input_ports          = true;
-                    ps.in_min_samples                = std::max(ps.in_min_samples, port.min_samples);
-                    ps.in_max_samples                = std::min(ps.in_max_samples, port.max_samples);
-                    ps.in_available                  = std::min(ps.in_available, port.streamReader().available());
-                    ps.in_at_least_one_port_has_data = ps.in_at_least_one_port_has_data | (port.streamReader().available() > 0);
-                    ps.in_at_least_one_tag_available = ps.in_at_least_one_port_has_data | (port.tagReader().available() > 0);
-                }
-            }
-            // if Async ports are present then we still want to process Tags fo these ports
-            if (port.isConnected()) {
-                ps.nSamplesToNextTag      = std::min(ps.nSamplesToNextTag, nSamplesUntilNextTag(port).value_or(std::numeric_limits<std::size_t>::max()));
-                ps.nSamplesToNextTagAfter = std::min(ps.nSamplesToNextTagAfter, nSamplesUntilNextTag(port, 1).value_or(std::numeric_limits<std::size_t>::max())); /* 1: in case nextTag == 0 */
-                ps.nSamplesToEosTag       = std::min(ps.nSamplesToEosTag, samples_to_eos_tag(port).value_or(std::numeric_limits<std::size_t>::max()));
-            }
-        };
-        for_each_port([&adjust_for_input_port](PortLike auto &port) { adjust_for_input_port(port); }, inputPorts<PortType::STREAM>(&self()));
-
-        auto adjust_for_output_port = [&ps = ports_status]<PortLike Port>(Port &port) {
-            if constexpr (std::remove_cvref_t<Port>::kIsSynch) {
-                ps.has_sync_output_ports = true;
-                ps.out_min_samples       = std::max(ps.out_min_samples, port.min_samples);
-                ps.out_max_samples       = std::min(ps.out_max_samples, port.max_samples);
-                ps.out_available         = std::min(ps.out_available, port.streamWriter().available());
-            }
-        };
-        for_each_port([&adjust_for_output_port](PortLike auto &port) { adjust_for_output_port(port); }, outputPorts<PortType::STREAM>(&self()));
-
-        ports_status.in_samples = ports_status.in_available;
-        if (ports_status.in_samples < ports_status.in_min_samples) ports_status.in_samples = 0;
-        if (ports_status.in_samples > ports_status.in_max_samples) ports_status.in_samples = ports_status.in_max_samples;
-
-        // By default N-in == N-out
-        // TODO: adjust `samples_to_proceed` to output limits?
-        ports_status.out_samples = ports_status.in_samples;
-
-        if (ports_status.has_sync_input_ports && ports_status.in_min_samples > ports_status.in_max_samples)
-            throw std::invalid_argument(fmt::format("Min samples for input ports ({}) is larger then max samples for input ports ({})", ports_status.in_min_samples, ports_status.in_max_samples));
-        if (ports_status.has_sync_output_ports && ports_status.out_min_samples > ports_status.out_max_samples)
-            throw std::invalid_argument(fmt::format("Min samples for output ports ({}) is larger then max samples for output ports ({})", ports_status.out_min_samples, ports_status.out_max_samples));
-
-        if (!ports_status.has_sync_input_ports) {
-            ports_status.in_samples   = 0;
-            ports_status.in_available = 0;
-        }
-        if (!ports_status.has_sync_output_ports) {
-            ports_status.out_samples   = 0;
-            ports_status.out_available = 0;
-        }
     }
 
 public:
@@ -618,11 +523,10 @@ public:
         , numerator(std::move(other.numerator))
         , denominator(std::move(other.denominator))
         , stride(std::move(other.stride))
-        , stride_counter(std::move(other.stride_counter))
+        , strideCounter(std::move(other.strideCounter))
         , msgIn(std::move(other.msgIn))
         , msgOut(std::move(other.msgOut))
-        , ports_status(std::move(other.ports_status))
-        , _output_tags_changed(std::move(other._output_tags_changed))
+        , _outputTagsChanged(std::move(other._outputTagsChanged))
         , _mergedInputTag(std::move(other._mergedInputTag))
         , _settings(std::move(other._settings)) {}
 
@@ -691,6 +595,7 @@ public:
             }
             notifyListeners(block::property::kSetting, settings().get());
         }
+        checkParametersAndThrowIfNeeded();
 
         // store default settings -> can be recovered with 'resetDefaults()'
         settings().storeDefaults();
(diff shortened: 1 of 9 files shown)


core/include/gnuradio-4.0/Port.hpp
```diff
@@ -955,7 +956,7 @@ template<typename T>
 concept TagPredicate = requires(const T &t, const Tag &tag, Tag::signed_index_type readPosition) {
     { t(tag, readPosition) } -> std::convertible_to<bool>;
 };
-inline constexpr TagPredicate auto defaultTagMatcher    = [](const Tag &tag, Tag::signed_index_type readPosition) noexcept { return tag.index >= readPosition || tag.index < 0; };
```
RalphSteinhagen, line 958: Why the drop of `tag.index < 0`? This used to indicate that a tag should be applied immediately regardless of other index constraints.

wirew0rm: This predicate was and is used to get the next matching tag and the matching tag after that in `nSamplesUntilNextTag(port, offset=0)`, where offset is the default of 0 for getting the first tag and 1 for getting the Tag not at position 1. But since `offset` ends up in `readPosition` of the predicate tag index only, as soon a there is  a tag with index < 0, it will also be return when looking for the second tag... I did not follow up more detailed how this worked previously, re-adding it makes qa_Block spin indefinitely.

Tag propagation and setting _cachedTag is not influenced by this since there the `-1` condition is checked separately in `Port::getTags(untilOffset)`.

I'm not completely sure on whether this is the correct change, this is just the rationale for removing it. I tried to keep the changes outside of the workInternal (and its subfunctions) small, as i've been bitten sometimes by doing extensive changes thoughout the codebase at too many places at once.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+                maxSync      = std::min(maxSync, port.max_samples);
+                maxAvailable = std::min(maxAvailable, available);
+            } else {                                 // async port
+                if (available >= port.min_samples) { // ensure that process function is called if at least one async port has data available
+                    hasAsync = true;
+                }
+            }
+        };
+        for_each_port([&adjust_for_input_port](PortLike auto &port) { adjust_for_input_port(port); }, std::forward<decltype(ports)>(ports));
+        return std::tuple{minSync, maxSync, maxAvailable, hasAsync};
```
drslebedev, line 1174: Would it be better if one returns a struct instead of tuple?
```cpp
struct PortLimits {
   std::size_t minSync{};
   std::size_t maxSync{};
   std::size_t maxAvailable{};
   bool hasAsync{};
};
```
And later one can do:
```cpp
const auto in = getPortLimits(inputPorts<PortType::STREAM>(&self()));
const auto out = getPortLimits(outputPorts<PortType::STREAM>(&self()));
```

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+        return std::numeric_limits<std::size_t>::max();
+    }
+
+    work::Status
+    invokeProcessBulk(auto &inputSpans, auto &writersTuple) {
+        // cannot use std::apply because it requires tuple_cat(inputSpans, writersTuple). The latter doesn't work because writersTuple isn't copyable.
+        return [&]<std::size_t... InIdx, std::size_t... OutIdx>(std::index_sequence<InIdx...>, std::index_sequence<OutIdx...>) {
+            return self().processBulk(std::get<InIdx>(inputSpans)..., std::get<OutIdx>(writersTuple)...);
+        }(std::make_index_sequence<std::tuple_size_v<std::remove_cvref_t<decltype(inputSpans)>>>(), std::make_index_sequence<std::tuple_size_v<std::remove_cvref_t<decltype(writersTuple)>>>());
+    }
```
RalphSteinhagen, line 1301: Not (yet) fully tested but the below could convert the present `std::vector<std::span>` or `std::vector<"ConsumableSpan">` to `std::span<std::span..> or `std::span<"ConsumableSpan">`

```cpp
work::Status invokeProcessBulk(auto &inputSpans, auto &writersTuple) {
    auto convertIfNeeded = [](auto&& arg) -> decltype(auto) {
        using T = std::decay_t<decltype(arg)>;
        if constexpr (std::is_same_v<T, std::vector<typename T::value_type>>>) {
            // convert std::vector to std::span of [Consumable,Producable]Spans
            return std::span<typename T::value_type>{arg.data(), arg.size()};
        } else if (...) {  // similar for std::array<...>
        } else { // pass-through for other types
            return std::forward<decltype(arg)>(arg);
        }
    };

    return [&]<std::size_t... InIdx, std::size_t... OutIdx>(std::index_sequence<InIdx...>, std::index_sequence<OutIdx...>) {
        return self().processBulk(convertIfNeeded(std::get<InIdx>(inputSpans))..., convertIfNeeded(std::get<OutIdx>(writersTuple))...);
    }(std::make_index_sequence<std::tuple_size_v<std::remove_cvref_t<decltype(inputSpans)>>>(), std::make_index_sequence<std::tuple_size_v<std::remove_cvref_t<decltype(writersTuple)>>>());
}
```

@drslebedev your view?

RalphSteinhagen: N.B. this way we do not expose if the dynamic input ports/spans are stored as vector or array and makes IMO a more maintainable and future-proof  API. Feedback is welcome.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+            nOutSamplesBeforeRequestedStop++;
+            // the block implementer can set `_outputTagsChanged` to true in `processOne` to prematurely leave the loop and apply his changes
+            if (_outputTagsChanged || lifecycle::isShuttingDown(this->state())) [[unlikely]] { // emitted tag and/or requested to stop
+                break;
+            }
+        }
+        if (nOutSamplesBeforeRequestedStop > 0) {
+            return std::tuple{work::Status::OK, nOutSamplesBeforeRequestedStop, nOutSamplesBeforeRequestedStop};
+        }
+        return std::tuple{work::Status::OK, processed, processed};
```
RalphSteinhagen, line 1354: return struct with named fields.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1312,7 +1374,35 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+     *       - MOVE_FW/MOVE_BW: move the tags to the first sample of the current/next chunk
+     *       - special case EOS tag: send incomplete chunk even if it violates work/block constraints -> implementations choose to drop/pad/...
+     *     - propagate tags:
+     *       - in the generic case the only tag in the current chunk is on the first sample
+     *       - different strategies, see TagPropagation
+     *   - settings
+     *      - apply and reset cached/merged tag
+     *   - get available samples count
+     *     - M: samples to consume on SYNC ports
+     *     - N: samples to produce on SYNC ports
```
RalphSteinhagen, line 1396: need easier/more descriptive names than 'M' and 'N'

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+     *   - maxAvailable: the maximum amount of that are available on all sync ports
+     *   - hasAsync: true if there is at least one async input/output that has available samples/remaining capacity
+     */
+    auto
+    getPortLimits(auto &&ports) {
+        std::size_t minSync               = 0UL;
+        std::size_t maxSync               = std::numeric_limits<std::size_t>::max();
+        std::size_t maxAvailable          = std::numeric_limits<std::size_t>::max();
+        bool        hasAsync              = false;
+        auto        adjust_for_input_port = [this, &minSync, &maxSync, &maxAvailable, &hasAsync]<PortLike Port>(Port &port) {
```
drslebedev, line 1155:
```suggestion
        auto        adjustForInputPort = [this, &minSync, &maxSync, &maxAvailable, &hasAsync]<PortLike Port>(Port &port) {
```

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1337,290 +1426,119 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+        const auto [minSyncIn, maxSyncIn, maxSyncAvailableIn, hasAsyncIn]     = getPortLimits(inputPorts<PortType::STREAM>(&self()));
+        const auto [minSyncOut, maxSyncOut, maxSyncAvailableOut, hasAsyncOut] = getPortLimits(outputPorts<PortType::STREAM>(&self()));
+        auto [hasTag, nextTag, nextEosTag, asyncEoS]                          = getTagPositions();
+        auto       maxChunk                                                   = getMergedBlockLimit(); // handle special cases for merged blocks. TODO: evaluate if/how we can get rid of these
+        const auto inputSkipBefore                                            = inputSamplesToSkipBeforeNextChunk(std::min({ maxSyncAvailableIn, nextTag, nextEosTag }));
+        const auto availableToProcess          = std::min({ maxSyncIn, maxChunk, (maxSyncAvailableIn - inputSkipBefore), (nextTag - inputSkipBefore), (nextEosTag - inputSkipBefore) });
+        const auto availableToPublish          = std::min({ maxSyncOut, maxSyncAvailableOut });
+        const auto [decimatedIn, decimatedOut] = decimate(minSyncIn, availableToProcess, minSyncOut, availableToPublish);
```
RalphSteinhagen, line 1437: I like the structured binding which makes the code more readable. Maybe these names could be also used in the generating methods w.r.t. naming which make those also more readable.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+     * Check the input ports for available samples
+     * @return tuple [nextTag, nextEosTag]
+     */
+    auto
+    getTagPositions() {
+        bool        hasTag                = false;
+        std::size_t nextTag               = std::numeric_limits<std::size_t>::max();
+        std::size_t nextEosTag            = std::numeric_limits<std::size_t>::max();
+        bool        asyncEoS              = false;
+        auto        adjust_for_input_port = [this, &hasTag, &nextTag, &nextEosTag, &asyncEoS]<PortLike Port>(Port &port) {
```
drslebedev, line 1187:
```suggestion
        auto        adjustForInputPort = [this, &hasTag, &nextTag, &nextEosTag, &asyncEoS]<PortLike Port>(Port &port) {
```

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+                    hasTag |= !tagData.empty() && tagData[0].index == port.streamReader().position() && !tagData[0].map.empty();
+                } else { // async port
+                    if (samples_to_eos_tag(port).transform([&port](auto n) { return n <= port.min_samples; }).value_or(false)) {
+                        asyncEoS = true;
+                    }
+                }
+            }
+        };
+        for_each_port([&adjust_for_input_port](PortLike auto &port) { adjust_for_input_port(port); }, inputPorts<PortType::STREAM>(&self()));
+        return std::tuple{hasTag, nextTag, nextEosTag, asyncEoS};
```
drslebedev, line 1204: Would it be better if one returns a struct instead of tuple? (see previous comment on getPortLimits())

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+                    inputSamplesToSkipAfter = std::min(static_cast<std::size_t>(stride.value), remaining);
+                    strideCounter           = stride.value - static_cast<gr::Size_t>(inputSamplesToSkipAfter);
+                }
+            }
+        }
+        return inputSamplesToSkipAfter;
+    }
+
+    auto
+    decimate(std::size_t minSyncIn, std::size_t maxSyncIn, std::size_t minSyncOut, std::size_t maxSyncOut) {
```
drslebedev, line 1251: Considering that it can do both decimation and interpolation, wouldn't it be more appropriate to rename it to resample()? This also aligns with other names.

RalphSteinhagen (approved):
@wirew0rm kudos, this was a tremendous amount of work to get it done this neatly and cleanly. :+1: 

The changes make the code much more readable and the now linear flow of processing w/o the earlier jumps easier to follow, to test and -- I believe -- also to maintain.

I added some small comments and suggestions for improvements. These are minor and the PR could from my point-of-view be merged as is, but I think these are quality items that would be beneficial and/or at least should be discussed/documented for a follow-up.

Main things are:
 * use of structs and named fields for function returns rather than std::tuple. This would complement and mirror the already used structured bindings.
 * use of `emitErrorMessage(..)` rather than throwing exceptions in core-lib code
 
 The others things and missing features you mentioned in the PR description could be done in a later follow-up.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+                if constexpr (StrideControl::kEnabled) { // with stride, we cannot process more than one chunk
+                    if (stride.value != 0 && stride.value != denominator) {
+                        nResamplingChunks = denominator <= maxSyncIn && numerator <= maxSyncOut ? 1 : 0;
+                    } else {
+                        nResamplingChunks = std::min(maxSyncIn / denominator, maxSyncOut / numerator);
+                    }
+                } else {
+                    nResamplingChunks = std::min(maxSyncIn / denominator, maxSyncOut / numerator);
+                }
+                if (nResamplingChunks * denominator < minSyncIn || nResamplingChunks * numerator < minSyncOut) {
```
drslebedev, line 1264: We might need to incorporate additional conditions to handle additional ill-formed parameters related to `numerator` and `denominator`. For instance, consider the scenario where `minSyncIn = 2`, `maxSyncIn = 3`, and `denominator = 2`. In this case, `nResamplingChunks = 1`, leading to a situation where it consistently returns {0, 0}, potentially resulting in an infinite loop.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+    work::Status
+    invokeProcessBulk(auto &inputSpans, auto &writersTuple) {
+        // cannot use std::apply because it requires tuple_cat(inputSpans, writersTuple). The latter doesn't work because writersTuple isn't copyable.
+        return [&]<std::size_t... InIdx, std::size_t... OutIdx>(std::index_sequence<InIdx...>, std::index_sequence<OutIdx...>) {
+            return self().processBulk(std::get<InIdx>(inputSpans)..., std::get<OutIdx>(writersTuple)...);
+        }(std::make_index_sequence<std::tuple_size_v<std::remove_cvref_t<decltype(inputSpans)>>>(), std::make_index_sequence<std::tuple_size_v<std::remove_cvref_t<decltype(writersTuple)>>>());
+    }
+
+    work::Status
+    invokeProcessOneSimd(auto &inputSpans, auto &writersTuple, auto width, std::size_t processed) {
```
drslebedev, line 1304:
```suggestion
    invokeProcessOneSimd(auto &inputSpans, auto &writersTuple, auto width, std::size_t nSamplesToProcess) {
```

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+                const auto results = simdize_tuple_load_and_apply(w, inputSpans, i, [&](auto &&...input_simds) { return invoke_processOne_simd(i, w, input_simds...); });
+                meta::tuple_for_each([i](auto &output_range, auto &result) { result.copy_to(output_range.data() + i, stdx::element_aligned); }, writersTuple, results);
+                i += w;
+            }
+        });
+        return work::Status::OK;
+    }
+
+    work::Status
+    invokeProcessOnePure(auto &inputSpans, auto &writersTuple, std::size_t processed) {
```
drslebedev, line 1321:
```suggestion
    invokeProcessOnePure(auto &inputSpans, auto &writersTuple, std::size_t nSamplesToProcess) {
```

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+        for (std::size_t i = 0; i < processed; ++i) {
+            auto results = std::apply([this, i](auto &...inputs) { return this->invoke_processOne(i, inputs[i]...); }, inputSpans);
+            meta::tuple_for_each([i]<typename R>(auto &output_range, R &&result) { output_range[i] = std::forward<R>(result); }, writersTuple, results);
+        }
+        return work::Status::OK;
+    }
+
+    auto
+    invokeProcessOneNonConst(auto &inputSpans, auto &writersTuple, std::size_t processed) {
```
drslebedev, line 1330:
```suggestion
    invokeProcessOneNonConst(auto &inputSpans, auto &writersTuple, std::size_t nSamplesToProcess) {
```

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1337,290 +1426,119 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
-        const auto inputSpans   = prepareInputStreams();
-        auto       writersTuple = prepareOutputStreams();
+        // for non-bulk processing, the processed span has to be limited to the first sample if it contains a tag s.t. the tag is not applied to every sample
+        const bool limitByFirstTag = (!HasProcessBulkFunction<Derived> && HasProcessOneFunction<Derived>) &&hasTag;
 
+        // call the block implementation's work function
+        const auto   inputSpans   = prepareStreams(inputPorts<PortType::STREAM>(&self()), limitByFirstTag ? 1 : decimatedIn);
+        auto         writersTuple = prepareStreams(outputPorts<PortType::STREAM>(&self()), limitByFirstTag ? 1 : decimatedOut);
+        work::Status ret;
+        std::size_t  processed    = limitByFirstTag ? 1 : decimatedIn;
```
drslebedev, line 1479:
```suggestion
        std::size_t  nSamplesToProcess    = limitByFirstTag ? 1 : decimatedIn;
```

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1337,290 +1426,119 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+        std::size_t  processedOut = limitByFirstTag ? 1 : decimatedOut;
```
drslebedev, line 1480:
```suggestion
        std::size_t  nSamplesToProcessOut = limitByFirstTag ? 1 : decimatedOut;
```

core/include/gnuradio-4.0/Block.hpp
drslebedev, line 1437:
```suggestion
        const auto [resampledIn, resampledOut] = decimate(minSyncIn, availableToProcess, minSyncOut, availableToPublish);
```

core/test/qa_Block.cpp
```diff
@@ -557,8 +556,8 @@ stride_test(const StrideTestData &data, std::shared_ptr<gr::thread_pool::BasicTh
     if (data.in_port_min >= 0) int_dec_block.in.min_samples = static_cast<size_t>(data.in_port_min);
 
     expect(eq(gr::ConnectionResult::SUCCESS, flow.connect<"out">(source).to<"in">(int_dec_block)));
-    auto sched = scheduler(std::move(flow), thread_pool);
-    sched.runAndWait();
+    auto sched = scheduler(std::move(flow), std::move(thread_pool));
+    expect(sched.runAndWait().has_value());
```
drslebedev, line 560: Not for this PR, but I think we need to add this `expect(sched.runAndWait().has_value());` for all test, potentially it saves debugging time as it was the case here.

drslebedev (approved):
Thanks for the PR! 
Really great work to simplify the code and make it clear and readable!
Since we already discussed your changes privately. I added some minor comments, please feel free to implement them.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1337,290 +1426,119 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
-                    }
-                    publishTag({ { gr::tag::END_OF_STREAM, true } }, 0);
-                    return { requested_work, ports_status.in_samples, work::Status::DONE };
-                }
-            }
-            write_to_outputs(ports_status.out_samples, writersTuple);
-            const bool success = consumeReaders(self(), nSamplesToConsume);
-            return { requested_work, ports_status.in_samples, success ? ret : work::Status::ERROR };
-
+            ret = invokeProcessBulk(inputSpans, writersTuple); // todo: evaluate how many where really produced...
```
frankosterfeld, line 1482: typo: where/were

core/include/gnuradio-4.0/Block.hpp
drslebedev: Using `std::span` for an array of ports as the signature for processBulk() looks good to me. 
`std::span<std::span..>` is better than `std::vector<std::span..>`.

However, I recommend postponing this to the next pull request, as it requires further modifications in the unit tests.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -818,69 +723,85 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+        }
+        if (minSyncOut > maxSyncOut) {
+            throw std::invalid_argument(fmt::format("Min samples for output ports ({}) is larger then max samples for output ports ({})", minSyncOut, maxSyncOut));
+        }
+        if (denominator > maxSyncIn) {
+            throw std::invalid_argument(fmt::format("resampling denominator ({}) is larger then max samples for input ports ({})", denominator, maxSyncIn));
+        }
+        if (numerator > maxSyncOut) {
+            throw std::invalid_argument(fmt::format("resampling numerator ({}) is larger then max samples for output ports ({})", numerator, maxSyncOut));
+        }
```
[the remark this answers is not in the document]
wirew0rm: :heavy_check_mark: done

core/include/gnuradio-4.0/Block.hpp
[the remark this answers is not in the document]
wirew0rm: :ballot_box_with_check: done

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+                        inputSamplesToSkipBefore = available;
+                        strideCounter -= static_cast<gr::Size_t>(available);
+                    } else {
+                        inputSamplesToSkipBefore = strideCounter;
+                        strideCounter            = 0;
+                    }
+                }
+            }
+        }
+        return inputSamplesToSkipBefore;
```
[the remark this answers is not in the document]
wirew0rm: looks good. I changed from the lambda to a simple bool, since it has no arguments and it is the first thing to be eavauated in the if so I didn't see the benefit of not directly evaluating the boolean expression. Let me know if I overlooked something there.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1292,6 +1137,223 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+                return std::tuple{static_cast<std::size_t>(nResamplingChunks * denominator), static_cast<std::size_t>(nResamplingChunks * numerator)};
+            } else {
+                std::size_t n = std::min(maxSyncIn, maxSyncOut);
+                return std::tuple{n, n};
+            }
+        } else { // no resampling
+            std::size_t n = std::min(maxSyncIn, maxSyncOut);
+            return std::tuple{n, n};
+        }
+    }
```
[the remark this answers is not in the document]
wirew0rm: done! I like the early returns and reduced nesting.
I thought about simplifying this before. With the first constexpr if, I was not sure if I can mix compile time and runtime expressions inside of one statement (denominator/numerator may be compile time constants or run-time member fields depending as chosen by the block author).

core/include/gnuradio-4.0/Block.hpp
wirew0rm: These are the only 2 vairables where I didn't manage to make them const, otherwise i'd agree with the rename, but these are also updated with the result of the processing...
What i'm missing here is the ability to define an uninitialized const variable and then have the compiler check that it will be assigned exactly once in every branch... Java does that, but in c++ the only way to do get something like that is to wrap it in a lambda and return from there, but that would make the code a lot less straightforward to read...

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -818,69 +723,85 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
-                        auto process_out = [available_values_count]<typename Out>(Out &out) {
+                    [nSamples]<typename OutputRange>(auto, OutputRange &outputRange) {
+                        auto processOneRange = [nSamples]<typename Out>(Out &out) {
                             if constexpr (Out::isMultiThreadedStrategy()) {
                                 if (!out.isFullyPublished()) {
-                                    fmt::print(stderr, "Block::write_to_outputs - did not publish all samples for MultiThreadedStrategy\n");
+                                    fmt::print(stderr, "Block::publishWriters - did not publish all samples for MultiThreadedStrategy\n");
                                     std::abort();
                                 }
                             }
```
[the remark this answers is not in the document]
wirew0rm: I know we had some discussion about this with @drslebedev I don't remember what what the outcome was there.

wirew0rm:
Addressed most of the review comments, open points:

- naming numerator/denominator
- moving publish logic into producable span destructor
- using span for dynamic ports

drslebedev:
> * naming numerator/denominator
> * moving publish logic into producable span destructor
> * using span for dynamic ports

At least second and third points should be addressed in the next PRs.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -807,80 +712,116 @@ class Block : public lifecycle::StateMachine<Derived>, //
             static_assert(HasProcessBulkFunction<Derived>, "Blocks which allow decimation/interpolation must implement processBulk(...) method. Remove 'ResamplingRatio<>' from the block definition.");
         } else {
             if (numerator != 1ULL || denominator != 1ULL) {
-                throw std::runtime_error(fmt::format("Block is not defined as `ResamplingRatio<>`, but numerator = {}, denominator = {}, they both must equal to 1.", numerator, denominator));
+                auto e = gr::Error(fmt::format("Block is not defined as `ResamplingRatio<>`, but numerator = {}, denominator = {}, they both must equal to 1.", numerator, denominator));
+                emitErrorMessage("Block::checkParametersAndThrowIfNeeded:", e);
+                requestStop();
```
frankosterfeld, line 717: Doesn't this return DONE where it should return ERROR?

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1337,290 +1451,127 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+            success                          = consumeReaders(resampledIn, inputSpans);
+            const auto inputSpansForSkipping = prepareStreams(inputPorts<PortType::STREAM>(&self()), inputSamplesToSkipAfter - resampledIn); // only way to consume is via the ConsumableSpan now
+            consumeReaders(inputSamplesToSkipAfter - resampledIn, inputSpansForSkipping);
+        } else {
+            success = consumeReaders(processedIn, inputSpans);
+        }
+        // if the block state changed to DONE, publish EOS tag on the next sample
+        if (ret == work::Status::DONE) {
+            this->setAndNotifyState(lifecycle::State::STOPPED);
+            publishTag({ { gr::tag::END_OF_STREAM, true } }, 1);
```
frankosterfeld, line 1570: Why index 1 here, and not 0 (like in other places?)?

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1337,290 +1451,127 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+            ret = work::Status::DONE;
```
frankosterfeld, line 1571: This line should be necessary (ret already DONE)

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1337,290 +1451,127 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
-                if (auto e = this->changeStateTo(lifecycle::State::STOPPED); !e) {
-                    emitErrorMessage("isShuttingDown -> STOPPED", e.error());
-                }
-                publishTag({ { gr::tag::END_OF_STREAM, true } }, 0);
-                return { requested_work, ports_status.in_samples, success ? work::Status::DONE : work::Status::ERROR };
+        } else { // block does not define any valid processing function
+            static_assert(gr::meta::always_false<gr::traits::block::stream_input_port_types_tuple<Derived>>, "neither processBulk(...) nor processOne(...) implemented");
+        }
+        forwardTags();
+        if (lifecycle::isShuttingDown(this->state()) || nextEosTag <= processedIn + 1) {
```
frankosterfeld, line 1543: Here something is going wrong for my Delay block (processBulk(ConsumableSpan, PublishableSpan): The block produces/consumes 0, but both `processedIn` and `processedOut` are `100`, which is the data available at the input port. This check then considers the EOS tag in range and finishes.

wirew0rm: Yes, see also this comment: https://github.com/fair-acc/graph-prototype/pull/297/files#diff-d0a6abe85a332778390dc74e1c498e4c580677fe46db76395cdf2afefc41fb1eR1507 Until now, this has not been a problem, because there was no block that relied on being able to not consume all input samples in process bulk. Probably the easiest and most generic way to fix it would be to remove the eosTag condition and only shutdown the block in the next call to process bulk...  have to check if this breaks something else.

core/include/gnuradio-4.0/Block.hpp
wirew0rm: there probably was a second condition in the if before... can be removed when fixing the eos behavior for process bulk that doesn't consume anything

core/include/gnuradio-4.0/Block.hpp
wirew0rm: The done tag always gets the index of the next sample (which will never be published). Otherwise a block could not publish EoS without publishing more samples. But you are right, that seems to be already considered in publishTag, it does not publish to the last written but to the next written index. Probably it doesn't matter because in the next iteration it sees the state and publishes another tag on the correct position... so maybe it should just be removed s.t. all eos tag handling is done at the beginning of the function...

PR #298, opened 2024-03-15
refactored Message passing to using MDP pattern

RalphSteinhagen:
* following similar to OpenCMW command structure https://github.com/fair-acc/opencmw-cpp/blob/main/docs/Majordomo_protocol_comparison.pdf that is derived from: https://rfc.zeromq.org/spec/7/ and https://rfc.zeromq.org/spec/18/
  * added user-facing API for `emitMessage(..)`, `emitErrorMessage(..)`, and `notifyListener(..)`
  * refactored scheduler to derive from Block and re-use the same message passing as blocks (N.B. scheduler will also need settings handling etc.)
  * added gr::exception which automatically tracks the message, source location, and time when the exception occurred
  * refactored `gr::lifecycle::ErrorType` to `gr::Error` and added `[std,gr]::exception` support
  * added property handling for
    * Heartbeat
    * Echo
    * LifeCycleState
    * [Staged]Setting
    * [Store, Reset]Defaults (used to recover/reset to default settings)
  * added basic subscription handler for [staged]settings and heartbeat
  
As a suggestion for reviewing: start with the 'qa_Messages' unit-test to get a feeling for the new API, then dive into the actual implementation.


core/test/qa_Messages.cpp
```diff
@@ -18,509 +20,633 @@ auto boost::ut::cfg<boost::ut::override> = boost::ut::runner<boost::ut::reporter
...
+
+        "Block<T>-level heartbeat tests"_test = [] {
+            gr::MsgPortOut toBlock;
+            TestBlock<int> unitTestBlock({ { "name", "UnitTestBlock" } });
+            gr::MsgPortIn  fromBlock;
+
+            expect(eq(ConnectionResult::SUCCESS, toBlock.connect(unitTestBlock.msgIn)));
+            expect(eq(ConnectionResult::SUCCESS, unitTestBlock.msgOut.connect(fromBlock)));
+
+            "w/o explicit serviceName"_test = [&] {
```
drslebedev, line 127: Would be good to add some documentation, that serviceName should be `""`, `Block::unique_name`, or `Block::name` or `unmatchedPropertyHandler` should be implemented.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1477,44 +1653,57 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
+                                  { self().unmatchedPropertyHandler(sv, m) } -> std::same_as<std::optional<Message>>;
+                              }) {
+                    callback = &Derived::unmatchedPropertyHandler;
                 }
+            }
 
-                Message settingsUpdated;
-                settingsUpdated[gr::message::key::Kind] = gr::message::kind::SettingsChangeRequested;
-                settingsUpdated[gr::message::key::Data] = settings().get();
+            if (callback == nullptr) {
```
drslebedev, line 1678: callback is also nullptr when cmd::Set is processed, probably one can also add this to the docs.

core/test/qa_Messages.cpp
```diff
@@ -18,509 +20,633 @@ auto boost::ut::cfg<boost::ut::override> = boost::ut::runner<boost::ut::reporter
...
+                expect(nothrow([&] { unitTestBlock.processScheduledMessages(); })) << "manually execute processing of messages";
+
+                expect(eq(fromBlock.streamReader().available(), 0UZ)) << "should not receive a reply";
+                expect(unitTestBlock.state() == lifecycle::State::INITIALISED);
+            };
+
+            "set - state - error cases"_test = [&] {
+                sendMessage<Set>(toBlock, "" /* serviceName */, block::property::kLifeCycleState /* endpoint */, {} /* no data  */);
+                expect(nothrow([&] { unitTestBlock.processScheduledMessages(); })) << "manually execute processing of messages";
+                expect(eq(fromBlock.streamReader().available(), 1UZ)) << "should not receive a reply";
```
drslebedev, line 297: There is no reply but error is still sent. Change to "should have one error".

drslebedev (approved):
Thank you very much for your pull request! 

I focused primarily on reviewing the API usage and `qa_Message.cpp`. The API appears to be very straightforward and expressive.  Additionally, `qa_Message` covers a broad range of use cases and can serve as valuable code examples for users. 

It would be beneficial to improve documentation for certain aspects, particularly covering corner cases and expected errors.

core/test/qa_Messages.cpp
RalphSteinhagen: done.

core/include/gnuradio-4.0/Block.hpp
RalphSteinhagen: The logic of that part of the code has changed: the 'callback' points to either the registered function for a given property or the 'unmatchedPropertyHandler(...)' that should be invoked. This is before any MDP command is being queried/executed.

RalphSteinhagen: Maybe the removed code was misleading... refactored this late because I found my original lambda-based logic needlessly complex.

drslebedev:
Despite CI build issues, which cannot be reproduced locally and have been present in previous PRs, we've decided to merge this PR regardless and address the failures separately.

PR #300, opened 2024-03-19
CI: enable uploading coredumps and coresponding binaries

wirew0rm:
This adds CI steps to enable the collection of coredump files and uploading them as an artefact together with the binary that caused them as an artefact for offline analysis.
The general steps where taken from
https://github.com/itamarst/gha-upload-cores
but adapted to our more complex repository setup with multiple executables in different directories.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,8 +87,14 @@ jobs:
         env:
           cache-name: cache-fetchContent-cache
         with:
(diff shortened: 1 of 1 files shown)


wirew0rm:
closing this for now, since setting up coredump collection inside the container is not as straightforward as I hoped. It should be generally possible, but needs additional configuration for docker and setting up paths correctly.

PR #302, opened 2024-03-21
Code cleanup and build improvements

wirew0rm:
Different commits factored out of #297 that are unrelated to the refactoring of the work function but fix warnings, unrelated test failures and improve the CI and project structure.

- CI: disable environment to reduce PR spam]
The coverage environment was originally used to store secrets for
coverage submission, but this is not used anymore but caused each gh
actions run to emit a non-collapsible message on the PR which can hide
important comments.
This commit removes the environment from the build step.
- Reduce the number of warnings and suggestions
mostly casts, unused variables, superfluous semicolons, etc
- Move qa_port_array from separate test into qa_Block
This was a lonely 20 lines test in a separate compilation unit which tested what now has become basic block functionality
- Move static graph tests into qa_Block
moved to the unittest to reduce the size of the block header. IMO lightweight static_asserts can be inline, but this was a significant chunk of code, which i also think should not have to be paid by every consumer of the header.
- qa_fourier: replace custom test block by generic one
count_source -> TagSource
- CircularBuffer: add isConsumed() to ConsumableSpan
thanks @drslebedev for the help with the consumable/producable span API.
- qa_sources: increase timeouts and use fatal assertions
expect does not immediately terminate by default, which leads to
segfaults later on in the test if e.g. a nullpointer is returned for a
poller due to nonsufficient timeouts. This leads to missleading asan
warnings when the nullpointer is later dereferenced when the test should
have already failed.


RalphSteinhagen (approved):
A very nice and well structured clean-up PR! :+1: 

@wirew0rm  I know how much time was spent on this but the elegance and ease to review this is a testimony to the quality you've put in! Thanks a lot!

This can be merged once the CI tests for the last-minute/final revision have passed.

PR #305, opened 2024-03-28
Block: fix EoS handling for process bulk not consuming samples

wirew0rm:
@frankosterfeld discovered a regression in one of the out of tree blocks in gr-digitizer, reported [here](https://github.com/fair-acc/graph-prototype/pull/297/files#r1542781655).

If the processBulk function does not consume all samples up to the eos tag, the previous code would terminate processing anyway, leading to the unprocessed samples being dropped.
This is becaue the ConsumableSpan does not provide a way to determine how much the processsing function has actually consumed.

This change solves this by only handling explicitly stopped blocks after the block has processed, which leaves the eos tag handling for the start of the next call, where it is sufficient to check if the eos tag is the first sample.

Also fixes a few minor inconsistencies discovered by @frankosterfeld.

### DIFF
--- a/core/include/gnuradio-4.0/Block.hpp
+++ b/core/include/gnuradio-4.0/Block.hpp
@@ -1559,7 +1559,7 @@ protected:
             static_assert(gr::meta::always_false<gr::traits::block::stream_input_port_types_tuple<Derived>>, "neither processBulk(...) nor processOne(...) implemented");
         }
         forwardTags();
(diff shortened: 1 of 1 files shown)


frankosterfeld:
LGTM! 👍
This fixes my test issues in opendigitizer's qa_GnuRadioWorker (still having issues related to setting changes, but that's another topic)

frankosterfeld (approved):

PR #306, opened 2024-04-02
ut: remove clang workaround after upstream fixes

wirew0rm:
UT fixed the initialisation of the command line arguments for clang in https://github.com/boost-ext/ut/commit/3cfd9937e1d4c9375f7d7904a5ba6a1926f7161f. This change removes the special case which disables the junit style ut reporter in most unit tests.

For qa_grc there is still a segfault if the junit reporter is enablet but this seems to be a slightly different case. The segfault occurs in the static initialisation of the block registy in gr::registerBlock.

### DIFF
--- a/blocks/basic/test/qa_DataSink.cpp
+++ b/blocks/basic/test/qa_DataSink.cpp
@@ -18,12 +18,6 @@
 
 using namespace std::chrono_literals;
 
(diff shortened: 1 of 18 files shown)


frankosterfeld (approved):
LGTM!

Issue #287, opened 2024-02-23
Make GRC load/save usable from emscripten


frankosterfeld:
To use the GRC serialization (Graph_yaml_importer.hpp) in emscripten (opendigitizer UI), it needs fixes. Currently it includes plugin.h, which in turn gnuradio-plugin_export.h, which isn't generated because the plugin library is disabled for emscripten.

frankosterfeld:
Fixed by https://github.com/fair-acc/graph-prototype/pull/308

PR #308, opened 2024-04-10
Enable GRC serialization for Emscripten

frankosterfeld:
Fixes issue #287 

Back to draft because it needs a patch to yaml-cpp for qa_grc to pass:

https://github.com/frankosterfeld/yaml-cpp/tree/fix/wasm-issue


RalphSteinhagen (approved):
Thanks for cleaning up and notably tracking down the the yaml- cpp fix! :+1: 

IMO can be merged once the CI passes.

PR #309, opened 2024-04-10
CI: use custom docker image for github actions

wirew0rm:
This builds a custom docker image which will be automatically updated on changes to the dockerfile. It includes all compilers and analysis tools necessary to build and test the project.
This should help local reproduction, reduce build times as it saves on fetching build dependencies in the generic image and makes updateing the image less dependent on image updates by github.

Since we want gcc14 and clang18 this has to use the ubuntu 24.04 development image. On the plus side, this saves us from using custom package repos for cmake and llvm (the code for enabling them is stil included as comments in case we require fresher versions).

The main CI gets a lot simpler by this since we do not have to care about installing anything. There is an additional pre-step now that checks if the Dockerfile was changed and if yes rebuilds the image and otherwise uses the existing image.

The sonarcube analysis is switched from using their wrappper to the more generic cmake compile_commands.json approach wich allows to treat all builds in the same way.


RalphSteinhagen (approved):
Thanks for the changes! Looked at it earlier. LGTM! :+1:

PR #310, opened 2024-04-11
Wrapped schduler.runAndWait() calls

drslebedev:
Wrapped `schduler.runAndWait()` calls with `expect()`  to ensure return value presence.


wirew0rm (approved):
Thanks for following this up, looks good and can be merged once the CI passes.

Having some consistent formatting will be nice, I trust that all changes that are not return value checks are format-only.

PR #314, opened 2024-04-17
ConsumableInputRange and buffer_reader updates:

drslebedev:
This PR contains the following changes:
* `ConsumableInputRange` counter
* Deferred `consume()`, consume is executed in destructor of the last range using proper `SpanReleasePolicy`
* Default `SpanReleasePolicy` is `ProcessNone`
* `reader.get()` by default returns all available samples
* Multiple consecutive `reader.get()` calls are possible, `nSamples` of the first `get()` call is the max for all subsequent calls.
* Remove `consume()` method from `buffer_reader`, consume is possible only via `ConsumableInputRange.consume()`

### DIFF
--- a/core/include/gnuradio-4.0/Block.hpp
+++ b/core/include/gnuradio-4.0/Block.hpp
@@ -801,7 +801,7 @@ public:
             meta::tuple_for_each_enumerate(
                     [nSamples, &success]<typename InputRange>(auto, InputRange &inputRange) {
                         auto processOneRange = [nSamples, &success]<typename In>(In &in) {
-                            if (!in.isConsumed()) {
+                            if (!in.isConsumeRequested()) {
                                 using enum gr::SpanReleasePolicy;
                                 if constexpr (In::spanReleasePolicy() == Terminate) {
                                     fmt::print(stderr, "Block::consumeReaders - samples were not consumed, default SpanReleasePolicy is {}\n", magic_enum::enum_name(Terminate));
@@ -1179,12 +1179,13 @@ protected:
     auto
     getPortLimits(P &&ports) {
         struct {
-            std::size_t minSync = 0UL;                                          // the minimum amount of samples that the block needs for processing on the sync ports
-            std::size_t maxSync = std::numeric_limits<std::size_t>::max();      // the maximum amount of that can be consumed on all sync ports
+            std::size_t minSync      = 0UL;                                     // the minimum amount of samples that the block needs for processing on the sync ports
+            std::size_t maxSync      = std::numeric_limits<std::size_t>::max(); // the maximum amount of that can be consumed on all sync ports
             std::size_t maxAvailable = std::numeric_limits<std::size_t>::max(); // the maximum amount of that are available on all sync ports
-            bool hasAsync = false;                                              // true if there is at least one async input/output that has available samples/remaining capacity
+            bool        hasAsync     = false;                                   // true if there is at least one async input/output that has available samples/remaining capacity
         } result;
-        auto        adjustForInputPort  = [&result]<PortLike Port>(Port &port) {
+
+        auto adjustForInputPort = [&result]<PortLike Port>(Port &port) {
             const std::size_t available = [&port]() {
                 if constexpr (gr::traits::port::is_input_v<Port>) {
                     return port.streamReader().available();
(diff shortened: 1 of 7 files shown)


core/include/gnuradio-4.0/CircularBuffer.hpp
```diff
@@ -715,32 +764,23 @@ class CircularBuffer
...
-            if constexpr (strict_check) {
-                if (nSamples <= 0) {
-                    _isRangeConsumed = true;
-                    return true;
-                }
-                if (nSamples > available()) {
-                    _isRangeConsumed = true;
-                    return false;
-                }
+            std::size_t nSamples = std::min(available(), nRequested);
```
RalphSteinhagen, line 776: I'd suggest to make this call to `available()` conditional only if `nRequested != std::numeric_limits<std::size_t>::max()`. The call to available involves atomics which are not cache friendly.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1212,20 +1213,22 @@ class Block : public lifecycle::StateMachine<Derived>, //
...
                     // get the tag after the one at position 0 that will be evaluated for this chunk.
                     // nextTag limits the size of the chunk except if this would violate port constraints
                     result.nextTag                        = std::min(result.nextTag, nSamplesUntilNextTag(port, 1).value_or(std::numeric_limits<std::size_t>::max()));
                     result.nextEosTag                     = std::min(result.nextEosTag, samples_to_eos_tag(port).value_or(std::numeric_limits<std::size_t>::max()));
-                    const gr::ConsumableSpan auto tagData = port.tagReader().get(port.tagReader().available());
-                    result.hasTag =  result.hasTag || (!tagData.empty() && tagData[0].index == port.streamReader().position() && !tagData[0].map.empty());
+                    const gr::ConsumableSpan auto tagData = port.tagReader().get();
+                    result.hasTag                         = result.hasTag || (!tagData.empty() && tagData[0].index == port.streamReader().position() && !tagData[0].map.empty());
+                    std::ignore                           = tagData.consume(0UZ);
```
RalphSteinhagen, line 1231: Couldn't this `consume(0UZ)` be avoided by setting the appropriate policy on the calling `get<SpanReleasePolicy::ProcessNone>()'?

RalphSteinhagen (approved):
Thanks for the PR. The changes and unit-test look consistent. 

Provided @wirew0rm checked/confirmed the changes to the `Block<T>` that potentially may affect the resampling behaviour (he has the better overview on this right now), there are just two minor comments/suggestions (see above) after which this PR could be merged IMO.

core/include/gnuradio-4.0/CircularBuffer.hpp
drslebedev: I think it is unavoidable: `nRequested == std::numeric_limits<std::size_t>::max()` means return all available samples -> `nSamples = available()`.

core/include/gnuradio-4.0/Block.hpp
drslebedev: This line is optional as long as default policy is `ProcessNone`. I remove it.

wirew0rm (approved):
Looks good from the perspective of the block logic :+1: 
Found one possible simplification, but I'm not sure if there are still cases where it is needed, see below.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -801,7 +801,7 @@ class Block : public lifecycle::StateMachine<Derived>, //
             meta::tuple_for_each_enumerate(
                     [nSamples, &success]<typename InputRange>(auto, InputRange &inputRange) {
                         auto processOneRange = [nSamples, &success]<typename In>(In &in) {
-                            if (!in.isConsumed()) {
+                            if (!in.isConsumed() && !in.isConsumeRequested()) {
```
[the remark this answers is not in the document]
drslebedev: You are right, assuming that one has only deferred consume which happens only in the destructor of the last range, `isConsumed()` is always false. I will update the code accordingly.

core/include/gnuradio-4.0/CircularBuffer.hpp
drslebedev: the default parameter is `nRequested = std::numeric_limits<std::size_t>::max()` which basically returns number of available samples.

drslebedev: One can introduce stick_check template parameter, and if it is false one can remove check with available()

PR #317, opened 2024-04-22
CMake: add pch support

wirew0rm:
Adds precompiled header support for gnuradio-core headers. For more information see:
- https://cmake.org/cmake/help/latest/command/target_precompile_headers.html
- https://discourse.cmake.org/t/how-to-share-precompiled-headers-via-an-interface-library/9933

As it stands this does not improve the build time, the pch is built but not used for compiling the targets, needs some more investigation into the cmake logic.


wirew0rm:
Closing since it was decided to first focus on optimising the general structure of the code and there are no immediate ideas on how to make this work.

PR #318, opened 2024-04-22
wrap user-defined function exceptions to error messages

RalphSteinhagen:
While library internal code should be kept exception free, users (often using other C/C++ libraries) are free to throw exceptions to indicate rare errors. 

These exceptions are internally wrapped and passed up the chain to the graph->scheduler. 

In case the scheduler 'msgOut' is not connected these error messages are in turn converted to exceptions again.


drslebedev (approved):
Thanks a lot for the PR! This is a good step towards better error handling!
The implementation looks solid to me.

I particularly appreciate the `emitErrorMessageIfAny` function, which helps maintain cleaner code. 
This PR is ready to be merged.

PR #321, opened 2024-04-23
Fixes for OD refactoring

frankosterfeld:
Small fixes needed for the opendigitizer refactoring.

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
@@ -257,7 +257,7 @@ using DefaultClockSource = ClockSource<T, true, std::chrono::system_clock, true>
 ENABLE_REFLECTION_FOR_TEMPLATE_FULL((typename T, bool useIoThread, typename ClockSourceType), (gr::basic::ClockSource<T, useIoThread, ClockSourceType>), out, n_samples_max, chunk_size, sample_rate,
                                     verbose_console);
 
-auto registerClockSource = gr::registerBlock<gr::basic::DefaultClockSource, float, double>(gr::globalBlockRegistry());
+auto registerClockSource = gr::registerBlock<gr::basic::DefaultClockSource, std::uint8_t, std::uint32_t, std::int32_t, float, double>(gr::globalBlockRegistry());
 static_assert(gr::HasProcessBulkFunction<gr::basic::ClockSource<float>>);
 
 #endif // GNURADIO_CLOCK_SOURCE_HPP
--- a/blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
@@ -15,6 +15,7 @@
 #include <fmt/ranges.h>
 
 #include <gnuradio-4.0/Block.hpp>
+#include <gnuradio-4.0/BlockRegistry.hpp>
 #include <gnuradio-4.0/reflection.hpp>
 #include <gnuradio-4.0/Tag.hpp>
 
--- a/blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
@@ -249,13 +249,14 @@ private:
     }
 };
 
+template<typename T>
+using DefaultClockSource = ClockSource<T, true, std::chrono::system_clock, true>;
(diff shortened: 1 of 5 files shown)


core/include/gnuradio-4.0/DataSet.hpp
```diff
@@ -108,8 +108,8 @@ static_assert(DataSetLike<DataSet<float>>, "DataSet<float> concept conformity");
 static_assert(DataSetLike<DataSet<double>>, "DataSet<double> concept conformity");
 
 // public type definitions to allow simple reflection
-using DataSet_float  = DataSet<double>;
-using DataSet_double = DataSet<float>;
+using DataSet_double = DataSet<double>;
```
RalphSteinhagen, line 111: Let's get rid of this entirely since this not used in the lib.

blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
```diff
@@ -249,13 +250,14 @@ ClockSource Documentation -- add here
     }
 };
 
+template<typename T>
+using DefaultClockSource = ClockSource<T, true, std::chrono::system_clock, true>;
```
RalphSteinhagen, line 254: Do we need this short-hand definition? Not used elsewhere and introduces a named object that we need to document then.

RalphSteinhagen (approved):
OK, please use the extra `using ...` declaration since they are not used within the lib and -- due to their indirection -- do not contribute to making the code more easily readable.

@frankosterfeld you may squash-merge the changes automatically once the fixes are in and CI/CD passed.

blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
frankosterfeld: It's not easily possible to remove these, as these are registered
    ENABLE_REFLECTION(gr::DataSet_float, ...)

That way when registering a block with T=DataSet_float, one can pass "DataSet_float" to the PluginLoader and the right handler is found to instantiate the block. With `ENABLE_REFLECTION_FOR_TEMPLATE(gr::DataSet)`, the type name used is `gr::DataSet<Ts...>` (that's what the meta::type_name returns for reflectable templates), so there's no handler for e.g. "gr::DataSet<float>" found.

frankosterfeld: Sorry, this was referring to DataSet_float and DataSet_double.

We need this because gr::registerBlock cannot handle types with additional, non-NTTP, template parameters, like FourierAlgorithm for FFT and ClockSourceType for ClockSource.

core/include/gnuradio-4.0/DataSet.hpp
frankosterfeld: It's not easily possible to remove these, as these are registered as `ENABLE_REFLECTION(gr::DataSet_float, ...)`

That way when registering a block with T=DataSet_float, one can pass "DataSet_float" to the PluginLoader and the right handler is found to instantiate the block. With `ENABLE_REFLECTION_FOR_TEMPLATE(gr::DataSet)`, the type name used is `gr::DataSet<Ts...>` (that's what the meta::type_name returns for reflectable templates), so there's no handler for e.g. "gr::DataSet<float>" found.

RalphSteinhagen:
```cpp
ENABLE_REFLECTION_FOR_TEMPLATE(gr::DataSet, timestamp, axis_names, axis_units, axis_values, extents, layout, signal_names, signal_units, signal_values, signal_errors, signal_ranges, meta_information, timing_events)
```
... is your friend.

blocks/basic/include/gnuradio-4.0/basic/clock_source.hpp
RalphSteinhagen: Causality is a bit wrong: we should change the signature of the `ClockSource` mixing template, NTTP, template, NTTP, ... isn't a good pattern. OK for now. Needs to be refactored later.

Added and pushed also `uint8_t` and `[u]int32_t` since these are important from a timing context.

core/include/gnuradio-4.0/DataSet.hpp
frankosterfeld: That's what i tried, but that gives `gr::DataSet<Ts...>` as name. This will also hit us when adding type support to load_grc()/ save_grc(), maybe we should solve then.

PR #340, opened 2024-05-08
Core: rework Tag handling

wirew0rm:
The automatic tag forwarding implemented in Block::workInternal was was not working for ASYNC ports since in this case the Block cannot determine a priori if samples will be consumed and consequently cannot perform an early return ahead of the unconditional tag forwarding and consumption.

This change introduces an extension of the ConsumableInputRange, which additionally takes care of consuming all tags up to the first sample of the currently processed chunk.

Also all tag merging logic is moved from Port to Block, while the logic for preparing the consumableInput or producableOutput ranges is moved into the Port.


blocks/basic/test/qa_DataSink.cpp
```diff
@@ -667,7 +667,7 @@ const boost::ut::suite DataSinkTests = [] {
         expect(eq(poller->drop_count.load(), 0UZ));
     };
 
-    "blocking multiplexed mode"_test = [] {
+    skip / "blocking multiplexed mode"_test = [] {
```
wirew0rm, line 670: This test blocks in `runAndWait`, I have to investigate some more.

RalphSteinhagen: Still blocking?

blocks/basic/include/gnuradio-4.0/basic/tag_handling.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <gnuradio-4.0/Block.hpp>
+#include <gnuradio-4.0/BlockRegistry.hpp>
+#include <gnuradio-4.0/DataSet.hpp>
+#include <gnuradio-4.0/reflection.hpp>
+
+namespace gr::basic {
+
+template<typename T>
+struct RealignTagsToChunks : gr::Block<RealignTagsToChunks<T>> {
```
RalphSteinhagen, line 12: Is this a left-over from testing? Maybe we should move this to the unit-test or document/test this more thoroughly.

core/include/gnuradio-4.0/BlockTraits.hpp
```diff
@@ -311,6 +311,12 @@ struct DummyConsumableSpan {
 };
 static_assert(ConsumableSpan<DummyConsumableSpan<int>>);
 
+template<typename T>
+struct DummyConsumablePortSpan: public DummyConsumableSpan<T> {
```
RalphSteinhagen, line 315: The tags etc should be absorbed in the `concept ConsumableSpan` etc.

wirew0rm: I'm not sure I understand.
There are 2 concepts `ConsumableSpan`(no tag handling) and `ConsumablePortSpan`(the former + added tag handling). `DummyConsumableSpan` and `DummyConsumablePortSpan` are mock implementations of the respective concepts needed by the machinery of `Block::prepareStreams`.

core/include/gnuradio-4.0/Sequence.hpp
```diff
@@ -26,7 +26,7 @@ using std::hardware_destructive_interference_size;
 inline constexpr std::size_t hardware_destructive_interference_size  = 64;
 inline constexpr std::size_t hardware_constructive_interference_size = 64;
 #endif
-static constexpr const std::make_signed_t<std::size_t> kInitialCursorValue = -1L;
+static constexpr const std::make_signed_t<std::size_t> kInitialCursorValue = 0L;
```
drslebedev, line 29: 👍

core/include/gnuradio-4.0/CircularBuffer.hpp
```diff
@@ -285,118 +285,121 @@ class CircularBuffer
...
+        }
 
-    [[nodiscard]] constexpr bool
-    isFullyPublished() const noexcept {
-        return _parent->_internalSpan.size() == _parent->_nSamplesPublished;
-    }
+        [[nodiscard]] constexpr bool
+        isPublished() const noexcept {
+            return _parent->_isRangePublished;
+        }
```
drslebedev, line 366: It looks like you are using wrong clang-format. The return type must be on the same line. This appear in several places. Please check.

core/test/qa_Settings.cpp
```diff
@@ -71,13 +76,15 @@ struct Source : public Block<Source<T>> {
...
+    void
+    settingsChanged(const property_map & /*oldSettings*/, property_map &newSettings, property_map &fwdSettings) {
         // optional init function that is called after construction and whenever settings change
         newSettings.insert_or_assign("n_samples_max", n_samples_max);
         fwdSettings.insert_or_assign("n_samples_max", n_samples_max);
     }
 
-    [[nodiscard]] constexpr T processOne() noexcept {
+    [[nodiscard]] constexpr T
+    processOne() noexcept {
```
drslebedev, line 87: It looks like you are using wrong clang-format. The return type must be on the same line. This appear in several places. Please check.

core/test/qa_buffer.cpp
```diff
@@ -76,15 +76,15 @@ const boost::ut::suite BasicConceptsTests = [] {
...
-                consumableInputRangeTest3(static_cast<std::span<const int32_t>>(data));
+                //{
+                //    auto value = writer.reserve(1);
+                //    expect(eq(1LU, value.size())) << "for " << typeName;
+                //    value.publish(1);
+                //}
+                //consumableInputRangeTest1(data);
+                //consumableInputRangeTest2(data);
+                //// consumableInputRangeTest3(data); // this does not compile
+                //consumableInputRangeTest3(static_cast<std::span<const int32_t>>(data));
```
drslebedev, line 87: Is this code is still valid? If not please remove.

drslebedev (approved):
Thanks for the great PR and all your effort in making all the unit tests run smoothly. 👍

The changes look good to me. We’ve already discussed some modifications privately, and I worked with this update while attempting to fix a failing test!

I’ve added a few minor comments and suggestions for improvements. The PR can be merged after reviewing the comments.

blocks/basic/test/qa_DataSink.cpp
```diff
@@ -631,7 +631,7 @@ const boost::ut::suite DataSinkTests = [] {
...
```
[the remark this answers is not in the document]
wirew0rm: changed, thanks for catching this

core/include/gnuradio-4.0/CircularBuffer.hpp
```diff
@@ -654,8 +665,8 @@ class CircularBuffer
     [[nodiscard]] constexpr const T* data() const noexcept { return _internalSpan.data(); }
     const T& operator [](std::size_t i) const noexcept  {return _internalSpan[i]; }
     const T& operator [](std::size_t i) noexcept { return _internalSpan[i]; }
-    operator const std::span<const T>&() const noexcept { return _internalSpan; }
-    operator std::span<const T>&() noexcept { return _internalSpan; }
+    explicit(false) operator const std::span<const T>&() const noexcept { return _internalSpan; }
+    explicit(false) operator std::span<const T>&() noexcept { return _internalSpan; }
```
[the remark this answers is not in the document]
wirew0rm: Code analyzers will complain on implicit conversion operators and ask you  to make them expclicit. We have that at multiple occasions, the original idea was that this should suppress these warnings (I think it does for nodiscard(false)). While it does not actually suppress the warning, it will still hopefully deter someone going through all sonartype issues to just make it explicit which would be a breaking change.

core/test/qa_Settings.cpp
```diff
@@ -20,41 +20,46 @@ namespace gr::setting_test {
 
 namespace utils {
 
-std::string format_variant(const auto& value) noexcept {
+std::string
```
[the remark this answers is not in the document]
wirew0rm: I only applied clang format to the files which had previously been formatted with it. I'm afraid that doing too heavy reformatting will cause conflicts with the other in-flight PRs. We'll need to do a full tree reformatting + enabling restyled.io at some point with not too many open PRs in the near future. I could try to only reformat new lines, but I'm not sure it's worth the effort.

core/test/qa_buffer.cpp
wirew0rm: reenabled, seems to only have caused trouble in some intermediate state. Thanks for catching this!

PR #342, opened 2024-05-13
feature: propagate `DONE` to and disconnect from upstream blocks

RalphSteinhagen:
1. When a block detects that it is no longer required (i.e., it has no downstream consumers), this feature ensures that the block declares itself 'DONE' and severs connections to upstream blocks. This feature is controlled by the Block<T>::disconnect_on_done parameter (default: true)

  2. added `buffer().n_writers()` interface tracking the number of writers and when a buffer and/or two or more ports are not connected anymore. Makes the first item simpler/cleaner.

  3. added some documented default blocks in-line with GR3: NullSource, NullSink, Copy, HeadBlock, CountingSource, CountingSink -- naming of the 'HeadBlock' -- while ubiquitously used in GR3 -- should be discussed with the architecture team because of semantics.

### DIFF
--- a/core/test/qa_Scheduler.cpp
+++ b/core/test/qa_Scheduler.cpp
@@ -1,6 +1,9 @@
 #include <boost/ut.hpp>
 
 #include <gnuradio-4.0/Scheduler.hpp>
+#include <gnuradio-4.0/testing/NullSources.hpp>
+
+#include <chrono>
 
 using TraceVectorType = std::vector<std::string>;
 
@@ -259,8 +262,8 @@ ENABLE_REFLECTION_FOR_TEMPLATE(LifecycleSource, out);
 
 template<typename T>
 struct LifecycleBlock : public gr::Block<LifecycleBlock<T>> {
-    gr::PortIn<T>  in{};
-    gr::PortOut<T> out{};
+    gr::PortIn<T>                in{};
+    gr::PortOut<T, gr::Optional> out{};
 
     int process_one_count{};
     int start_count{};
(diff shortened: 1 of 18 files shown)


daniestevez:
Great! I've tested this branch with my `Head` block in https://github.com/daniestevez/gr4-packet-modem/tree/test-gnuradio4-pr342 and DONE propagation works as expected. A `NullSource -> Head(100) -> VectorSink` flowgraph terminates and at the end the `VectorSink` contains 100 items (see `examples/head.cpp`).

core/test/qa_Scheduler.cpp
```diff
@@ -259,8 +262,8 @@ ENABLE_REFLECTION_FOR_TEMPLATE(LifecycleSource, out);
 
 template<typename T>
 struct LifecycleBlock : public gr::Block<LifecycleBlock<T>> {
-    gr::PortIn<T>  in{};
-    gr::PortOut<T> out{};
+    gr::PortIn<T>                in{};
+    gr::PortOut<T, gr::Optional> out{};
```
drslebedev, line 266: Great job identifying the problem! We should remember to mark ports as `Optional` if they are not in use.

drslebedev (approved):
Thanks a lot for the PR!

Also, thanks to @vimpostor for fixing the HttpBlock problem.

This PR looks solid, and the new DONE mechanism for disconnecting upstream blocks when no longer required works as expected. The unit tests effectively cover potential use cases.

The new blocks are well-documented. 

Here are two name suggestions for HeadBlock: FirstNSamples, SampleLimiter.

Great job overall!

ivan-cukic (approved):

PR #345, opened 2024-05-21
compile-time improvements - for_each std::tuple -> std::array

RalphSteinhagen:
* eliminated looping over a tuple and for_each (expensive to instantiate) and to use a straight `std::array<std::string_view, ...> kDefaultTags` instead. The latter can also be evaluated during compile-time.
  * added ccache support (can be disabled) which should help for local development that needs to switch between branches.
  
Comparative timings for the qa_Scheduler unit-tests:

Before (full parallel build):
```text
real    8m43.031s
user    51m13.146s
sys     2m13.891s
```
ClangBuildAnalyzer output for `qa_scheduler` [before](https://github.com/fair-acc/gnuradio4/files/15384756/qa_scheduler_analysis_before.txt)

After:
```text
real    6m52.364s
user    42m5.309s
sys     1m41.661s
```
ClangBuildAnalyzer output for `qa_scheduler`  [after](https://github.com/fair-acc/gnuradio4/files/15384765/qa_scheduler_analysis_after.txt)

Visualisation of the remaining traces:
image

  * remaining compile-time hot spots:
     * `BlockRegistry.hpp:25:16`: lambda to instantiate each block for a given stream type
     * `Graph.hpp:323:31`: nested lambdas in `BlockWrapper::createDynamicPortsLoader()`
     * `fmt::format_to ... rva::variant<...>...`: pmt-related printouts
  * some preliminary findings:
     * clang performs 2x better than gcc w.r.t. compile-times but produces similar (the same) asm output
     * (especially recursive) lambdas seem to be more compile-time expensive than free-standing functions/methods
     * looping over `std::array` is cheaper as over `std::tuple` (where applicable).
     * initial pre-compiled headers do not seem to have a large positive effect (at least on the chosen unit-test)
     
### Some thoughts, ideas, and next steps:
 
 * [ ] simplify `BlockRegistry` and `BlockWrapper` to only support minimal API (notably constructors)
 * [ ] test roll-out `<gnuradio-4.0/core.hpp>` PCH -> goal: one include dependency for blocklib developer
 * [ ] use `__attribute__((noinline))` to prevent non-performance critical code being inlined/optimised on global scope
 * [ ] roll out at least three additional pipe-lines for: `-fsanitize=address`, `-fsanitize=undefined`, `-fsanitize=thread` (combine where possible)
 * [ ] centrally suppress plugin system for local/non-runtime builds (e.g. for unit-tests, static/compile-time graphs) ... tbd.
 * ...
 
raw `-ftime-trace` traces:
[qa_Scheduler.cpp.before.json.zip](https://github.com/fair-acc/gnuradio4/files/15384968/qa_Scheduler.cpp.before.json.zip)
[qa_Scheduler.cpp.after.json.zip](https://github.com/fair-acc/gnuradio4/files/15384969/qa_Scheduler.cpp.after.json.zip)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,7 @@ jobs:
             cxx: g++-14
           - cc: clang-18,
             cxx: clang++-18
-            cmake_flags: "-DCMAKE_LINKER=/usr/bin/clang-17"
+            cmake_flags: "-DCMAKE_LINKER=/usr/bin/clang-18"
           - cmake_wrapper: emcmake
             cc: emcc
             cmake_flags: "-DENABLE_COVERAGE=OFF -DCMAKE_CROSSCOMPILING_EMULATOR=${SYSTEM_NODE}"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,17 @@ if (NOT DEFINED GR_TOPLEVEL_PROJECT)
     endif ()
 endif ()
 
+# Use ccache if found and enabled
+find_program(CCACHE_PROGRAM ccache)
+option(USE_CCACHE "Use ccache if available" ON)
+if (CCACHE_PROGRAM AND USE_CCACHE)
+    message(STATUS "ccache found and will be used")
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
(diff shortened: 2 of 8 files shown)


wirew0rm (approved):
Looks good, just for comparison I also looked up the build times on the CI:

before:
image
after:
image

Impressive 25% reduction for gcc-14(Debug)! Interesting that there's almost no effect on the other compilers, especially since you used clang tooling to determine the hotspots.

PR #346, opened 2024-05-22
Implement CtxSettings, use instead of BasicSettings.

drslebedev:
This PR includes the implementation of `CtxSettings`, which will  replace the existing `BasicSettings`. 

Several tasks need to be addressed following this PR to fully integrate `CtxSettings` into the codebase:
1) Remove Base Settings Class and `BasicSettings`. The old `BasicSettings` and its base class need to be removed to eliminate redundancy and avoid confusion.
2) Eliminate usage of `unique_ptr` for Block::_settings
3) Update function names to clearly indicate they are associated with `CtxSettings`, enhancing code readability.
4) Expand Unit Testing. Add comprehensive unit tests specifically targeting `CtxSettings` .
5)  `using MatchPredicate = std::function<std::optional<bool>(const pmtv::pmt &, const pmtv::pmt &, std::size_t)>;`
replace this with the new matcher structure. See [here](https://github.com/fair-acc/gnuradio4/blob/filterStreamToDataSet/core/include/gnuradio-4.0/TriggerMatcher.hpp) for details.

### DIFF
--- a/core/include/gnuradio-4.0/Settings.hpp
+++ b/core/include/gnuradio-4.0/Settings.hpp
@@ -21,10 +21,39 @@
 namespace gr {
 
 namespace settings {
+
+template<typename T>
+inline constexpr static bool
+isSupportedVectorType() {
+    if constexpr (gr::meta::vector_type<T>) {
+        return std::is_arithmetic_v<typename T::value_type> || std::is_same_v<typename T::value_type, std::string>;
+    } else {
+        return false;
+    }
+}
+
 template<typename T>
 inline constexpr static bool
 isSupportedType() {
-    return std::is_arithmetic_v<T> || std::is_same_v<T, std::string> || gr::meta::vector_type<T> || std::is_same_v<T, property_map>;
+    return std::is_arithmetic_v<T> || std::is_same_v<T, std::string> || isSupportedVectorType<T>() || std::is_same_v<T, property_map>;
+}
+
+template<typename T, typename TMember>
+inline constexpr static bool
+isWritableMember(TMember member) {
+    return traits::port::is_not_any_port_or_collection<T> && !std::is_const_v<T> && is_writable(member) && settings::isSupportedType<T>();
+}
+
+template<typename T, typename TMember>
+inline constexpr static bool
+isReadableMember(TMember member) {
+    return traits::port::is_not_any_port_or_collection<T> && is_readable(member) && settings::isSupportedType<T>();
+}
+
+inline constexpr uint64_t
+convertTimePointToUint64Ns(const std::chrono::time_point<std::chrono::system_clock> &tp) {
+    const auto ns = std::chrono::duration_cast<std::chrono::nanoseconds>(tp.time_since_epoch()).count();
+    return static_cast<uint64_t>(ns);
 }
 } // namespace settings
 
@@ -43,17 +72,8 @@ hash_combine(std::size_t &seed, const T &v) noexcept {
 } // namespace detail
 
 struct SettingsCtx {
-    // using TimePoint = std::chrono::time_point<std::chrono::utc_clock>; // TODO: change once the C++20 support is ubiquitous
-    using TimePoint               = std::chrono::time_point<std::chrono::system_clock>;
-    std::optional<TimePoint> time = std::nullopt; /// UTC time-stamp from which the setting is valid
-    property_map             context;             /// user-defined multiplexing context for which the setting is valid
-
-    SettingsCtx() {}
-
-    explicit SettingsCtx(const TimePoint &t, const property_map &ctx = {}) {
-        time    = t;
-        context = ctx;
-    }
+    uint64_t  time    = 0ULL; // UTC-based time-stamp in ns, time from which the setting is valid, 0U is undefined time
+    pmtv::pmt context = "";   // user-defined multiplexing context for which the setting is valid
 
     bool
     operator==(const SettingsCtx &) const
@@ -62,19 +82,16 @@ struct SettingsCtx {
     bool
     operator<(const SettingsCtx &other) {
         // order by time
-        return !time || (other.time && *time < *other.time);
+        return time < other.time;
     }
 
     [[nodiscard]] std::size_t
     hash() const noexcept {
         std::size_t seed = 0;
-        if (time) {
-            detail::hash_combine(seed, time.value().time_since_epoch().count());
+        if (time != 0ULL) {
+            detail::hash_combine(seed, time);
         }
-        for (const auto &[key, val] : context) {
-            detail::hash_combine(seed, key);
-            detail::hash_combine(seed, pmtv::to_base64(val));
-        }
+        detail::hash_combine(seed, pmtv::to_base64(context));
         return seed;
     }
 };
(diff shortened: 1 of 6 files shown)


core/test/qa_Settings.cpp
```diff
@@ -546,8 +544,8 @@ const boost::ut::suite SettingsCtxTests = [] {
         SettingsCtx a;
         SettingsCtx b;
         expect(a == b);
-        a.time = std::chrono::system_clock::now();
-        b.time = std::chrono::system_clock::now() + std::chrono::seconds(1);
+        a.setTimePoint(std::chrono::system_clock::now());
```
RalphSteinhagen, line 547: As public API, on order to be exposable via Python the `setTimePoint(..)` interface should take the canonical `uint64_t` UTC timestamp in [ns].

core/test/qa_Settings.cpp
```diff
@@ -558,58 +556,126 @@ const boost::ut::suite TransactionTests = [] {
...
+        auto  settings = CtxSettings(block);
+        expect(block.settings().applyStagedParameters().forwardParameters.empty());
+        block.settings().storeDefaults();
+
+        const auto timeNow = std::chrono::system_clock::now();
+        // Store t = 0, t = 2, t = 4
+        // Test t = -1, t = 0, t = 1, t = 2, t = 3, t = 4, t = 5, t = nullopt
+        auto ctx0 = SettingsCtx(timeNow);
+        expect(settings.set({ { "name", "TestName10" }, { "scaling_factor", 10.f } }, ctx0).empty()) << "successful set returns empty map";
+        auto ctx2 = SettingsCtx(timeNow + std::chrono::seconds(2));
```
RalphSteinhagen, line 571: Public API: time -> `uint64_t` UTC timestamp [ns]

core/test/qa_grc.cpp
```diff
@@ -28,14 +28,14 @@ ENABLE_REFLECTION_FOR_TEMPLATE(ArraySource, outA, outB);
...
 struct ArraySink : public gr::Block<ArraySink<T>> {
-    std::array<gr::PortIn<T>, 2>                                     inA;
-    std::array<gr::PortIn<T>, 2>                                     inB;
-    gr::Annotated<bool, "bool setting">                              bool_setting = false;
-    gr::Annotated<std::string, "String setting">                     string_setting;
-    gr::Annotated<std::vector<bool>, "Bool vector setting">          bool_vector;
-    gr::Annotated<std::vector<std::string>, "String vector setting"> string_vector;
-    gr::Annotated<std::vector<double>, "Double vector setting">      double_vector;
-    gr::Annotated<std::vector<int16_t>, "int16_t vector setting">    int16_vector;
+    std::array<gr::PortIn<T>, 2>                                     inA{};
```
RalphSteinhagen, line 31: This `{}` initialiser should not be necessary.

core/include/gnuradio-4.0/Settings.hpp
```diff
@@ -21,16 +21,39 @@
...
+inline constexpr static bool
+isReadableMember(TMember member) {
+    return traits::port::is_not_any_port_or_collection<T> && is_readable(member) && settings::isSupportedType<T>();
 }
 } // namespace settings
 
 struct ApplyStagedParametersResult {
-    property_map forwardParameters; // parameters that should be forwarded to dependent child blocks
-    property_map appliedParameters;
+    property_map forwardParameters{}; // parameters that should be forwarded to dependent child blocks
```
RalphSteinhagen, line 55: `{}` should not be necessary as the property_map is default initialised to being empty.

core/include/gnuradio-4.0/Settings.hpp
```diff
@@ -44,37 +67,45 @@ hash_combine(std::size_t &seed, const T &v) noexcept {
...
 
     SettingsCtx() {}
 
-    explicit SettingsCtx(const TimePoint &t, const property_map &ctx = {}) {
+    explicit SettingsCtx(uint64_t t, const pmtv::pmt &ctx = "") {
         time    = t;
         context = ctx;
     }
 
+    explicit SettingsCtx(const TimePoint &tp, const pmtv::pmt &ctx = "") {
```
RalphSteinhagen, line 81: Please remove this constructor. functionally identical to the one above and widens the public API by a non-Python wrappable constructor.

Actually, do we need a dedicated constructor? :thinking:

core/include/gnuradio-4.0/Transactions.hpp
```diff
@@ -42,20 +42,46 @@ class CtxSettings : public SettingsBase {
...
-    TBlock                                       *_block = nullptr;
-    mutable std::mutex                            _lock{};
-    property_map                                  _active{};
-    property_map                                  _staged{};
-    std::set<std::string, std::less<>>            _auto_update{};
-    std::set<std::string, std::less<>>            _auto_forward{};
-    std::unordered_map<SettingsCtx, property_map> _settings{};
-    property_map                                  _default_settings{};
-    MatchPredicate                                _match_pred = nullMatchPred;
+    using MatchPredicate = std::function<std::optional<bool>(const pmtv::pmt &, const pmtv::pmt &, std::size_t)>;
```
RalphSteinhagen, line 45: N.B. the follow-up PR should replace this with the new matcher structure. See [here](https://github.com/fair-acc/gnuradio4/blob/filterStreamToDataSet/core/include/gnuradio-4.0/TriggerMatcher.hpp) for details.

core/include/gnuradio-4.0/Transactions.hpp
```diff
@@ -42,20 +42,46 @@ class CtxSettings : public SettingsBase {
...
+                // TODO: throw if types are not the same?
+                return lhs.index() < rhs.index();
+            } else {
+                return std::visit(
+                        [&](const auto &left) -> bool {
+                            using T = std::decay_t<decltype(left)>;
+                            if constexpr ((pmtv::String<T> || pmtv::Scalar<T>) && !pmtv::Complex<T>) {
+                                return left < std::get<T>(rhs);
+                            } else {
+                                throw std::runtime_error("Invalid CtxSettings context type " + std::string(typeid(T).name()));
```
RalphSteinhagen, line 62: replace with `gr::exception(...)`. 

We should aim at minimising throwing exceptions in the core lib core in favour of using `std::expected<..>`.

RalphSteinhagen (approved):
This looks good. Thanks for rationalising the CtxSettings and implementing the missing features. There are only a few minor comments (limiting/reducing public API surface) that should be addressed before this can be merged.


Next step/PR should eliminate the old basic settings implementation in favour of having only one/CtxSettings. During that follow-up step, please be also mindful about the compile-time complexity and check with clang's `-ftime-trace' output.

core/test/qa_Settings.cpp
drslebedev: This change was made to facilitate unit testing. I removed the functionality from the `SettingsCtx` context and added function for conversions instead.

core/include/gnuradio-4.0/Settings.hpp
drslebedev: It was used  to facilitate unit testing. I removed it.

core/include/gnuradio-4.0/Transactions.hpp
drslebedev: Added to the list of follow-up issues.

PR #348, opened 2024-05-24
Updates of the  clang-format code style

drslebedev:
This PR includes:
* Updates of the  clang-format code style.
* Code formatting section to developer guidelines.


.clang-format
```diff
@@ -1,227 +1,14 @@
...
-Standard: c++20
-TabWidth: 4
-UseTab: Never
-
+AccessModifierOffset: -4
+IndentRequiresClause: false
+AlignConsecutiveAssignments: true
+AlignConsecutiveDeclarations: true
+AlignConsecutiveMacros: true
+MaxEmptyLinesToKeep: 1
```
RalphSteinhagen, line 14: missing empty line at the end of the file.

RalphSteinhagen (approved):
As discussed, I agree with the proposed changes. I added two statements to make simple require clauses a bit more terse and one to enforce braces, notably after short `if(...)` statements to make it functionally safer. If you agree this could be squash-merged as is.

While there is some cognitive psychology research supporting this formatting choice, it remains a contentious issue. Reaching a consensus among developers and users is unlikely, and compromises may lead to suboptimal outcomes.

Therefore, once merged, I propose a one-year moratorium (until August 2025) on further changes unless there are strong technical reasons. This will allow developers to adjust to the new code style. We can review the pros and cons and make necessary adjustments then.

Our primary goal should be to avoid bike-shedding and focus on high-level algorithms and correctness during PR/code reviews.

PR #349, opened 2024-05-27
Improvements to the block registry

ivan-cukic:


core/include/gnuradio-4.0/BlockRegistry.hpp
```diff
@@ -82,6 +92,25 @@ class BlockRegistry {
         return _block_types;
     }
 
+    bool
+    isBlockKnown(std::string_view block) const {
+        return _block_type_handlers.find(std::string(block)) != _block_type_handlers.end();
+    }
+
+    auto
+    knownBlockParametrizations(std::string_view block) const {
```
RalphSteinhagen, line 101: spelling: 'parameterisation' or 'parameterization' ... otherwise, we are para-metricizing the blocks. ;-)

core/include/gnuradio-4.0/PluginLoader.hpp
```diff
@@ -140,6 +140,21 @@ class PluginLoader {
     BlockRegistry           *_registry;
     std::vector<std::string> _knownBlocks;
 
+    gr_plugin_base *
+    handlerForName(std::string_view name) const {
+        auto it = _handlerForName.find(std::string(name));
+        if (it == _handlerForName.end()) {
+            fmt::print("Error: No handler for name {}, available ", name);
```
RalphSteinhagen, line 147: API: should either be an `std::expected<>` (preferred, if this is used more often), throw a `gr::exception(..)`, but not be a print-out. This won't be visible and/or actionable by the developer/user.

core/include/gnuradio-4.0/PluginLoader.hpp
```diff
@@ -192,16 +207,31 @@ class PluginLoader {
         return result;
     }
 
+    std::vector<std::string>
+    knownBlockParametrizations(std::string_view block) const {
```
RalphSteinhagen, line 211: as above...

core/include/gnuradio-4.0/PluginLoader.hpp
```diff
@@ -192,16 +207,31 @@ class PluginLoader {
...
     instantiate(std::string_view name, std::string_view type, const property_map &params = {}) {
         // Try to create a node from the global registry
         if (auto result = _registry->createBlock(name, type, params)) {
             return result;
         }
-        auto it = _handlerForName.find(std::string(name)); // TODO avoid std::string here
-        if (it == _handlerForName.end()) return {};
 
-        auto &handler = it->second;
+        auto *handler = handlerForName(name);
```
RalphSteinhagen, line 231: try to avoid 'auto'. An explicit strong/named type would be better here.

core/include/gnuradio-4.0/Port.hpp
```diff
@@ -843,6 +843,10 @@ class DynamicPort {
         [[nodiscard]] virtual bool
         updateReaderInternal(InternalPortBuffers buffer_other) noexcept
                 = 0;
+
+        [[nodiscard]] virtual std::string
+        valueTypeName() const
```
RalphSteinhagen, line 848: Do we need this as mandatory public-API, or is this more a UI convenience function to disentangle types?

We need to shrink/clean-up the public-API and I am not sure whether this is needed as part of the Port interface. Stand-alone helper function?

RalphSteinhagen (changes requested):
Needs to be rebased to the main branch. 

There are some public API extensions that we should minimise. Notably the naming of types is a fairly generic requirement and does not have to be part of the `Port` API.

core/include/gnuradio-4.0/Port.hpp
ivan-cukic: We need to be able to tell the type of a dynamic port. Needs to be here as it is a part of type erasure of a typed port.

core/include/gnuradio-4.0/PluginLoader.hpp
ivan-cukic: Removed output. The function already returns nullptr on error,  wrapping in expected would not bring much to the table.

core/include/gnuradio-4.0/Port.hpp
RalphSteinhagen: There is an existing `std::any Port<...>::defaultValue()` method.  It's type erased and can be used at the lib and user-end to derive the value type of the port.

Could we use that instead?

RalphSteinhagen (approved):
OK. Thanks for the changes!

PR #351, opened 2024-05-27
Auto update policy when context is not found in stored settings.

drslebedev:
This PR contains implementation of the `AutoUpdatePolicy` when context is not found in stored settings.
Two possible options are available: `AddToStored` and `Ignore`.

### DIFF
--- a/core/include/gnuradio-4.0/Transactions.hpp
+++ b/core/include/gnuradio-4.0/Transactions.hpp
@@ -30,6 +30,16 @@ namespace gr {
 
 static auto nullMatchPred = [](auto, auto, auto) { return std::nullopt; };
 
+namespace settings {
+/**
+ * Policy for handling Settings Auto Update when `context` is not found in storedParameters.
+ */
+enum class AutoUpdatePolicy {
+    Ignore,     // do nothing when context is not in storedParameters
+    AddToStored // Add the new context to storedParameters with empty parameters and apply any new changes present in the Tag
+};
+} // namespace settings
+
 template<typename TBlock>
 class CtxSettings : public SettingsBase {
     /**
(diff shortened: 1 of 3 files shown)


core/include/gnuradio-4.0/Transactions.hpp
```diff
@@ -77,8 +87,9 @@ class CtxSettings : public SettingsBase {
     std::set<std::string, std::less<>>                                                         _allWritableMembers{};   // all `isWritableMember` class members
     std::map<pmtv::pmt, std::set<std::string, std::less<>>, PMTCompare>                        _autoUpdateParameters{}; // for each SettingsCtx.context auto updated members are store separately
     std::set<std::string, std::less<>>                                                         _autoForwardParameters{};
-    MatchPredicate                                                                             _matchPred = nullMatchPred;
-    pmtv::pmt                                                                                  _activeCtx = "";
+    MatchPredicate                                                                             _matchPred        = nullMatchPred;
+    pmtv::pmt                                                                                  _activeCtx        = "";
+    settings::AutoUpdatePolicy                                                                 _autoUpdatePolicy = settings::AutoUpdatePolicy::AddToStored;
```
RalphSteinhagen, line 92: This is not accessible via a public API. We should discuss whether this should be part of the block configuration and/or eventually dropped.

RalphSteinhagen (approved):
OK, before merging. Could you also change the default setting implementation `Block<T>::_settings` in the default constructor from 'BasicSettings? back to `CtxSettings` and verify that the 'qa_UI_Integration' unit-tests works as expected. Thanks in advance.

RalphSteinhagen (approved):
OK

PR #353, opened 2024-05-30
new block: BasicFile[Sink, Source]

RalphSteinhagen:
... nomen est omen.

@xaratustrah enjoy!

### DIFF
--- a/blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
+++ b/blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
@@ -1,0 +1,307 @@
+#ifndef BASICFILEIO_HPP
+#define BASICFILEIO_HPP
+
+#include <fmt/chrono.h>
+#include <fmt/format.h>
+#include <gnuradio-4.0/Block.hpp>
+#include <gnuradio-4.0/BlockRegistry.hpp>
+#include <magic_enum.hpp>
+
+#include <chrono>
+#include <complex>
+#include <filesystem>
+#include <fstream>
+#include <span>
+#include <string_view>
+
+namespace gr::blocks::fileio {
+
+namespace detail {
+[[nodiscard]] inline std::string getIsoTime() noexcept {
+    std::chrono::system_clock::time_point now = std::chrono::system_clock::now();
+    return fmt::format("{:%Y-%m-%dT%H:%M:%S}.{:06}",               // ms-precision ISO time-format
+        fmt::localtime(std::chrono::system_clock::to_time_t(now)), //
+        std::chrono::duration_cast<std::chrono::milliseconds>(now.time_since_epoch()).count() % 1'000);
+}
+
+inline void ensureDirectoryExists(const std::filesystem::path& filePath) { std::filesystem::create_directories(filePath.parent_path()); }
+
+inline std::vector<std::filesystem::path> getSortedFilesContaining(const std::string& fileName) {
+    std::filesystem::path filePath(fileName);
+    if (!std::filesystem::exists(filePath.parent_path())) {
+        throw gr::exception(fmt::format("path/file '{}' does not exist.", fileName));
+    }
+
+    std::vector<std::filesystem::path> matchingFiles;
+    std::copy_if(std::filesystem::directory_iterator(filePath.parent_path()), std::filesystem::directory_iterator{}, std::back_inserter(matchingFiles), [&](const auto& entry) { return entry.is_regular_file() && entry.path().string().find(filePath.filename().string()) != std::string::npos; });
+
+    std::sort(matchingFiles.begin(), matchingFiles.end());
+    return matchingFiles;
+}
+
+[[nodiscard]] inline std::uintmax_t getFileSize(const std::filesystem::path& filePath) {
+    if (!std::filesystem::exists(filePath)) {
+        throw std::runtime_error(fmt::format("file '{}' does not exist.", filePath.string()));
+    }
+    return std::filesystem::file_size(filePath);
+}
+
+[[maybe_unused]] inline std::vector<std::string> deleteFilesContaining(const std::string& fileName) {
+    std::filesystem::path filePath(fileName);
+    if (!std::filesystem::exists(filePath.parent_path())) {
+        return {};
+    }
+
+    std::vector<std::string> deletedFiles;
+    for (const auto& entry : std::filesystem::directory_iterator(filePath.parent_path())) {
+        if (entry.is_regular_file() && entry.path().string().find(filePath.filename().string()) != std::string::npos) {
+            deletedFiles.push_back(entry.path().string());
+            std::filesystem::remove(entry.path());
+        }
+    }
+
+    return deletedFiles;
+}
+
+} // namespace detail
+
+enum class FileMode { override, append, multi };
+
+template<typename T>
+struct BasicFileSink : public gr::Block<BasicFileSink<T>> {
+    using Description = Doc<R""(A sink block for writing a stream to a binary file.
+The file can be played back using a 'BasicFileSource' or read by any program that supports binary files (e.g. Python, C, C++, MATLAB).
+For complex types, the binary file contains [float, double]s in IQIQIQ order. No metadata is included with the binary data.)
+Important: this implementation assumes a host-order, CPU architecture specific byte order!)"">;
+    template<typename U, gr::meta::fixed_string description = "", typename... Arguments>
+    using A = gr::Annotated<U, description, Arguments...>; // optional shortening
+
+    PortIn<T> in;
+
+    A<std::string, "file name", Doc<"base filename, prefixed if ">, Visible>              file_name;
+    FileMode                                                                              _mode              = FileMode::override;
+    A<std::string, "mode", Doc<"mode: \"override\", \"append\", \"multi\"">, Visible>     mode               = std::string(magic_enum::enum_name(_mode));
+    A<gr::Size_t, "max bytes per file", Doc<"max bytes per file, 0: infinite ">, Visible> max_bytes_per_file = 0U;
+
+    std::size_t   _totalBytesWritten{0UZ};
+    std::size_t   _totalBytesWrittenFile{0UZ};
+    std::ofstream _file;
+    std::size_t   _fileCounter{0UZ};
+    std::string   _actualFileName;
+
+    void settingsChanged(const property_map& /*oldSettings*/, const property_map& /*newSettings*/) {
+        _mode = magic_enum::enum_cast<FileMode>(mode, magic_enum::case_insensitive).value_or(_mode);
+        if (lifecycle::isActive(this->state())) {
+            closeFile();
+            openNextFile();
+        }
+    }
+
+    void start() {
+        _totalBytesWritten = 0UZ;
+        openNextFile();
+    }
+
+    void stop() { closeFile(); }
+
+    [[nodiscard]] constexpr work::Status processBulk(ConsumableSpan auto& dataIn) {
+        std::size_t nBytesMax = dataIn.size() * sizeof(T);
+        if (max_bytes_per_file.value != 0U) {
+            nBytesMax = std::min(nBytesMax, static_cast<std::size_t>(max_bytes_per_file.value - _totalBytesWrittenFile));
+        }
+        _file.write(reinterpret_cast<const char*>(dataIn.data()), static_cast<std::streamsize>(nBytesMax));
+        if (!dataIn.consume(nBytesMax / sizeof(T))) {
+            throw gr::exception("could not consume input samples");
+        }
+
+        if (!_file) {
+            throw gr::exception(fmt::format("failed to write to file '{}'.", _actualFileName));
+        }
+
+        _totalBytesWritten += nBytesMax;
+        _totalBytesWrittenFile += nBytesMax;
+        if (max_bytes_per_file.value != 0U && _totalBytesWrittenFile >= max_bytes_per_file.value) {
+            closeFile();
+            openNextFile();
+        }
+
+        return work::Status::OK;
+    }
+
+private:
+    void closeFile() {
+        if (_file.is_open()) {
+            _file.flush();
+            _file.close();
+        }
+    }
+    void openNextFile() {
+        closeFile();
+        _totalBytesWrittenFile = 0UZ;
+
+        detail::ensureDirectoryExists(file_name.value);
+
+        std::filesystem::path filePath(file_name.value);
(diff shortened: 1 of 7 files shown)


blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
```diff
@@ -0,0 +1,307 @@
...
+Important: this implementation assumes a host-order, CPU architecture specific byte order!)"">;
+    template<typename U, gr::meta::fixed_string description = "", typename... Arguments>
+    using A = gr::Annotated<U, description, Arguments...>; // optional shortening
+
+    PortIn<T> in;
+
+    A<std::string, "file name", Doc<"base filename, prefixed if ">, Visible>              file_name;
+    FileMode                                                                              _mode              = FileMode::override;
+    A<std::string, "mode", Doc<"mode: \"override\", \"append\", \"multi\"">, Visible>     mode               = std::string(magic_enum::enum_name(_mode));
+    A<gr::Size_t, "max bytes per file", Doc<"max bytes per file, 0: infinite ">, Visible> max_bytes_per_file = 0U;
```
daniestevez, line 84: Maybe it's good to enforce that `max_bytes_per_file % sizeof(T) == 0`. Otherwise it seems that awkward things will happen, as items will be split over different files and lose alignment.

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        return work::Status::OK;
+    }
+
+private:
+    void closeFile() {
+        if (_file.is_open()) {
+            _file.flush();
+            _file.close();
```
daniestevez, line 135: Doesn't `close()` already perform a flush?

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
```diff
@@ -0,0 +1,307 @@
...
+            _actualFileName = filePath.parent_path() / (detail::getIsoTime() + "_" + std::to_string(_fileCounter++) + "_" + filePath.filename().string());
+            _file.open(_actualFileName, std::ios::binary);
+            break;
+        }
+        default: throw gr::exception("unsupported file mode.");
+        }
+
+        if (!_file) {
+            throw gr::exception(fmt::format("failed to open file '{}'.", _actualFileName));
+        }
```
daniestevez, line 170: I'm not sure if I'm right, but doesn't `_file.open()` throw an exception if the file cannot be opened? That exception would currently not be caught by this function, so it seems better to open the file in a `try` block, catch the exception, and re-raise it as a `gr::exception`.

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
```diff
@@ -0,0 +1,307 @@
...
+        if (!_file) {
+            throw gr::exception(fmt::format("failed to write to file '{}'.", _actualFileName));
+        }
+
+        _totalBytesWritten += nBytesMax;
+        _totalBytesWrittenFile += nBytesMax;
+        if (max_bytes_per_file.value != 0U && _totalBytesWrittenFile >= max_bytes_per_file.value) {
+            closeFile();
+            openNextFile();
+        }
```
daniestevez, line 126: If this `processBulk()` call is the last one (there is no more data after it) and a file was just filled up, this will create a new empty file which will never have data. It seems better to do this check at the beginning of `processBulk()` rather than at the end: if we cannot write any data because the current file is full, we create a new file and use that instead. That logic guarantees that we don't end up with empty files.

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
RalphSteinhagen: yes. Force of habit ... just to make sure. ;-)

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
```diff
@@ -0,0 +1,307 @@
...
+For complex types, the binary file contains [float, double]s in IQIQIQ order. No metadata is expected in the binary data.
+Important: this implementation assumes a host-order, CPU architecture specific byte order!)"">;
+
+    template<typename U, gr::meta::fixed_string description = "", typename... Arguments>
+    using A = gr::Annotated<U, description, Arguments...>; // optional shortening
+
+    PortOut<T> out;
+
+    A<std::string, "file name", Doc<"Base filename, prefixed if necessary">, Visible>          file_name;
+    FileMode                                                                                   _mode        = FileMode::override;
```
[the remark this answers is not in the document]
RalphSteinhagen: Yes, agree with the reasonin but didn't want to create a second enum class.

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
```diff
@@ -0,0 +1,307 @@
...
+    A<gr::Size_t, "max bytes per file", Doc<"max bytes per file, 0: infinite ">, Visible> max_bytes_per_file = 0U;
+
+    std::size_t   _totalBytesWritten{0UZ};
+    std::size_t   _totalBytesWrittenFile{0UZ};
+    std::ofstream _file;
+    std::size_t   _fileCounter{0UZ};
+    std::string   _actualFileName;
+
+    void settingsChanged(const property_map& /*oldSettings*/, const property_map& /*newSettings*/) {
+        _mode = magic_enum::enum_cast<FileMode>(mode, magic_enum::case_insensitive).value_or(_mode);
```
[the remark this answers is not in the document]
RalphSteinhagen: Yes, correct. If it's thrown then an exception may be thrown that is intercepted by `Block<T>` (as all exceptions are in user-defined functions), converted to an error message and forwarded via the scheduler back to the user.

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
```diff
@@ -0,0 +1,307 @@
...
+    template<typename U, gr::meta::fixed_string description = "", typename... Arguments>
+    using A = gr::Annotated<U, description, Arguments...>; // optional shortening
+
+    PortOut<T> out;
+
+    A<std::string, "file name", Doc<"Base filename, prefixed if necessary">, Visible>          file_name;
+    FileMode                                                                                   _mode        = FileMode::override;
+    A<std::string, "mode", Doc<"mode: \"override\", \"append\", \"multi\"">, Visible>          mode         = std::string(magic_enum::enum_name(_mode));
+    A<bool, "repeat", Doc<"true: repeat back-to-back">>                                        repeat       = false;
+    A<gr::Size_t, "offset", Doc<"file start offset in bytes">, Visible>                        offset       = 0U;
```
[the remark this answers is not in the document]
RalphSteinhagen: Some raw binary files have a header information (usually provided in bytes) which need to be skipped for each file. the length is to read only the first N samples from each file. This is useful for chunked data (together with the start trigger) like packet data or triggered acquisitions.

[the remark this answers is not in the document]
daniestevez: Good point. I guess there are two potential use cases for `offset`:
- Replay only a time segment of a recording (very useful in combination with `length`).
- Skip the header of a file.
These can be handled in the same way for a single file, but they lead to conflicting requirements with the `multi` option.

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
RalphSteinhagen: The alignment is respected in any case because the counting is done in samples i.e. `sizeof(T)`. The length is more of a maximum size. I'd expect something like in the range of a few MBs and/or GBs.

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
RalphSteinhagen: done

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
```diff
@@ -0,0 +1,307 @@
...
+            std::filesystem::remove(entry.path());
+        }
+    }
+
+    return deletedFiles;
+}
+
+} // namespace detail
+
+enum class FileMode { override, append, multi };
```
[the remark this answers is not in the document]
RalphSteinhagen: done

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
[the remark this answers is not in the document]
RalphSteinhagen: N.B. nothing needs to be done here because the function already has a try-catch block in the calling function e.g.
https://github.com/fair-acc/gnuradio4/blob/f00cba2c81422f143c9d48552921ec63bbaef652/core/include/gnuradio-4.0/Block.hpp#L921-L935
and
https://github.com/fair-acc/gnuradio4/blob/f00cba2c81422f143c9d48552921ec63bbaef652/core/include/gnuradio-4.0/Block.hpp#L508-L524

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
RalphSteinhagen: changed it.

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
daniestevez: Sounds good.

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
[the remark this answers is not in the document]
daniestevez: My point was not about throwing exceptions. I believe that `value_or()` can't throw any exception in this case.

My point was that this piece of code is converting the settings provided string to a an enum value, or keeping the current `_mode` if the conversion results in an error (that's what `value_or()` is doing here).

Thus, it appears that the behaviour of this code is that when new settings are given, the `_mode` is changed to the one indicated by the new settings *if the new `mode` can be parsed* and otherwise the parse error is silently ignored and the old `_mode` is kept. This is what I find potentially surprising for the user. It would be better to error out (for instance throw an exception) if the `mode` cannot be parsed.

blocks/fileio/test/qa_FileIo.cpp
```diff
@@ -0,0 +1,151 @@
...
+void runTest(const gr::blocks::fileio::Mode mode, const std::shared_ptr<gr::thread_pool::BasicThreadPool>& threadPool) {
+    using namespace boost::ut;
+    using namespace gr::blocks::fileio;
+    using namespace gr::testing;
+    using scheduler = gr::scheduler::Simple<>;
+
+    constexpr gr::Size_t nSamples    = 1024U;
+    const gr::Size_t     maxFileSize = mode == gr::blocks::fileio::Mode::multi ? 256U : 0U;
+    std::string          modeName{magic_enum::enum_name(mode)};
+    std::string          fileName = fmt::format("/tmp/gr4_file_sink_test/TestFileName_{}.bin", modeName);
```
wirew0rm, line 43: Could use [std::temp_directory_path](https://en.cppreference.com/w/cpp/filesystem/temp_directory_path) for portability, but we probably have to do some changes once we try to extend to other platforms, anyway, so it can be done then.

wirew0rm (approved):
Nice implementation, that should nicely fit 90% of usecases and could be extended to fit most. It also nicely showcases the expected API for a bit more complex blocks.

Also thanks to @daniestevez for the feedback. Added small remarks to the open comments to document information from chat.

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
RalphSteinhagen: I disagree the `max_bytes_per_file` config parameter is simple and does what its name implies w/o surprises:

The goal of this parameter is to limit the upper bound size of the file in terms of what is relevant for the file system w/o doing or requiring mental math and how many samples this corresponds to by the user. 

The latter can be also quite difficult (and/or independent) depending on the type and `sizeof(T)` since `T` can be not just fixed-sized fundamental types, but more complex ones, and compound types (eventually `Packet<T>`, `Tensor<T>`, or `DataSet<T>`) that are even variably sized.

The use-case example is: "user wants to write data to file and the file not exceeding, for example, 1GB per file regardless whether it stores, `std::int16_t`, `float`, `std::complex<float>`, or a `DataSet<T>`...

This also makes the `max_bytes_per_file % sizeof(T) != 0` difficult to compute and annoying to the user, who usually does not care how big the individual samples/entities in the file are.

*N.B. `_totalBytesWritten` (not `totalBytesWritten` as mentioned) is declared as a private implementation-specific variable and not a public API. We could make it a runtime parameter (then: `total_bytes_written`), for example, for the UI and/or another monitoring process to track how many bytes are written. Together with `max_bytes_per_file` this could provide an indicator of whether progress is made on the block and/or when the file is complete.*

blocks/fileio/include/gnuradio-4.0/fileio/BasicFileIo.hpp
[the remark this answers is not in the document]
RalphSteinhagen: > For moderate amounts of samples, the global range could also be done with a separate head block and only integrated into BasicFileSource once this turns out to be a performance problem.

No you cannot replicate this with other blocks as post processing. 

The `BasicBlockSource` reads N-files where for each file an `offset` amount of files has to be skipped and `length` amount of samples. This information cannot be easily reconstructed once the information is put into the buffer  (N.B. 'bytes' vs 'samples').

The common real-world use case is: 
"for a burst or transient analysis: open all files, skip their (optional) header information, and show the first N samples."

Examples for transients:
GSI/FAIR: RF bunch oscillations at injection/start of the RF ramp. 
Radar domain: emitted or received pulses
Medical: MRI/CT or ultrasound pulses (conceptually similar to radar).

blocks/fileio/test/qa_FileIo.cpp
RalphSteinhagen: Thanks for the info. Didn't know about `std::temp_directory_path`.

PR #355, opened 2024-06-03
Fix clang-18 compilation error

drslebedev:
Fix clang-18 compilation error: use std::ranges::find instead of std::ranges::contains.


RalphSteinhagen (approved):
Thanks for the fixing/downgrading the expression.

The [std::ranges::contains](https://en.cppreference.com/w/cpp/algorithm/ranges/contains) seems to be a C++23 feature, and with the exception of `std::expected` and eventually `std::println(...)` we should minimise the use of C++23 features for now.

Can be merged as is.

PR #358, opened 2024-06-06
Perf

drslebedev:
This PR consists of 2 commits:
1) Add unit test checking `autoUpdate` and stored parameters in `CtxSettings`.

2) Add Performance monitoring block:
* Implement repeated tags for `TagSource`.
* Add infinite sample for `TagSource`.
* Add `PerformanceMonitor` block.
* Performance metrics can be save to cvs file.
* Add `qa_PerformanceMonitor` test.

Here are some performance results produced with `qa_PerformanceMonitor` for simple graph:
```text
                                                        ↱ tagSink
src -> monitorBulk -> monitorOne -> monitorPerformance -> tagSink
                                                        ↳ tagSink
```

The results include 200 seconds of running.
Case 1: no tags (in blue)
Case 2: 1 tag per 10000 samples (in orange)
Case 3: 1 tag per 1 samples (in green, right axis), evaluation rate was set to 1000 samples.

result


RalphSteinhagen (approved):
Well done and good to see that there isn't a memory growth/leak issue! :+1: 

... some minor improvement suggestions we discussed in person.

Below/above is a more verbose list of things we discussed. Once you incorporated and the CI passes, please squash-merge this at your convenience.

Please do not forget to add the history graph of the memory and throughput footprint.

Keep up the good work!

PR #359, opened 2024-06-10
forwarding member functions Annotated<...> & Soapy integration

RalphSteinhagen:
### forwarding member functions Annotated<...>

These changes improve the flexibility, usability, and robustness of the Annotated wrapper:
 - Added const and constexpr variants to forwarding member functions (operator(), operator[], operator->*).
 - Improved operator-> to ensure both non-const and const access.
 - Ensured implicit conversions to the underlying type T are correctly handled.
 - Made the wrapper more transparent and easier to use in various contexts.

... and added corresponding unit-tests.

### new block: Soapy Source

 - Implements a SoapySDR::Device-like via C-API wrapper as a workaround to ensure ABI compatibility for GR4 being compiled with libc++ and the Soapy wrapper with GCC and vice-versa.
 - Added corresponding unit-test and soapy-lime-based example for clarity.

Tested for the following device driver: audio, rtlsdr, lime

@xaratustrah, here's a new feature for you to play with. Have fun!


blocks/soapy/include/gnuradio-4.0/soapy/SoapyRaiiWrapper.hpp
```diff
@@ -0,0 +1,805 @@
...
+        friend class Device;
+
+    public:
+        using value_type                       = TValueType;
+        constexpr static std::size_t Direction = direction;
+
+        Stream() noexcept                = default;
+        Stream(const Stream&)            = delete;
+        Stream& operator=(const Stream&) = delete;
+        Stream(Stream&& other) noexcept : Stream() { swap(*this, other); }
```
ivan-cukic, line 514: While this is legal, as a user, it can have unexpected behavour:
- moved-from state is not empty
- moved from state contains old state of the thing we moved into

blocks/soapy/test/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+add_executable(qa_Soapy qa_Soapy.cpp)
+target_include_directories(qa_Soapy PRIVATE ${CMAKE_BINARY_DIR}/include ${CMAKE_CURRENT_BINARY_DIR})
+target_link_libraries(qa_Soapy PRIVATE gnuradio-options gnuradio-core fmt refl-cpp yaml-cpp::yaml-cpp fftw gr-soapy gr-testing ut)
+add_test(NAME qa_Soapy COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} ${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME})
```
drslebedev, line 4:
```suggestion
add_test(NAME qa_Soapy COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} ${CMAKE_CURRENT_BINARY_DIR}/qa_Soapy)
```
When running `All CTest` with `clion` ${TEST_NAME} is always empty and test fails.

blocks/soapy/include/gnuradio-4.0/soapy/SoapyRaiiWrapper.hpp
```diff
@@ -0,0 +1,805 @@
...
+            argInfo.optionNames.resize(cInfo.numOptions);
+            for (size_t j = 0; j < cInfo.numOptions; ++j) {
+                assert(cInfo.optionNames[j] != nullptr && "Option name string should not be null");
+                argInfo.optionNames[j] = std::string(cInfo.optionNames[j]);
+            }
+        }
+
+        SoapySDRArgInfo_clear(&cInfo);
+    }
+    SoapySDR_free(infos);
```
drslebedev, line 124:
```suggestion
    }
    SoapySDRArgInfoList_clear(infos, size);
```
There is a function to clear an array of `SoapySDRArgInfo`. Internally it does the same. However, address sanitizer still indicates some memory leaks.

blocks/soapy/test/qa_Soapy.cpp
```diff
@@ -0,0 +1,348 @@
...
+                    }
+                }
+                expect(ge(receivedSamples, 1000)) << fmt::format("did not received enough samples for deviceDriver = {}\n", deviceDriver);
+                rxStream.deactivate();
+            };
+
+            fmt::println("Basic API test - deviceDriver '{}' -- DONE", deviceDriver);
+        } //
+        | availableDeviceDriver;
+    //  |  std::array{"rtlsdr"s, "lime"s}; // alt: fixed tests
```
drslebedev, line 230: It might be better to use predefined devices. In my tests, all available devices included some remote and audio devices that failed the tests.

drslebedev (commented):
Thanks for the PR. I tested the code locally with `RTL-SDR` and `LimeSDR` devices. The `RTL-SDR` worked without any issues. However, some tests failed with the `LimeSDR`, which might be related to problems with my device. 

There is still a memory leak identified by the memory sanitizer. I haven't found a solution yet, but I will continue investigating it.

blocks/soapy/include/gnuradio-4.0/soapy/SoapyRaiiWrapper.hpp
```diff
@@ -0,0 +1,805 @@
...
+        argInfo.value       = cInfo.value ? std::string(cInfo.value) : "";
+        argInfo.name        = cInfo.name ? std::string(cInfo.name) : "";
+        argInfo.description = cInfo.description ? std::string(cInfo.description) : "";
+        argInfo.units       = cInfo.units ? std::string(cInfo.units) : "";
+        argInfo.type        = static_cast<ArgInfo::Type>(cInfo.type);
+        argInfo.range       = {cInfo.range.minimum, cInfo.range.maximum, cInfo.range.step};
+
+        if (cInfo.options != nullptr) {
+            argInfo.options.resize(cInfo.numOptions);
+            for (size_t j = 0; j < cInfo.numOptions; ++j) {
```
drslebedev, line 108:
```suggestion
            for (std::size_t j = 0; j < cInfo.numOptions; ++j) {
```

blocks/soapy/include/gnuradio-4.0/soapy/SoapyRaiiWrapper.hpp
```diff
@@ -0,0 +1,805 @@
...
+                assert(cInfo.options[j] != nullptr && "Option string should not be null");
+                argInfo.options[j] = std::string(cInfo.options[j]);
+            }
+        }
+
+        if (cInfo.optionNames != nullptr) {
+            argInfo.optionNames.resize(cInfo.numOptions);
+            for (size_t j = 0; j < cInfo.numOptions; ++j) {
```
drslebedev, line 116:
```suggestion
            for (std::size_t j = 0; j < cInfo.numOptions; ++j) {
```

blocks/soapy/include/gnuradio-4.0/soapy/SoapyRaiiWrapper.hpp
```diff
@@ -0,0 +1,805 @@
...
+        }
+    }
+
+    ~KwargsWrapper() {
+        for (size_t i = 0; i < _cArgs.size; ++i) {
+            delete[] _cArgs.keys[i];
+            delete[] _cArgs.vals[i];
+        }
+        delete[] _cArgs.keys;
+        delete[] _cArgs.vals;
```
drslebedev, line 57:
```suggestion
SoapySDRKwargs_clear(&_cArgs);
```
Can one use standard clear function from SoapySDR?

blocks/soapy/test/CMakeLists.txt
RalphSteinhagen: Done.

blocks/soapy/include/gnuradio-4.0/soapy/SoapyRaiiWrapper.hpp
RalphSteinhagen: done

blocks/soapy/include/gnuradio-4.0/soapy/SoapyRaiiWrapper.hpp
RalphSteinhagen: done

blocks/soapy/include/gnuradio-4.0/soapy/SoapyRaiiWrapper.hpp
RalphSteinhagen: done.

blocks/soapy/test/qa_Soapy.cpp
RalphSteinhagen: For me the test passed for 'rtlsdr', 'lime', and 'audio'. 

Some of the earlier failures I noticed where due to the drivers not implementing all the same capabilities (e.g. mutually exclusive sampling rates between audio and proper SDR devices, w/ vs. w/o gain setting capabilities). I'ts a pity that there is no default sw-only test-device that we could use in the CI ...

We should investigate more, if there are more/other devices that fail... thought that Soapy is a generic abstraction that would define either a lowest common denominator and/or provide dummy/empty implementations for those devices that cannot physically implement certain features.

If not, then this Soapy integration will be quite hard in terms of testing and guaranteeing that it works with any Soapy-wrapped SDR driver. 

Also the residual (static?) memory leak seems to be device/driver specific which seem to point to some leak in the soapy module implementation. For example, the same code shows more ASAN leaks for 'audio' than for 'rtlsdr' or 'lime'. :thinking: 

*N.B. need to find a solution, otherwise this will become a never-ending integration/testing issue....*

blocks/soapy/test/qa_Soapy.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+        //auto [watchdogThread, externalInterventionNeeded] = createWatchdog(sched, 6s);
+
+        auto retVal = sched.runAndWait();
+        expect(retVal.has_value()) << fmt::format("scheduler execution error: {}", retVal.error());
+
+        // if (watchdogThread.joinable()) {
+        //     watchdogThread.join();
+        // }
+        // expect(!externalInterventionNeeded->load(std::memory_order_relaxed));
```
drslebedev, line 341: Please uncomment watchdog and enable timeout.

drslebedev (approved):
Thank you for making the changes!

I tested the code again with audio, LimeSDR and RTLSDR devices and everything worked perfectly. However, it fails with the remote device. We should consider whether it's necessary to include it now, or we could filter it out for the moment.

Please uncomment the watchdog to enable the timeout.

I believe the PR is ready to be merged.

CMakeLists.txt
```diff
@@ -225,6 +221,28 @@ FetchContent_Declare(
 
 FetchContent_MakeAvailable(fmt pmt ut yaml-cpp vir-simd cpp-httplib)
 
+# Fetch SoapySDR -- needed since the distribution version is incompatible w.r.t. stdlibc++ vs. libc++
+if (CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)") # WIP
+    find_package(SoapySDR CONFIG)
+
+elseif (CMAKE_CXX_COMPILER_ID MATCHES "(Clang|GNU)")
```
zuckschwerdt, line 228: Will this ever match? It's the same condition as the `if`, isn't it?

RalphSteinhagen: Yes ... :see_no_evil: 

A left-over from debugging....

PR #360, opened 2024-06-11
CtxSettings type test, support std::complex for settings

drslebedev:
The PR consists of 2 commits:

Use std::hold_alternative instead of std::visit for PMTCompare + CtxSettings type test

Support of std::complex in Settings and YAML


RalphSteinhagen (approved):
This is a good extension! :+1: 

As a caution: we should limit further extensions using the yaml-cpp lib as this becomes a hard-to-maintain dependency especially because of the duplication of a `std::any<>` and `pmtv` based intermediate representation. Should be solved in the pmtv library... 

Maybe we can find some volunteers for this. @mormj @jsallay would you know about volunteers? :grinning:

PR #363, opened 2024-06-17
Add support to store UncertainValue samples for TagMonitor and TagSink 

drslebedev:
Add support to store UncertainValue samples for TagMonitor and TagSink by removing error part.
This is neede for some unit tests.


RalphSteinhagen (approved):
LGTM. :+1: 

as discussed.

PR #365, opened 2024-06-20
Fix optional gcc dependency on TBB

### LINKED ISSUE #361: Missing link to TBB with libstdc++ on Fedora
On Fedora 40 with GCC 14.1 and libstdc++ I get a couple of ```/usr/bin/ld: algorithm/test/CMakeFiles/qa_FilterTool.dir/qa_FilterTool.cpp.o: in function `void tbb::detail::(...)```. The following change made it compile for me with my system compiler:

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,6 +143,8 @@ endif ()
 # Mainly for FMT
 set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)

+find_package(TBB REQUIRED)
+
 add_library(gnuradio-options INTERFACE)
 include(cmake/CompilerWarnings.cmake)
 set_project_warnings(gnuradio-options)
--- a/algorithm/CMakeLists.txt
+++ b/algorithm/CMakeLists.txt
@@ -1,6 +1,6 @@
 add_library(gnuradio-algorithm INTERFACE)
 target_include_directories(gnuradio-algorithm INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/> $<INSTALL_INTERFACE:include/>)
-target_link_libraries(gnuradio-algorithm INTERFACE gnuradio-options gnuradio-meta vir fftw fmt magic_enum)
+target_link_libraries(gnuradio-algorithm INTERFACE gnuradio-options gnuradio-meta vir fftw fmt magic_enum TBB::tbb)

 if (ENABLE_TESTING)
     add_subdirectory(test)

```

However, I am not too familiar with your build system, so someone else should review this and revise a proper fix. See also:

> `-ltbb` - Linking to tbb (Thread Building Blocks) is required for use of the Parallel Standard Algorithms and execution policies in `<execution>`.

https://gcc.gnu.org/onlinedocs/libstdc++/manual/using.html#manual.intro.using.flags

wirew0rm:
For implementing std::execution::par, gcc relies on the TBB library. On systems without TBB installed it falls back to the non-parallel implementation, but as soon as the TBB header is available it will try to use it and require linking against TBB.
https://github.com/gcc-mirror/gcc/blob/releases/gcc-14.1.0/libstdc%2B%2B-v3/include/bits/c%2B%2Bconfig#L874-L882 Since TBB is a bigger dependency, making it a required dependency is not an option, so this introduces the option ENABLE_TBB which by default is off but can be used to add the dependency.

The alternative would be to add TBB as a non-required dependency and only add the linker flag if it is found, following the approach from gcc. Would be one less cmake option but I think having it explicit is better allow reproduction between different systems.

fixes #361

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,8 @@ set(CMAKE_CXX_STANDARD 23)
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
 
+add_library(gnuradio-options INTERFACE)
+
 if (CMAKE_CXX_COMPILER MATCHES "/em\\+\\+(-[a-zA-Z0-9.])?$") # if this hasn't been set before via e.g. emcmake
     message(" Transpiling to WASM: using: Emscripten (${CMAKE_CXX_COMPILER})")
     set(EMSCRIPTEN ON)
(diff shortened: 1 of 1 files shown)


RalphSteinhagen (approved):
LGTM! :+1:

PR #370, opened 2024-06-26
Rename numerator to output_chunk_size and denominator to input_chunk_size

drslebedev:
Naming is hard... 

In the Block class, we had members `denominator` and `numerator`, which can be confusing due to how they are used. These fields define the resampling ratio as `numerator/denominator`. 
The exact implementation takes `denominator` input samples and creates `numerator` output samples. Meaning that `n_input = k * denominator` and `n_output = k * numerator`. They also act as constraints for the minimum number of input/output samples.

In this PR  `numerator` is renamed to `output_chunk_size` and `denominator` to `input_chunk_size`  which should reflect the purpose of these variables clearly.

In addition documentation was also updated.


RalphSteinhagen (approved):
LGTM!

Thanks for the follow-up, hunting down the opinions and refactoring the code. :+1: 

*famous last words: I trust this renaming will stand the trial-of-time*. :smile:

PR #371, opened 2024-06-27
Soapy: added set vs. actual center frequency check

RalphSteinhagen:


blocks/soapy/include/gnuradio-4.0/soapy/Soapy.hpp
```diff
@@ -201,6 +213,14 @@ This block supports multiple output ports and was tested against the 'rtlsdr' an
         for (std::size_t i = 0UZ; i < nChannels; i++) {
             _device.setCenterFrequency(SOAPY_SDR_RX, rx_channels->at(i), rx_center_frequency->at(std::min(i, nFrequency - 1UZ)));
         }
+        std::vector<double> rx_center_frequency_actual;
+        for (std::size_t i = 0UZ; i < nChannels; i++) {
+            rx_center_frequency_actual.push_back(_device.getCenterFrequency(SOAPY_SDR_RX, rx_channels->at(i)));
+        }
+        auto equalWithinOnepercent = [](const std::vector<double>& a, const std::vector<double>& b) { return a.size() == b.size() && std::equal(a.begin(), a.end(), b.begin(), [](double x, double y) { return std::abs(x - y) <= 0.01 * std::max(std::abs(x), std::abs(y)); }); };
```
wirew0rm, line 220: lambda duplicates `detail::equalWithOnePercent()` helper function.

wirew0rm (approved):
LGTM

one small leftover duplication that I think was missed during cleanup.

Just to see if I understood it right and document a bit the underlying issue: If the user requests a setting value that is not supported by the actual hardware, it might be changed to an arbitrary value that the driver considers closest to the requested value. This is a problem since the further processing may depend on settings like samplerate and gain having the configured value.
With this change, requesting a value that the hardware cannot accommodate is an exception (stopping flowgraph execution?).
I guess this fail-early paradigm is less error-prone than trying to reconcile the requested and actual settings.

Q: recovering is possible by setting a supported setting and restarting the graph processing?

blocks/soapy/include/gnuradio-4.0/soapy/Soapy.hpp
RalphSteinhagen: in.

PR #375, opened 2024-07-09
suppressing deprecation warnings for  atomic_load_explicit and atomic_compare_exchange_weak methods

drslebedev:
Currently, suppressing deprecation warnings for `std::atomic_load_explicit` and `std::atomic_compare_exchange_weak` methods. Note: While `std::atomic<std::shared_ptr<std::vector<std::shared_ptr<Sequence>>>>` is compatible with GCC, it is not yet supported by libc++.
This workaround is necessary to maintain compatibility and avoid deprecation warnings in GCC. For more details, see the following example implementation: https://godbolt.org/z/xxWbs659o

Revisit this code once libc++ adds support for `std::atomic<std::shared_ptr<std::vector<std::shared_ptr<Sequence>>>>`.


RalphSteinhagen (approved):
Thanks for the changes. Can be merged once the CI passes.

PR #379, opened 2024-07-12
Doc: add development setup documentation

wirew0rm:
I've written a bit of documentation on how to get started compiling gr4, would be good if some people (ideally ones who don't work with the project on a daily basis) could provide some feedback if things are missing or are unclear.


DEVELOPMENT.md
```diff
@@ -0,0 +1,53 @@
...
+Note that while the binaries inside of `./build` can be accessed on the host system, they are linked against the libraries of the container and will most probably not run on the host system.
+
+```bash
+me@host$ cd gnuradio4
+me@host$ docker run \
+    --user `id -u`:`id -g` \
+    --volume="/etc/group:/etc/group:ro" \
+    --volume="/etc/passwd:/etc/passwd:ro" \
+    --volume="/etc/shadow:/etc/shadow:ro" \
+    --workdir=/build --volume `pwd`:/build  -it \
```
marcusmueller, line 23: maybe rather name the source directory `/src` instead of `/build`, to avoid confusion

marcusmueller (approved):

DEVELOPMENT.md
```diff
@@ -0,0 +1,53 @@
...
+
+To apply the project's formatting rules, you'll also need the correct formatters, `clang-format-18` and `cmake-format`. With these installed you can use the scripts in the repository to reformat your changes. Otherwise 
+
+Once these are installed, you should be able to just compile and run GNURadio4:
+
+```bash
+me@host$ cd gnuradio4
+me@host$ cmake -S . -B build
+me@host$ cmake --build build
+me@host$ cd build && ctest .
```
dennisklein, line 52:
```suggestion
me@host$ ctest --test-dir build
```

since CMake 3.20: https://cmake.org/cmake/help/latest/manual/ctest.1.html#cmdoption-ctest-test-dir

wirew0rm: I'll have to check if this also sets the working directory, iirc there were some tests that depended on that (probably the path to the plugins to dl-open).

DEVELOPMENT.md
wirew0rm: `/work/src`

PR #381, opened 2024-07-18
ClaimStrategy: fix bulk update of AtomicBitset

### LINKED ISSUE #380: Investigate Regression in MultiProducerStdMapMultipleWriters qa_buffer test
It seems the recent changes to the circular buffer introduced a subtle bug in the multiple writers case. It was observed as sometimes failing the CI on multiple subsequent PRs (even ones without any code-changes).
It can be reliably reproduced locally by increasing the number of writers above the number of cores:
``` diff
diff --git a/core/test/qa_buffer.cpp b/core/test/qa_buffer.cpp
index 70a724d..26a89f6 100644
--- a/core/test/qa_buffer.cpp
+++ b/core/test/qa_buffer.cpp
@@ -515,8 +515,8 @@ const boost::ut::suite CircularBufferTests = [] {
 
     "MultiProducerStdMapMultipleWriters"_test = [] {
         // now actually use multiple writers, and ensure we see all expected values, in a valid order.
-        constexpr auto kNWriters = 5UZ;
-        constexpr auto kWrites   = 20000UZ;
+        constexpr auto kNWriters = 50UZ;
+        constexpr auto kWrites   = 2000UZ;
 
         gr::CircularBuffer<std::map<int, int>, std::dynamic_extent, gr::ProducerType::Multi> buffer(1024);
         using WriterType                  = decltype(buffer.new_writer());
```
leading to the following repeated error message and then eventually segfaulting: 
```
in: /tmp/graph-prototype/core/test/qa_buffer.cpp:552 - test condition:  [false]FAILED
```

wirew0rm:
This fixes the case where setSlotStates is called with (n, n + size). Before that would lead to no Slots being published, now it should correctly update the buffer.
This bug can only be reproduced on low-core machines ore by increasing the number of writers in qa_buffer MultiProducerStdMapMultipleWriters.

fixes #380

Since I wasn't that deep into the code over there, I mainly concentrated on the diff of the commit introducing the failures. I could trace the problem to `ClaimStategy::setSlotsStates(...)` where uncommenting the non-bulk update of the AtomicBitset solved the issue. Invoking the non-bulk API one level deeper still caused the problem, so the problem had to be in that function itself.
The change now should make the bulk API behave the same way as the non-bulk one. Since this is currently breaking the build for all PRs I'd like to merge this rather soon, but we should later revisit if this is the actually correct solution or just hiding the problem again.

### DIFF
--- a/core/include/gnuradio-4.0/ClaimStrategy.hpp
+++ b/core/include/gnuradio-4.0/ClaimStrategy.hpp
@@ -37,7 +37,7 @@ public:
     TWaitStrategy                                           _waitStrategy;
     std::shared_ptr<std::vector<std::shared_ptr<Sequence>>> _readSequences{std::make_shared<std::vector<std::shared_ptr<Sequence>>>()}; // list of dependent reader sequences
 
-    explicit SingleProducerStrategy(const std::size_t bufferSize = SIZE) : _size(bufferSize){};
+    explicit SingleProducerStrategy(const std::size_t bufferSize = SIZE) : _size(bufferSize) {};
     SingleProducerStrategy(const SingleProducerStrategy&)  = delete;
     SingleProducerStrategy(const SingleProducerStrategy&&) = delete;
     void operator=(const SingleProducerStrategy&)          = delete;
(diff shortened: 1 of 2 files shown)


drslebedev (approved):
Thanks a lot for the PR!

For me the changes look good!

ivan-cukic (approved):

PR #382, opened 2024-07-19
Update CONTRIBUTING.md to broaden the intent of the contributor agreement

RalphSteinhagen:
- Merged DCO Signing and Contributor License Agreement sections for clarity.
- Clarified the authority required for code submissions and the potential rights of employers.
- Simplified the DCO signing instructions and emphasized the need for legal compliance.
- Streamlined language to reduce redundancy and maintain legal soundness.
- Added explicit confirmation of rights granted by contributors, including employer permissions.
- Emphasized GNU Radio’s principles of remaining free/libre and compliant with evolving laws.
- Highlighted the support for public-private partnerships and commercial contributions without disclosing proprietary stacks.
- Maintained the spirit of open-source contributions and community collaboration.
- Ensured the guidelines are clear, concise, and foster an inclusive environment.

See the updated CONTRIBUTING.md for details.

Signed-off-by: Josh Morman <jmorman@gnuradio.org>
Signed-off-by: Ralph J. Steinhagen <r.steinhagen@gsi.de>


flynn378:
Should this for the Linux kernel standard (and many other projects), a SPDX entry on each file with copyright per file.  It removes the ambiguity and clearly labels each file.  https://reuse.software/ is a tool that helps with this process.

Based on conversions with the lawyers at one entity which supports and contributes to open source projects, assigning a copyright when a license term can be changed so that they cannot use it in the future will prevent submitting code.   It stopped submissions to GNU Radio when copyright had to be given to the FSF.

I could be misreading the language, but it looks like this includes an agreement the license could be changed, and in essence a copyright is being assigned, which allows this.

daniestevez:
The paragraph "Notably by contributing to GNU Radio" to me looks very much like a CLA that we don't have in GNU Radio 3. "You grant this project" is hardly legalspeak, because it isn't clear what legal entity is "the project", especially since at the moment GNU Radio is an unincorporated association which perhaps cannot even hold the rights that are said to be granted. In the way that this is written, it appears to me that the project would get the rights to re-license the contributed code in any way (including under a closed-source license), despite the motivations given below saying that the intention is that GNU Radio shall ever remain open source (these intentions are not legally binding). I don't think the project would ever abuse this position, but given the number of open source projects that have suddenly become closed, I think this is a valid concern to express.

Why not exactly the same licensing scheme we have in GNU Radio 3? (This is: a DCO, the contribution must be under GPLv3 or later or compatible, and the contributor remains the only copyright holder).

Moreover, probably having a CLA would require some form of signed document by contributors, as GNU Radio 3 used to do in the FSF days, instead of just a simple `Signed-off-by` in git commits. This requirement turned out to be problematic for several contributors.

As a more practical example of why this kind of CLA is bad, I think I cannot take some GR 3 that I haven't written (or any other open source code, for that matter), adapt it to GNU Radio 4, and contribute it. This is because my contribution is a derived work and I can't grant rights for the original work.

mormj:
Hope this isn't overkill, but added more detail about copyright assignment so it is clear that we are not requiring or dealing with CLAs.  Also another sentence on irrevocability as this was one of the concerns with not having copyright assignment

CONTRIBUTING.md
```diff
@@ -37,7 +37,33 @@ Please be sure to follow the usual process for submitting PRs:
...
+
+GNU Radio does not claim ownership of any contributions you make. All copyrights remain with the original author of the contribution. As the author, you are responsible for maintaining and marking your copyright in the appropriate locations.
+
+When modifying or adding new files, you must include a copyright header in each file you touch. The copyright header should contain the following information:
+
+```
+/*
+ * Copyright (C) [Year] [Name or Pseudonym of Author]
+ *
+ * SPDX-License-Identifier: [License of module or file, default LGPL-3.0 for core]
```
daniestevez, line 50: Since 3882ef71ac70f70a71aec455c6bd8b57e137b448, the SPDX for the core is `LGPL-3.0-linking-exception` rather than `LGPL-3.0`.

I'm just marking this as an item to review and keep in sync with the core `LICENSE` file before merging.

RalphSteinhagen: We can safely revert this to the vanilla LGPL-3.0 w/o additional disclaimer. We got meanwhile the info that this has been explicitly [confirmed](http://listengine.tuxfamily.org/lists.tuxfamily.org/eigen/2009/01/msg00083.html) for another header-only C++ library (Eigen) by Brett Smith (Licensing Compliance Engineer, [Free Software Foundation](http://fsf.org/)).

For more practical details, see [here](https://www.eigen.tuxfamily.org/index.php?title=Licensing_FAQ&oldid=1117#So_what_does_the_LGPL_require_me_to_do.3F).

I hope this concludes this legal semantic debate.

daniestevez: Is the link to the eigen mailing list correct? For me it goes to a thread "[eigen] problem to contact svn repository".

I'm not sure that eigen is a good example to advocate for LGPL-3.0 w/o linking exception for a header-only library. In 2012 [they changed their license to MPL 2.0](http://eigen.tuxfamily.org/index.php?title=ChangeLog#Eigen_3.1.1). In [their own words](http://eigen.tuxfamily.org/index.php?title=Licensing_FAQ), when eigen was LGPL they had a long FAQ about the implications that it isn't needed anymore with the MPL. It would be good to know more about the background and reasons for the license change.

The old [LGPL FAQ](http://eigen.tuxfamily.org/index.php?title=Licensing_FAQ&oldid=1117#So_what_does_the_LGPL_require_me_to_do.3F) mentions in two places that they asked the FSF about what happens with a header-only library, but the links they give to the mailing list are wrong (or most likely have become broken): http://listengine.tuxfamily.org/lists.tuxfamily.org/eigen/2009/01/msg00083.html and http://listengine.tuxfamily.org/lists.tuxfamily.org/eigen/2008/02/msg00008.html (one of these is the link you mentioned, the other points to a thread that does speak about FSF and licensing, but there are no clear answers there).

I think that in practice it doesn't matter if someone from the FSF in 2008 or 2009 said something about header-only libraries under the LGPL-3.0. What matters if we want to promote/simplify using GNU Radio 4.0 in closed source applications in industry is whether a legal department can review the license terms and have high confidence that there is no "GPL-like viral contamination" of the company's closed source code. If there is any reasonable doubt, they'll err on the side of caution and treat the license as a GPL-like viral license.

RalphSteinhagen: @daniestevez corrected the link. We could also opt for a more liberal license, but the decision of whether LGPLv3, or more liberal one is not mine to make and should be moderated/coordinated by the GR leadership team.

The repo is technical, and we should find a different platform to discuss the nuances of GNU Radio Project leadership and roadmap issues.

CONTRIBUTING.md
```diff
@@ -45,37 +37,73 @@ Please be sure to follow the usual process for submitting PRs:
...
```
RalphSteinhagen, line 50: On a similar note as above:

In my view, it's important to keep licensing and (c)lean, maintainable code as separate concerns. Including a mandatory copyright header in every file doesn’t necessarily add value but does increase maintenance overhead and refactoring complexity.

Moreover, such headers can create unnecessary ambiguity, leading to questions about whether changes or add-ons conflict with top-level licensing agreements and a lot of bikeshedding. This ambiguity can require further reviews from legal experts, increasing the burden on contributors and maintainers.

To simplify this process, we already have a [contributor list](https://github.com/fair-acc/gnuradio4#user-content-contributors), which acknowledges contributions equally, regardless of the size or nature of the improvement. In line with the principle of "Keep It Simple, Stupid" (KISS), I believe we avoid legalese in the source code and focus on the core utility and quality of the library itself.

CONTRIBUTING.md
daniestevez: Thanks for finding the correct link. Still Brett Smith's response looks unconvincing to me. The question he is answering to is "I would like to know if the LGPLv3 can be suitable for a C++ pure template library, where all the code is in headers?", which is vague since "suitable" can mean any number of things. Brett simply says "yes, it is suitable, because of LGPLv3 Section 3".

The question that needs an answer is "can a closed-source application use an LGPLv3 header-only library, and if so, which requirements need to be followed". I know about section 3 in the LGPLv3, but it says "Give prominent notice with each copy of the object code that the Library is used in it and that the Library and its use are covered by this License". My reading is that Section 4 d) is still applicable, so users need to have a means to modify the Library in the Combined Work (and that's impossible if the Combined Work is a closed-source application and the Library is header-only).

> The repo is technical, and we should find a different platform to discuss the nuances of GNU Radio Project leadership and roadmap issues.

Happy to take the discussion somewhere else. I think the reason we're discussing in this PR is that this got brought up in the Architecture Matrix channel and between discussing it in the channel or discussing it in the PR, it was decided that it was better to discuss it in the PR.

CONTRIBUTING.md
daniestevez: I get your point, but GNU Radio 3.10 does this and it doesn't seem to generate much burden.

mormj:
Suggesting based on the following article for best practices of modern projects:

https://www.linuxfoundation.org/blog/blog/copyright-notices-in-open-source-software-projects

1) Remove explicit copyright assignments from the headers - or at least don't require them because they are hard to maintain
2) Remove the additional license terms beyond the DCO

RalphSteinhagen:
@mormj you know that you are allowed to push to that branch (and also already did before). 😉 

I like the simplification and also dropping of the 'SPDX-License-Identifier' in every file which is (legally) complicated and doesn't add functional (w.r.t. signal-processing) value to the code. The LF itself states:
> Most LF project communities do not require or recommend that every contributor include their copyright notice in contributed files. See below for more details on why not.

I think having a top-level `LICENSE` file is enough, add a separate file for sub-modules that differs (or legally requires it), and otherwise follow the suggestions as outlined in [1]:

> [1] https://www.linuxfoundation.org/blog/blog/copyright-notices-in-open-source-software-projects

May I suggest the following additional changes:
  1. Copyright assignment note in the README.md:
  ```markdown
  Copyright (C) 2001-September 2020 GNU Radio Project -- managed by Free Software Foundation, Inc.
  Copyright (C) September 2020-2024 GNU Radio Project -- managed by SETI Institute
  Copyright (C) 2018-2024 FAIR -- Facility for Antiproton & Ion Research, Darmstadt, Germany
  ```
  to simply
  ```markdown
  Copyright The GNU Radio Authors
  Copyright Contributors to the GNU Radio Project
  Copyright FAIR - Facility for Antiproton & Ion Research, Darmstadt, Germany
  ```
  Subtlety: the first are people paid by the GNU Radio project (copyright transfer) and the second for anybody else.
  
  2. Actually, exchange the [LICENSE](https://github.com/fair-acc/gnuradio4/blob/main/LICENSE) with the vanilla [MIT](https://opensource.org/license/mit) and the above copyright holder lines. The years are not important and can be derived from the git track record.

  3. clearly state that no legal decorations (i.e. SPDX identifier) should be added to new source files unless mandated by reasons outlined in [1]. Also, we should add [1] as a reference for posterori.

  4. Add a small section on why 
    * the GR4 core is 'MIT' licensed (and why), while 
    * GR4 block-libs are MIT-preferred and should include a different `LICENSE` file if (and only if) the contributed (usually copied/already licensed) code requires it. 
  If MIT can't be achieved, then I propose that the default for the GR4 block-lib sub-module be 'LGPLv3' which includes nearly all other variants but 'GPLv3' to minimise the license zoo. 
  Of course, OOT tree modules license choices remain free, but we should recommend either MIT (if this is to become part of the GR4 core) or (L)GPLv3. The latter recommendation helps maintaining the 'GNU' spirit, while minimises friction with various legal departments and cross-checks, etc. 
  We aim to keep the project efficient, low-maintenance, and technically relevant.
  
I believe such an explanation would go a long way with the community, and the statement should be drafted by one of the **GNU Radio leadership** (@mormj) and not FAIR to ensure that FAIR remains a contributor among many and to keep an open and fair playing field.

daniestevez:
The changes suggested by @mormj look good to me. It might be good to link the LF blog post also in the README (maybe in the motivations section), saying that it is good background information (but not legally binding).

Regarding @RalphSteinhagen's comments, some things to point out (my own numbering, not correlated to Ralph's numbering):

1. Generally we cannot change
```
Copyright (C) 2001-September 2020 GNU Radio Project -- managed by Free Software Foundation, Inc.
```
into
```
Copyright The GNU Radio Authors
```
The Linux Foundation blog post makes it clear that existing copyright statements cannot be changed without written consent of the copyright holder. In the case of GNU Radio code prior to September 2020, the copyright owner was the FSF, since there was a CLA that assigned copyright to the FSF.

However, probably none or very little code in the `gnuradio4` repository has been copied over from GNU Radio 3 code that is owned by the FSF. I know that in my `gr4-packet-modem` OOT I copied/adapted code from some GNU Radio 3 blocks for practical reasons, and clearly identified it by maintaining the copyright statements in the affected files. But I cannot think of any file in the `gnuradio4` repo that is affected by FSF copyright. So maybe the statement
```
Copyright (C) 2001-September 2020 GNU Radio Project -- managed by Free Software Foundation, Inc.
```
can be removed because it is not applicable, simplifying this discussion.

Also, the original FSF copyright statement from legacy GNU Radio code is something like
```
Copyright 2004,2009,2010,2013 Free Software Foundation, Inc.
```
if this is applicable in some part of the code base, we cannot change it to read something like `GNU Radio Project -- managed by Free Software Foundation, Inc`.

2.

>Subtlety: the first are people paid by the GNU Radio project (copyright transfer) and the second for anybody else.

It would be good to have clarity on who owns copyright in the case of paid work. Don't quote me on this, because I am not a lawyer, but my understanding is that the GNU Radio project itself is unincorporated. So I don't know whether it can own copyright.

For pre September 2020, all code submitted to GNU Radio had its copyright assigned to the FSF, as far as I know. Post September 2020, work paid by GNU Radio through SETI Institute probably has copyright assigned to SETI Institute. At least that was the case when I worked in `gr4-packet-modem`. It was clearly written in the contract I signed, and I would expect similar contributors to have signed the same contract template unless they negotiated other terms.

For work paid by FAIR I think we are all clear that the copyright is owned by FAIR and there is nothing to discuss there.

3. I think it is fine that parts of the project have different licenses. In particular an attractive that I see is that by allowing GPLv3 blocks in-tree, we ease the transition from GNU Radio 3 by allowing GNU Radio 3 blocks (which are GPLv3) to be ported over to GR4 (which would preserve the GPLv3, as the ported block would be derived work) rather than having to be re-implemented from scratch in a way that makes it clear that the GR4 block is not derived from the GR3 block.

However, I think that the repo should be organized in such a way that makes clear which parts of the code base have which licenses. SPDX headers on files are a means to do this, but probably not the best means, since they would require someone to check all the files to make sure. Maybe we can target something like different in-tree modules (such as the gr-analog, gr-audio, etc. of GR3) that contain code with different licenses (probably the zoo can be limited to MIT, LGPLv3 and GPLv3). Building these modules should be optional via CMake. The main use case I have in mind is the following: Let's say I'm in a corporate setting in which I can only take in permissibly licensed code (MIT, BSD, etc., but not LGPLv3 or GPLv3), for whatever legal reasons. I want to use gnuradio4, but I want to exclude those parts of the code that have unacceptable copyright for me. If I was able to do that via `-DENABLE_GPLV4_BLOCKS=False -DENABLE_LGPLV4_BLOCKS=False` or something like that in `cmake`, that would be great.

RalphSteinhagen:
updated copyright statement and LICENSE file to MIT

All work until now has, to the best of our knowledge, been exclusively done by or assigned (via CLA) to GSI/FAIR. 

Also, on behalf of:
  * GSI Helmholtzzentrum für Schwerionenforschung, Darmstadt, Germany, www.gsi.de
  * FAIR – Facility for Antiproton and Ion Research in Europe, Darmstadt, Germany, fair-center.eu/

in alignment with their support for Public Money, Public Code and their open-source software guidelines:
 * https://www.gsi.de/work/forschung/open-science
 * https://publiccode.eu/en/#organisations
 * [GSI/FAIR press release](https://www.gsi.de/en/start/news/details?cHash=0b34d60edad7e34f975bcc01b785ca12&tx_news_pi1%5Baction%5D=detail&tx_news_pi1%5Bcontroller%5D=News&tx_news_pi1%5Bnews%5D=5627)
 
 — This contribution is made freely, in the spirit of scientific openness and technological progress, for the benefit of all: public, academic, industrial, and governmental communities alike. Knowledge grows when shared.

CONTRIBUTING.md
```diff
@@ -1,81 +1,97 @@
...
 For general "how-to" and guidance questions about using GNU Radio to build and run applications, please have a look at the various
 examples or if you cannot find anything that fits your use-case use GitHub's discussions forum.
 
 ## Contributing code and content
+
 We welcome all forms of contributions from the community. Please read the following guidelines to maximise the chances of your PR being merged.
 
+### Communication
+
+- Before starting work on a feature, check if there isn't already an examples in the 'samples' sub-module.
```
ivan-cukic: s/an examples/an example/

Where is the 'samples' module located?

CONTRIBUTING.md
```diff
@@ -1,81 +1,97 @@
...
+
+- For any work related to setting up build, test, and CI for GNU Radio on GitHub, or for small patches or bug fixes, please open an issue
+  for tracking purposes, but we generally don't need a discussion prior to opening a PR.
+
+### Development process
+
+Please be sure to follow the usual process for submitting PRs:
+
+- Fork the repository
+- Make sure it compiles w/o errors against the current release 'main' branch:
```
ivan-cukic, line 36: Strangely formulated -- the "against main" part. Is the meaning of this to rebase on main?

mormj: maybe:

Make sure it compiles w/o errors when rebasing against the target merge branch of the PR?

CONTRIBUTING.md
```diff
@@ -1,81 +1,97 @@
...
+
+#### Non-Revocability of Contributions
+
+By submitting a contribution to the GNU Radio project, you agree that your contribution is non-revocable. Once a contribution is accepted and merged into the GNU Radio repository, it cannot be withdrawn or removed by the original author. This ensures that the integrity and continuity of the project's codebase are preserved.
+
+### License Terms
+
+GNU Radio intends to maintain the existing license terms under which contributions are made. We do not intend to change the licensing terms of any contributions after submission. Any potential changes to the license, such as re-licensing, would require an explicit, agreed-upon process involving the contributor and the project maintainers. Initial license terms are specified in the LICENSE file of the submodule, or in the SPDX header of the individual file if not consistent with the overall submodule.
```
wirew0rm: I think the last sentence would fit better at the end of the section after the explanation that there might be different licenses in the repository.

Also "submodule" could maybe be replaced by "block library"? We are not using "submodule" anywhere in code or documentation until now. If this is a deliberate choice, what is the difference between a submodule and a block library? In general I'm fine with anything, we should just try to be consistent.

Regarding file-level license headers, my assumption would be that the file level headers have to be compatible with the blocklib's LICENSE file
  - example: 
    - `spdx: MIT` in `blocks/myGPLBlocklib/file.hpp` -> ok, allows to move code from that file into MIT parts of the codebase
    - `spdx: GPL` in `blocks/myMITBlocklib/file.hpp` -> not ok, 
  - If this is the case I think it would be useful to make this a bit more explicit (but we could also add that once we add the first non MIT blocklib).
  - If this assumption is not true, I think it will make it very hard to ensure an MIT-only build (either manually or by supplying a `-DMIT_ONLY` cmake flag).

README.md
```diff
@@ -106,13 +110,13 @@ describe how to set up a local development environment.
 
 ## License and Copyright
 
-Unless otherwise noted: SPDX-License-Identifier: LGPL-3.0-linking-exception
-All code contributions to GNU Radio will be integrated into a library under the LGPL, ensuring it remains free/libre (FLOSS) for both personal and commercial use, without further constraints on either.
+Unless otherwise noted: SPDX-License-Identifier: MIT<br>
+All code contributions to GNU Radio will be integrated into a library under the MIT, ensuring it remains free/libre (FLOSS) for both personal and commercial use, without further constraints on either.
```
wirew0rm: This states "All code", but in the contribution guidelines say that some parts may be GPLv3. This should be mentioned here, too. It's good to have that information in the contribution guidelines, but they target a very different audience and I wouldn't expect most users and packagers to read them.

At the moment this is more theoretical as we don't have any GPLv3 parts, so we could also merge it as is and add this once we add the first GPL blocklib.

wirew0rm (approved):
Looks good. I have some small things i stumbled upon while reading. Since these are probably more subjective, I would leave it to you whether you want to follow them up or merge it as is. The general message works for me, and I think having this clear distinction with MIT/GPLv3 should have a lot less potential for licensing complications than the status quo.
Thanks at everyone who contributed to figuring this all out, I know this is a topic than can easily get swamped in endless discussions, but at the same time is important to get it at least mostly right at the start.

CONTRIBUTING.md
mormj: @wirew0rm  - thank you for the review and suggestions.  I moved the last sentence there to the end of the other section, and updated the verbage around submodule.  Most of the spdx file level header details have been omitted since I think there was still ongoing discussion and it doesn't need to hold up the overall license change.

README.md
mormj: Good catch here - I had forgotten that these changes were needed in the top level README

PR #392, opened 2024-08-15
Updates in Settings and FunctionGenerator

drslebedev:
This PR includes 4 commits.

### Commit 1
Settings refactoring: use CtxSettings, remove BasicSettings

### Commit 2
Updates in `CtxSettings` including:
* Prepare `CtxSettings` for use with the `FunctionGenerator`
* Add an additional test case related to `CtxSettings`.
* Note that parameters modified through the `Tag` are not persistent across context switches.
* Always All parameters are stored in storedParameters.
* Add `activateContext() ` to activate `SettingsCtx`
* Introduce a `setStaged()` method to allow parameter adjustments via messages. These staged parameters are not persistent across context switches.
* Introduce expiry time for stored parameters.
* removeExpiredParameters are executed as a last step of Settings::set() call.

### Commit 3
Change `FunctionGenerator` to use `CtxSettings` to switch between parameter sets.

### Commit 4
Create a separate `ThreadPool` for each test in `qa_Scheduler` to avoid data races.
* This is a temporary workaround, and the underlying issue will be addressed in future commits.

### Commit 5
Fix PythonBlock compilation error. 
qa_PythonBlock fails and will be updated in the next PR.

### Note
All commits are already in the correct order, so there's no need to squash them.


wirew0rm (approved):
Thanks, looks good now.

We'll have to follow up the Python block which is not tested in the CI and that has some regression that the `script` setting is not correctly applied and so no python code is actually executed.

PR #394, opened 2024-08-23
Use requestedWork as a soft limit in the work function

drslebedev:
Please first review https://github.com/fair-acc/gnuradio4/pull/392

Use `requestedWork` as a soft limit in the work function.

`work()` function should aim to honor the `requestedWork` value whenever feasible.

### DIFF
--- a/core/include/gnuradio-4.0/Block.hpp
+++ b/core/include/gnuradio-4.0/Block.hpp
@@ -1235,7 +1235,10 @@ protected:
         return 0UZ;
     }
 
-    auto computeResampling(std::size_t minSyncIn, std::size_t maxSyncIn, std::size_t minSyncOut, std::size_t maxSyncOut) {
+    auto computeResampling(std::size_t minSyncIn, std::size_t maxSyncIn, std::size_t minSyncOut, std::size_t maxSyncOut, std::size_t requestedWork = std::numeric_limits<std::size_t>::max()) {
+        if (requestedWork == 0UZ) {
+            requestedWork = std::numeric_limits<std::size_t>::max();
+        }
         struct ResamplingResult {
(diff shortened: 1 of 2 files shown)


core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1513,10 +1513,21 @@ class Block : public lifecycle::StateMachine<Derived>, public std::tuple<Argumen
...
+
         if (asyncEoS || (resampledIn == 0 && resampledOut == 0 && !hasAsyncIn && !hasAsyncOut)) {
-            return {requested_work, 0UZ, resampledStatus};
+            return {requestedWork, 0UZ, resampledStatus};
+        }
+
+        // Apply soft cut on requestedWork
+        if (availableToProcess > requestedWork) {
+            // include requestedWork in computeResampling calculation
+            const auto [resampledIn2, resampledOut2, resampledStatus2] = computeResampling(std::min(minSyncIn, nextEosTag), std::min(availableToProcess, requestedWork), minSyncOut, availableToPublish);
```
RalphSteinhagen, line 1526: The clamping looks OK but doesn't handle the corner case where requestedWork is < minSyncIn. The requestedWork clamping should not suppress the block to process data if it would have been able to if the requested work would have been max.

RalphSteinhagen: Otherwise OK.

RalphSteinhagen:
Some of the CI targets do not pass because of flaky unit-tests or CI resource limitations unrelated to this PR. Will merge this neverthless. The flaky tests would need to be addressed in separate PRs.

RalphSteinhagen (approved):

PR #399, opened 2024-09-03
algorithm: fix gcc warnings in fftw wrapper

daniestevez:
gcc is giving -Wsubobject-linkage warnings when gr::algorithm::FFTw is used as a member of a block because it says that fields such as fftwIn have internal linkage. This is caused by a lambda in the deleter of the std::unique_ptr used in InUniquePtr, etc. The warning can be removed by using a deleter struct instead of a lambda.

### DIFF
--- a/algorithm/include/gnuradio-4.0/algorithm/fourier/fftw.hpp
+++ b/algorithm/include/gnuradio-4.0/algorithm/fourier/fftw.hpp
@@ -9,6 +9,62 @@
 
 namespace gr::algorithm {
 
+namespace detail {
+    template<typename TData>
+    struct FFTwImplTypes {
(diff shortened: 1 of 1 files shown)


daniestevez:
The CI was still failing. It turns out that I messed up the syntax for the template specialization in the `double` case. I've just fixed and force-pushed. CI needs to be allowed to run.

daniestevez:
CI looks good. The pipelines that failed are:

- gcc-14 Debug. Failed on a seemingly unrelated test `qa_Messages`.
- gcc-14 Release. Didn't finish building in 6 hours: "The operation was cancelled"
- gcc-13 Debug. It looks that this didn't even ran.

PR #400, opened 2024-09-04
Fix qa_PythonBlock failed tests.

drslebedev:
All tests for `qa_PythonBlock` are now passing again.

The behavior of `Settings` changed after the introduction of `CtxSettings`. Unlike before, `CtxSettings` are not initialized in the `Block` constructor. Instead, a separate `Block::init(...)` function must be called. 
This change was necessary we need to determine which parameters are provided by the user and which should be set to default values, which is not possible in the Block constructor.

The `Block::init()` function internally calls `CtxSettings::init()` and `CtxSettings::applyStagedParameters()` to handle this initialization process.

### DIFF
--- a/blocks/basic/test/qa_PythonBlock.cpp
+++ b/blocks/basic/test/qa_PythonBlock.cpp
@@ -72,7 +72,7 @@ def process_bulk(ins, outs):
 )";
 
         PythonBlock<std::int32_t> myBlock({{"n_inputs", 3U}, {"n_outputs", 3U}, {"pythonScript", pythonScript}});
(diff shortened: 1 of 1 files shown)


wirew0rm (approved):
Thanks for the fix and documenting the changed behavior, looks good and can be merged once the CI has passed :+1:

wirew0rm:
I'll merge this even though there are too many failing tests, but they were introduced previously, are very likely to be unrelated (only touches a single test that is not even covered by the CI) and fixes a regression introduced by a previous commit.

PR #402, opened 2024-09-04
refactored & simplified Settings.hpp and pmtv format

RalphSteinhagen:
## Details:

* code and templating structure (via and in response to -ftime-trace analysis)
* replaced std::ranges::find(..) by more compile-time friendly version
* replaced std::sort by binary-search+merge-sort
* added non-inline to methods to avoid unnecessary compile-time optimisations
* optimised custom pmtv formatted by reducing the `std::visit(...)` pattern (not yet fully eliminated.

Compile-time differences (~4% ... hoped to be more, but it's a start).

**before:**
```text
real    20m21.229s
user    89m12.953s
sys     3m8.205s
```

**this PR:**
```text
real    19m33.855s
user    85m44.586s
sys     3m35.693s
```

ClangAnalyser profile with this PR (based on Clang's `-ftime-trace` feature) in the `./build` directory:
```bash
ClangBuildAnalyzer --all . gr4_all_before.bin; ClangBuildAnalyzer --analyze gr4_all_before.bin > gr4_all_before.txt
```
**before:**  [gr4_all_before.txt](https://github.com/user-attachments/files/16884919/gr4_all_before.txt)
```text
Analyzing build trace from 'gr4_all_before.bin'...
**** Time summary:
Compilation (184 times):
  Parsing (frontend):         2265.8 s
  Codegen & opts (backend):   8065.3 s
  [..]
```
**after:** [gr4_all_after.txt](https://github.com/user-attachments/files/16875660/gr4_all_after.txt)
```text
Analyzing build trace from 'gr4_all_after.bin'...
**** Time summary:
Compilation (184 times):
  Parsing (frontend):         2121.1 s
  Codegen & opts (backend):   4473.3 s
  [..]
```

### DIFF
--- a/core/include/gnuradio-4.0/Settings.hpp
+++ b/core/include/gnuradio-4.0/Settings.hpp
@@ -23,6 +23,12 @@
 #include <fmt/chrono.h>
 #pragma GCC diagnostic pop
 
+#if defined(__GNUC__) || defined(__clang__)
+#define NO_INLINE __attribute__((noinline))
+#else
+#define NO_INLINE
+#endif
+
 namespace gr {
 
 namespace detail {
@@ -102,8 +108,8 @@ inline constexpr void hash_combine(std::size_t& seed, const T& v) noexcept {
 } // namespace detail
 
 struct SettingsCtx {
-    uint64_t  time    = 0ULL; // UTC-based time-stamp in ns, time from which the setting is valid, 0U is undefined time
-    pmtv::pmt context = "";   // user-defined multiplexing context for which the setting is valid
+    std::uint64_t time    = 0ULL; // UTC-based time-stamp in ns, time from which the setting is valid, 0U is undefined time
+    pmtv::pmt     context = "";   // user-defined multiplexing context for which the setting is valid
 
     bool operator==(const SettingsCtx&) const = default;
 
@@ -295,6 +301,14 @@ class CtxSettings : public SettingsBase {
 
     const std::size_t _timePrecisionTolerance = 100; // ns, now used for emscripten
 
+    template<typename Func>
+    inline constexpr static void processMembers(Func func) {
+        // reflect and iterate via known base-class member fields
+        refl::util::for_each(refl::reflect<typename std::remove_cvref_t<TBlock>::base_t>().members, func);
+        // reflect and iterate via known derived class member fields
+        refl::util::for_each(refl::reflect<TBlock>().members, func);
+    }
+
 public:
     // Settings configuration
     std::uint64_t expiry_time{std::numeric_limits<std::uint64_t>::max()}; // in ns, expiry time of parameter set after the last use, std::numeric_limits<std::uint64_t>::max() == no expiry time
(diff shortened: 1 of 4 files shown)


core/include/gnuradio-4.0/Settings.hpp
```diff
@@ -506,47 +520,53 @@ class CtxSettings : public SettingsBase {
...
+                // auto                         notAutoUpdateView       = parameters.value() | std::views::filter([&](const auto& pair) { return !currentAutoUpdateParams.contains(pair.first); });
+                // property_map                 notAutoUpdateParams(notAutoUpdateView.begin(), notAutoUpdateView.end());
+
+                // the following is more compile-time friendly
+                property_map notAutoUpdateParams;
+                for (const auto& pair : parameters.value()) {
+                    if (!currentAutoUpdateParams.contains(pair.first)) {
+                        notAutoUpdateParams.insert(pair);
+                    }
                 }
```
drslebedev, line 549: I'm not sure about the gain of compilation time, but we might still need to consider using ranges, where possible?

core/include/gnuradio-4.0/Settings.hpp
```diff
@@ -914,17 +934,18 @@ class CtxSettings : public SettingsBase {
...
             _autoUpdateParameters[ctx] = getBestMatchAutoUpdateParameters(ctx).value_or(_allWritableMembers);
         }
 
-        _storedParameters[ctx.context].push_back({ctx, newParameters});
-        auto& vec = _storedParameters[ctx.context];
-        std::sort(vec.begin(), vec.end(), [](const auto& a, const auto& b) { return a.first.time < b.first.time; });
+        std::vector<std::pair<SettingsCtx, property_map>>& sortedVectorForContext = _storedParameters[ctx.context];
+        // binary search and merge-sort
+        auto it = std::ranges::lower_bound(sortedVectorForContext, ctx.time, std::less<>{}, [](const auto& pair) { return pair.first.time; });
+        sortedVectorForContext.insert(it, {ctx, newParameters});
```
drslebedev, line 945: `std::vector::insert` can be quite costly, so using `std::sort` might be a better option in this case. However, without concrete benchmark measurements, it's difficult to make a definitive judgment, and I don't have a strong preference either way.

drslebedev (approved):
Thanks for the PR!

Your changes look good. 
I left some comments in the code for consideration, but the PR can be merged as is.

core/include/gnuradio-4.0/Settings.hpp
RalphSteinhagen: The cost for this insert sort is lower than a regular sort because the list is already sorted. I.e. in most cases the field is added to the end. In the other it's one insert followed by a linear shift ( `O = N`) ... much faster than any other `O = N log N` sorts

core/include/gnuradio-4.0/Settings.hpp
RalphSteinhagen: re gain: see numbers above. Unfortunately quite significant and not in favour of ranges. :-|

PR #405, opened 2024-09-05
Partial support of MacOs

drslebedev:
This PR focuses on compiling and running GNU Radio 4 on macOS.

### Changes
- Several adjustments were made to ensure compatibility and successful execution on macOS.
- Tested on macOS 14.6.1 with both `gcc-13` and `gcc-14` installed via Homebrew.
- Encountered issues with the `libc++` library when using `clang-18`; the default Apple compiler is `clang-15`.

### Known Issues
- **HttpBlock**: Disabled due to compatibility issues with MacOsSDK. Further investigation is needed if support is required.
- **Thread Affinity**: Currently not supported. Further investigation may be required if needed.

### DIFF
--- a/core/test/qa_buffer.cpp
+++ b/core/test/qa_buffer.cpp
@@ -115,7 +115,9 @@ const boost::ut::suite SequenceTests = [] {
     "Sequence"_test = [] {
         using namespace gr;
         using signed_index_type = std::make_signed_t<std::size_t>;
+#if not defined(__APPLE__)
         expect(eq(alignof(Sequence), 64UZ));
+#endif
         expect(eq(0L, kInitialCursorValue));
         expect(nothrow([] { Sequence(); }));
         expect(nothrow([] { Sequence(2); }));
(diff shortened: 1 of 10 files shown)


core/test/qa_buffer.cpp
```diff
@@ -115,7 +115,9 @@ const boost::ut::suite SequenceTests = [] {
     "Sequence"_test = [] {
         using namespace gr;
         using signed_index_type = std::make_signed_t<std::size_t>;
+#if not defined(__APPLE__)
         expect(eq(alignof(Sequence), 64UZ));
```
RalphSteinhagen, line 119: Why does this not hold for iOS?

RalphSteinhagen (approved):
LGTM! 

We need to ensure that the missing implementation for the thread handing on iOS (and eventually Windows) isn't missed. Users may have an expectation that when the respective functions are used that these either succeed or transparently fail. Presently these are silent functional failures.

PR #406, opened 2024-09-09
CircularBuffer: do not reset empty PublishableOutputRange

daniestevez:
When an empty PublishableOutputRange is requested, do not reset the index, offet and internal span of the buffer.

This solution was suggested by @drslebedev to fix bugs in which the buffer state would become corrupted and report negative values (casted to uint64_t) for the available number of elements, etc.

---

I have tested this change on the gr4-packet-modem codebase and everything seems to work correctly with the single-threaded scheduler. Things that mysteriously failed before now work. All the QA tests in gr4-packet-modem and gnuradio4 pass. There are still some things in gr4-packet-modem that don't work with the multi-threaded scheduler, but probably the problem is unrelated.

### DIFF
--- a/core/include/gnuradio-4.0/CircularBuffer.hpp
+++ b/core/include/gnuradio-4.0/CircularBuffer.hpp
@@ -288,9 +288,6 @@ class CircularBuffer {
 
         PublishableOutputRange() = delete;
         explicit PublishableOutputRange(Writer<U>* parent) noexcept : _parent(parent) {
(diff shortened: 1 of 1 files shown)


drslebedev (approved):
Thank you for the PR! 

The changes look good to me. 
The PR can be merged once all tests have passed.

wirew0rm:
@all-contributors please add @daniestevez for code

Issue #409, opened 2024-09-16
[5SP] Improve test code coverage


ivan-cukic:
The coverage is currently lower than it needs to be.
Write the tests to cover code that is not currently
tested. Refactor tests if needed

PR #413, opened 2024-09-19
Coverage: configure Sonarcube exclusions

wirew0rm:
Fix sonarcube configuration to correctly exclude everything that is not run by the CI and therefore cannot have any coverage data. Also exclude a bunch of auxiliary files that are part of the build system or generic helper tools.

### DIFF
--- a/.github/workflows/restyled.yml
+++ b/.github/workflows/restyled.yml
@@ -1,0 +1,23 @@
+name: Restyled
+
+on:
+  pull_request:
+
+concurrency:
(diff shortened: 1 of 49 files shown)


wirew0rm:
Fixing the CI related coverage issues is a prerequisite to #409 but due to the `qa_Messages` test not passing on the CI, no coverage/sonarcube reports are published and this cannot be tested properly...

RalphSteinhagen (approved):
LGTM. Can be merged once CI passes.

PR #414, opened 2024-09-20
Update of StreamToDataSet block

drslebedev:
* Support several parallel DataSet accumulation
* Support sample accumulation between work() iterations
* Update unit tests

Notes:
- `n_pre` must be less than the size of the output port's `CircularBuffer`. The blok waits until all `n_pre` samples can be written to the output in a single iteration.
- In stream-to-stream mode, the "stop" Tag is not included in the output.
- Stream-to-stream mode has limited support for overlapping triggers. Configurations such as Start-Stop-Start-Stop work as expected, but configurations like Start-Start-Stop-Stop result in undefined behavior. If data accumulation is active, the second "start" is ignored, and the output may vary depending on subsequent tags.
- In stream-to-dataset modes, the "stop" Tag is included in the DataSet output.
- Tags for pre-samples are not stored and are not included in the output.
- It is expected that "start" and "stop" Tags will come from different iterations. If multiple "start" Tags arrive in a single merged tag, only one DataSet is created. The same applies for "stop" Tags. This may lead to incorrect or unexpected output.
- Currently used Matcher doesn't really support overlapping DataSets

Example of Stream-to-Dataset with overlapping datasets and pre/post samples:
IMG_6168

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/StreamToDataSet.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/StreamToDataSet.hpp
@@ -2,20 +2,30 @@
 #define GNURADIO_STREAMTODATASET_HPP
 
 #include "gnuradio-4.0/TriggerMatcher.hpp"
-#include <gnuradio-4.0/algorithm/dataset/DataSetUtils.hpp>
 #include <gnuradio-4.0/Block.hpp>
 #include <gnuradio-4.0/DataSet.hpp>
 #include <gnuradio-4.0/HistoryBuffer.hpp>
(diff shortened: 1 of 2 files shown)


RalphSteinhagen (approved):
LGTM! :+1: 

I like the way you refactored the `processBulk(..)` into two independent sub-versions, one for 'stream-to-stream' and the other 'stream-to-dataset' this makes the testing much easier. Also, the latter allows for (partially overlapping) start-stop sequences to produce independent DataSets.... well done.

Thanks also for documenting the edge-cases that are presently UB. Perhaps, it would be better to transfer these from PR/code-only documentation to proper `Doc<"...">` doc string. This could be done in a follow-up commit thought.

PR #415, opened 2024-09-24
PublishablePortSpan

drslebedev:
This PR introduces `PublishablePortSpan`, which extends `PublishableSpan` by adding a convenient `publishTag` method.

Additionally, the following issues were addressed in this PR:
- Renamed `Block::input_tags_present()` to `inputTagPresent()`.
- Fixed a bug in the parameter name for the `Soapy` block.


core/include/gnuradio-4.0/BlockTraits.hpp
```diff
@@ -345,6 +345,12 @@ struct DummyPublishableSpan {
 };
 static_assert(PublishableSpan<DummyPublishableSpan<int>>);
 
+template<typename T>
+struct DummyPublishablePortSpan: public DummyPublishableSpan<T> {
```
RalphSteinhagen, line 349: @drslebedev @wirew0rm I am wondering whether we should merge both concepts to create basically only one `PublishableSpan`. :thinking:

core/include/gnuradio-4.0/Port.hpp
```diff
@@ -268,6 +268,11 @@ concept ConsumablePortSpan = ConsumableSpan<T> && requires(T span) {
     { *span.tags.begin() } -> std::same_as<const gr::Tag&>;
 };
 
+template<typename T>
+concept PublishablePortSpan = PublishableSpan<T> && requires(T span, property_map& tagData, Tag::signed_index_type index) {
```
RalphSteinhagen, line 272: Merge with PublishableSpan? Are there use-cases where we need to keep them separate?

core/include/gnuradio-4.0/Port.hpp
```diff
@@ -365,13 +371,29 @@ struct Port {
...
+    struct PublishablePortOutputRange : public WriterSpanType<spanReleasePolicy> {
+        ThisPortFullType* _port;
+        // explicit PublishablePortOutputRange(WriterSpanType<spanReleasePolicy>& span) : WriterSpanType<spanReleasePolicy>(span) {};
+
+        template<typename TSpan>
+        explicit PublishablePortOutputRange(TSpan&& span, ThisPortFullType* port) : WriterSpanType<spanReleasePolicy>(std::forward<TSpan>(span)), _port(port){};
+
+        ~PublishablePortOutputRange() = default;
+
+        void publishTag(property_map& tagData, Tag::signed_index_type index) { _port->publishTag(tagData, index); }
```
RalphSteinhagen, line 393: Maybe rename `index` to `tagOffset` to better express the intention that this is the index/offset w.r.t. the initial sample that is presented to the user.

RalphSteinhagen (approved):
Thanks for the update. 

Some minor comments/questions:
 * do we need both concept definitions (see inline comments)?
 * the use of the `ProducableSpan::publishTag(...)` should be changed in the few uses in the block lib (e.g. in: BasicFileIo.hpp, clock_source.hpp, TagMonitors.hpp).
 * Rename the parameter `index` to`tagOffset` and remove the '0' since this would be the default parameter value.

This should make the changes/naming more expressive. Once these minor changes are in, the PR could be merged as-is without further review.

P.S. please do not forget to 'check' this tasks in the corresponding issue. Thanks in advance.

core/include/gnuradio-4.0/Port.hpp
RalphSteinhagen: @wirew0rm the question is about
 a) naming/proliferation of concepts (similar dependent definitions)
 b) we need this named concept since users -- with this PR -- do not usually interact directly with the buffer-related concept.

core/include/gnuradio-4.0/CircularBuffer.hpp
```diff
@@ -187,6 +187,14 @@ enum class SpanReleasePolicy {
...
 };
 
+/*
+ * The enum specifies whether to use the blocking reserve() method or the non-blocking tryReserve() method
+ */
+enum class PublishableSpanReservePolicy {
+    Reserve,    // use blocking reserve() method
+    TryReserve, // use non-blocking tryReserve() method
+};
+
```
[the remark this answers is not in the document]
drslebedev: The plan is to transition everything to use non-blocking tryReserve, though it’s only partially implemented at the moment.

core/include/gnuradio-4.0/WaitStrategy.hpp
```diff
@@ -245,7 +245,7 @@ class SpinWait {
             yieldProcessor();
         }
     }
-#if defined( __EMSCRIPTEN__)
+#if defined( __EMSCRIPTEN__) || defined(__APPLE__)
```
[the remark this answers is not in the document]
drslebedev: with gcc-14 on MacOs I got 
```text
error: unrecognized instruction mnemonic, did you mean: ret, rev?
        rep
```
Therefore, I’ve temporarily disabled it for now.

PR #416, opened 2024-09-24
Use stdlib errorchecking

mattkretz:
I extracted the two commits that enable stdlib facilities for improved error checking from the update-vir-simd-0.4 branch.

If we are serious about providing a build mode that can detect all detectable errors at run-time, then I believe these flags should be used.


RalphSteinhagen:
closing this as stale.

PR #417, opened 2024-09-24
CMake: Default color diagnostics to force-on with ccache/ninja

mattkretz:
Reading diagnostics with ANSI color codes in between is unnecessarily hard. We now make use of the cmake documented switch to control color output. And we only force color on if the tooling suggest it might be necessary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,14 @@ else ()
     message(STATUS "ccache will not be used")
 endif ()
 
+# Prefer forced colored compiler output if there's a chance we'd otherwise get none at all.
+# Ninja and ccache "consume" the compiler output, breaking the terminal detection of compilers.
+# ccache tries to solve the problem, but can only do so if it determines that it's calling GCC or Clang. It uses a very lightweight heuristic, which breaks easily.
+if ((CMAKE_GENERATOR STREQUAL "Ninja" OR (CCACHE_PROGRAM AND USE_CCACHE)) AND NOT DEFINED CMAKE_COLOR_DIAGNOSTICS)
+    message(STATUS "Forcing compiler color output due to the use of Ninja and/or ccache. Use -DCMAKE_COLOR_DIAGNOSTICS=OFF to turn it off.")
+    set(CMAKE_COLOR_DIAGNOSTICS ON)
+endif()
+
 set(CMAKE_EXT_DEP_WARNING_GUARD "")
 if (DISABLE_EXTERNAL_DEPS_WARNINGS) # enable warnings for external dependencies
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -48,6 +48,14 @@ else ()
     message(STATUS "ccache will not be used")
 endif ()
 
+# Prefer forced colored compiler output if there's a chance we'd otherwise get none at all.
+# Ninja and ccache "consume" the compiler output, breaking the terminal detection of compilers.
+# ccache tries to solve the problem, but can only do so if it determines that it's calling GCC or Clang. It uses a very lightweight heuristic, which breaks easily.
+if ((CMAKE_GENERATOR STREQUAL "Ninja" OR (CCACHE_PROGRAM AND USE_CCACHE)) AND NOT DEFINED CMAKE_COLOR_DIAGNOSTICS)
```
wirew0rm, line 54:
```suggestion
if ( NOT DEFINED CMAKE_COLOR_DIAGNOSTICS)
```
Just always default to colored output but allow to turn it off in the standard cmake way?

wirew0rm (changes requested):
Diff looks good, in general I'm all for using the the facilities that cmake is providing.

Unfortunately it does not seem to detect the CI correctly and switches off color there.
E.g. compare [main](https://github.com/fair-acc/gnuradio4/actions/runs/11027351234/job/30625507896#step:7:1480) and [this PR](https://github.com/fair-acc/gnuradio4/actions/runs/11010089323/job/30571207094?pr=417#step:7:1479).

It should imo either default more aggressively to colored output or the `-DCMAKE_COLOR_DIAGNOSTICS=ON` flag would at least have to be added to the CI.

CMakeLists.txt
mattkretz: If tried this before and it'd be a very unfortunate outcome. GCC and Clang are designed to do the right thing automatically without getting told from outside. I.e. if I build from vim or pipe output from make into a file, the compiler outputs no color codes. But if I build in a terminal (also from a terminal inside vim) I get color output. Except when I force it off via `CMAKE_COLOR_DIAGNOSTICS=OFF`.

The "problem" when running as GitHub action is that the compiler doesn't see a terminal. So in order to see colored output, yes one has to force it to `ON`, that's the technically correct solution here.

wirew0rm (approved):
Thanks for also changing the CI to keep the output colored there.

I also tested in my IDE (clion, with CCACHE and with `-DUSE_CCACHE=OFF` and gnumake to not run in the forced mode also in both building via the Build->BuildProjct menu and the builtin terminal) and it correctly used colored output everywhere.

I don't see negative side effects of this and form my PoV this could be merged.

Issue #407, opened 2024-09-11
Improve handling of tags with the processBulk Consumable/PublishableRange API


wirew0rm:
When using the `processBulk` API, special care has to be taken when publishing samples and the corresponding tags. At the moment this works with the automatic tag propagation which has some shortcomings:
- can only correctly forward tags on the first sample
- difficult to disable/override
To fix these, the ` Consumable/PublishablePortRange` API should be extended to allow for an intuitive way to control the propagation of tags. There was some initial offline discussion which is persisted in this issue involving @RalphSteinhagen, @drslebedev and @wirew0rm, but others are of course invited to chime in if they have a specific usecase or improvement in mind. The Proposed changes are roughly structured into four individual steps that can be tackled more or less independently.

- [x] I: add convenience methods for 
   - consuming tags in `ConsumableInputPortRange` (Concept: `ConsumablePortSpan`)
  -  at the moment, one has to directly interact with the `PublishableInputRange<Tag>`, which is a member of the PortRange
  - [x] `tag()` -> returning range of tuple/pair `<size_t, property_map>` (size_t w.r.t. local index) -- complementary to the `tag` -> renaming -> `rawTags` field interface
  - [x] implement `consumableInputPortRange::consume(index=size::MAX)`, which will consume all tags up to the given sample index (= all tags belonging to samples in the range)
- [ ] II: Automatic Tag Propagation Policies (only relevant for non-'N_in=1:M_out=1' chunked data constraints)
  -  optional block attribute (like e.g. `Resampling` or `BlockingIO`)
  - different policies:
    - forward-propagate (default): treat the tag as if it belonged to the first sample of the next `processBulk(..)` invocation
    - backward-propagate: treat the tag as if it were on the first sample of the current `processBulk(..)` invocation
    - [ignore]: tag propagation needs to be handled by the user in the `processBulk(..)` function (see point IV below)
- [x] III: convenience methods for publishing Tags in PublishablePortRange
  - [x] implement  `PublishableOutputPortRange`
  - [x] `PublishableOutputPortRange::publishTag(index = 0)`  N.B. `index` w.r.t. local indexing of the output-data-span
- [ ] IV: forwarding policy
  - at the moment we have all-to-all and one-to-one strategies defined, but only all-to-all is implemented
  - [We discussed providing a customization point where users could plug in predefined or their own strategies -> abandoned]
[...]

PR #422, opened 2024-09-26
ConsumableInputPortRange Tag API

wirew0rm:
As outlined in #407 I.) this PR introduces a way to directly access and consume tags on the current ConsumableSpan by introducing a `tags()` function which will return a view of all tags with the indices shifted to correspond to the local span indices and a `consumeTags(untilLocalIndex = 1)` which will consume all tags up to the local sample index in the span. The original Tags with the global index can still be accessed via the `rawTags` member.

This is mostly done, there is still a small issue with the concept for the `tags()` function, which is for now commented out and has to be either fixed or removed.

### DIFF
--- a/core/include/gnuradio-4.0/Block.hpp
+++ b/core/include/gnuradio-4.0/Block.hpp
@@ -744,10 +744,10 @@ public:
                         destinationMap.insert_or_assign(key, value);
                     }
                 };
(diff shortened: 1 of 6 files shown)


RalphSteinhagen (approved):

drslebedev (approved):
Thanks a lot for your PR!

As also discussed privately the changes look good and PR can be merged!

Issue #427, opened 2024-09-27
qa_buffer fails with a precondition failure in span::operator[]


mattkretz:
Compile with `-Og -g -D_GLIBCXX_ASSERTIONS` should do the trick to repro this one. (I triggered it with my patched hardened GCC 15.)

```
Program received signal SIGILL, Illegal instruction.
0x000000000042a4da in std::span<int const, 18446744073709551615ul>::operator[] (this=this@entry=0x7fffffffc950, __idx=__idx@entry=0) at /home/mkretz/.local/gcc-master/include/c++/15.0.0/span:286
286             __glibcxx_precondition(__idx < size());
(gdb) bt
#0  0x000000000042a4da in std::span<int const, 18446744073709551615ul>::operator[] (this=this@entry=0x7fffffffc950, __idx=__idx@entry=0) at /home/mkretz/.local/gcc-master/include/c++/15.0.0/span:286
#1  0x000000000042a4ee in gr::CircularBuffer<int, 18446744073709551615ul, (gr::ProducerType)0, gr::SleepingWaitStrategy>::ConsumableInputRange<int, (gr::SpanReleasePolicy)2>::operator[] (this=0x7fffffffc940, i=0) at /home/mkretz/src/gnuradio4-main/core/include/gnuradio-4.0/CircularBuffer.hpp:572
#2  consumableInputRangeTest1<gr::CircularBuffer<int, 18446744073709551615ul, (gr::ProducerType)0, gr::SleepingWaitStrategy>::ConsumableInputRange<int, (gr::SpanReleasePolicy)2> > (in=...) at /home/mkretz/src/gnuradio4-main/core/test/qa_buffer.cpp:25
#3  0x0000000000418e50 in operator()<CircularBufferTestTypes<gr::CircularBuffer<int>, AllocatorPosix> > (__closure=__closure@entry=0x7fffffffcbdf) at /home/mkretz/src/gnuradio4-main/core/test/qa_buffer.cpp:105
[...]

PR #430, opened 2024-09-30
Fix qa_buffer precondition failure #427

drslebedev:
See #427


core/test/qa_buffer.cpp
```diff
@@ -22,13 +22,13 @@ struct TestStruct {
...
 
-void consumableInputRangeTest1(auto in) { [[maybe_unused]] const auto inFirst = in[0]; }
+void consumableInputRangeTest1(auto in) { [[maybe_unused]] volatile auto isEmpty = in.empty(); }
 
-void consumableInputRangeTest2(gr::ConsumableSpan auto in) { [[maybe_unused]] const auto inFirst = in[0]; }
+void consumableInputRangeTest2(gr::ConsumableSpan auto in) { [[maybe_unused]] volatile auto isEmpty = in.empty(); }
 
 template<typename T>
 void consumableInputRangeTest3(std::span<const T> in) {
-    [[maybe_unused]] const auto inFirst = in[0];
```
mattkretz, line 31: What is this testing anyway? The previous code seems to test that `operator[]` exists. The new test is whether `empty()` exists? How useful is this? To me that looks like something you'd rather `static_assert` and then these tests can be removed altogether?

drslebedev: I think the main idea of these test is to test the acceptable signatures of the function.

drslebedev: I think these tests can be removed, there are similar test in `qa_Block` which test different `processBulk` signatures.

mattkretz: LGTM :+1:

drslebedev: Could you please approve the PR? Thank you!

wirew0rm (approved):
LGTM, thanks for following this up! :+1:

PR #431, opened 2024-09-30
Block: add std::complex as a special case in reflFirstTypeName

daniestevez:
This makes `reflFirstTypeName<std::complex<float>>()` return `"std::complex<float>"s` instead of `"std::complex<T>"s`, and likewise for `std::complex<double>`.

### DIFF
--- a/core/include/gnuradio-4.0/Block.hpp
+++ b/core/include/gnuradio-4.0/Block.hpp
@@ -2024,6 +2024,8 @@ std::string reflFirstTypeName() {
         return fmt::format("gr::DataSet<{}>", reflFirstTypeName<typename Type::value_type>());
     } else if constexpr (UncertainValueLike<Type>) {
         return fmt::format("gr::UncertainValue<{}>", reflFirstTypeName<typename Type::value_type>());
+    } else if constexpr (pmtv::Complex<Type>) {
+        return fmt::format("std::complex<{}>", reflFirstTypeName<typename Type::value_type>());
     } else if constexpr (refl::is_reflectable<Type>()) {
         return refl::reflect<Type>().name.str();
 
--- a/core/test/qa_Block.cpp
+++ b/core/test/qa_Block.cpp
@@ -890,4 +890,18 @@ const boost::ut::suite<"Requested Work Tests"> _requestedWorkTests = [] {
         expect(eq(resultSink.performed_work, 100UZ));
     };
 };
+
+const boost::ut::suite<"reflFirstTypeName Tests"> _reflFirstTypeNameTests = [] {
+    using namespace boost::ut;
+    using namespace gr::detail;
+    using namespace std::literals::string_literals;
+
+    "std::complex"_test = []() {
+        expect(eq(reflFirstTypeName<std::complex<double>>(),
+                  "std::complex<double>"s));
+        expect(eq(reflFirstTypeName<std::complex<float>>(),
+                  "std::complex<float>"s));
+    };
+};
+
 int main() { /* not needed for UT */ }


core/include/gnuradio-4.0/Block.hpp
```diff
@@ -2024,6 +2037,8 @@ std::string reflFirstTypeName() {
         return fmt::format("gr::DataSet<{}>", reflFirstTypeName<typename Type::value_type>());
     } else if constexpr (UncertainValueLike<Type>) {
         return fmt::format("gr::UncertainValue<{}>", reflFirstTypeName<typename Type::value_type>());
+    } else if constexpr (is_complex_v<Type>) {
+        return fmt::format("std::complex<{}>", reflFirstTypeName<typename Type::value_type>());
```
drslebedev, line 2041: You can also use `pmtv::Complex` concept here:
`if constexpr (pmtv::Complex<Type>)`

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -2010,6 +2010,19 @@ template<BlockLike TBlock>
...
+};
+
+template<typename T>
+struct is_complex<std::complex<T>> {
+    static constexpr bool value = true;
+};
+
+template<typename T>
+constexpr bool is_complex_v = is_complex<T>::value;
+
```
drslebedev, line 2025: See also comment below:  we already have pmtv::Complex concept, please use it.

drslebedev (approved):
Thank you for the PR.
Looks good to me!

It can be merged once the comments are addressed.

core/include/gnuradio-4.0/Block.hpp
daniestevez: Looking at https://github.com/gnuradio/pmt/blob/main/include/pmtv/type_helpers.hpp#L67, that concept seems insufficient if someone attempts to use something like `std::complex<int16_t>`, which can happen when doing fixed-point arithmetic implementations. But in any case I'll do this change, since I didn't know that `pmtv` had a `Complex` concept and it doesn't make sense to duplicate it here. If needed, the `pmtv::Complex` concept can be extended.

drslebedev: I believe `std::complex` only supports value types of `float`, `double`, and `long double`.

daniestevez: It also supports integer types, and it works as you would expect: https://godbolt.org/z/soxf5c6oG

daniestevez: Updated to use `pmtv::Complex` and force-pushed.

mattkretz:
> return `"std::complex"s` instead of `"std::complex"s`

@daniestevez I assume GitHub ate some of your characters. I'd like to understand what you're improving here as I was looking at a generalization of this problem.

core/include/gnuradio-4.0/Block.hpp
drslebedev: Even if it is possible to instantiate `std::complex` with integer types like `int16_t`, doing so is generally discouraged because complex number operations, especially those involving trigonometric or transcendental functions, rely on floating-point precision. Using integer types can lead to truncation, overflow, and incorrect results for functions like `std::sin()`, which are not designed to handle complex integers. The provided code, while it compiles, may exhibit undefined behavior or errors when performing certain operations due to the lack of proper floating-point precision. 

In any case, this can be discussed if one wants to include complex integer support in the `pmt` library, but it is probably out-of-scope for this PR.

mattkretz:
Ah, your commit log has it right. :+1: all good

daniestevez:
> > return `"std::complex"s` instead of `"std::complex"s`
> 
> @daniestevez I assume GitHub ate some of your characters. I'd like to understand what you're improving here as I was looking at a generalization of this problem.

It was an issue with Markdown formatting in the PR description (which was copied from the commit log). I've edited it for clarity.

PR #436, opened 2024-10-04
InputSpan and OutputSpan

drslebedev:
* `InputSpan` non-const, get() and consume() - non-const
* `ConstSpanLvalueLike` is replaced with `ConstSpanLike`, also `operator std::span<const T>&&() = delete;` is removed from `ReaderSpan`. 

### New names
Here is the new naming scheme (ConceptName -> ImplementationName)
* BufferLike -> CircularBuffer
  * BufferReaderLike -> CircularBuffer::Reader
    * ReaderSpanLike -> CircularBuffer::ReaderSpan
    * InputSpanLike -> Port::InputSpan
  * BufferWriterLike -> CircularBuffer::Writer
    * WriterSpanLike -> CircularBuffer::WriterSpan
    * OutputSpanLike -> Port::OutputSpan

### Important Note:
References are required for processBulk when manipulating Tags inside processBulk because InputSpan and OutputSpan have internal states (e.g., tagsPublished) that are unique to each instance. Copying these objects without proper state management can lead to incorrect behavior when the publishTag() method is called.
This issue must be resolved in future updates, see https://github.com/fair-acc/gnuradio4/issues/439

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
@@ -350,9 +350,9 @@ public:
                 return false;
             }
 
-            const auto readData = reader.get(nProcess);
+            auto readData = reader.get(nProcess);
             if constexpr (requires { fnc(std::span<const T>(), std::span<const Tag>()); }) {
-                const auto tags         = tag_reader.get();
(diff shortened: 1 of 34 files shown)


wirew0rm (approved):
Looks good, thanks for following this up and documenting the still existing problems. Can be merged once the CI has passed.

PR #437, opened 2024-10-06
Add edge definition API

ivan-cukic:


core/include/gnuradio-4.0/Port.hpp
```diff
@@ -568,6 +565,24 @@ struct Port {
 
     [[nodiscard]] constexpr static bool isOptional() noexcept { return kIsOptional; }
 
+    [[nodiscard]] constexpr std::size_t nReaders() const noexcept {
+        if constexpr (kIsInput) {
+            return -1UZ;
```
wirew0rm, line 570: Was wondering if this should/could be a runtime or compile time error(the function only defined for Outputs)? Compile time error probably won't work because of the virtual interface and use in the runtime API, runtime error does not work because of noexcept. Would it be useful to have an `assert(false)`?

Functionally I think this is good, but it kind of communicates that the user will have to check the number of readers against MAX_SIZE_T which in my understanding of the code should never happen -> it's e.g. sufficient to check for `edge.readers > 0`  to check for presence of readers.

Ties a bit into the bigger discussion on how we handle errors in general, so this is probably not really actionable right here.

wirew0rm (approved):
Looks good, 2 small comments/comprehension questions.

core/include/gnuradio-4.0/Port.hpp
```diff
@@ -568,6 +565,24 @@ struct Port {
...
+        } else {
+            return _ioHandler.buffer().n_readers() > 0;
+        }
+    }
+
+    [[nodiscard]] constexpr std::size_t nWriters() const noexcept {
+        if constexpr (kIsInput) {
+            return _ioHandler.buffer().n_writers();
+        } else {
+            return -1UZ;
```
frankosterfeld, line 580: Why, `-1UZ`? Why not `0`, or, if this is considered an error, add a failing static_assert?

core/include/gnuradio-4.0/Port.hpp
```diff
@@ -568,6 +565,24 @@ struct Port {
 
     [[nodiscard]] constexpr static bool isOptional() noexcept { return kIsOptional; }
 
+    [[nodiscard]] constexpr std::size_t nReaders() const noexcept {
+        if constexpr (kIsInput) {
+            return -1UZ;
+        } else {
+            return _ioHandler.buffer().n_readers() > 0;
```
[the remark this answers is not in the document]
ivan-cukic: Fixed

core/include/gnuradio-4.0/Port.hpp
ivan-cukic: assert(false) would be ok, but I'm not sure we should do it. As this is a general/unified interface, I'm more for communicating the error as -1 - the user is not going to need to first check the type of the port and then call the functions, just call and ignore -1 if that is the result.

core/include/gnuradio-4.0/Port.hpp
ivan-cukic: See reply to Alex above

PR #438, opened 2024-10-07
Switch Sequence and Tag indices to unsigned

frankosterfeld:
This PR changes sequence and tag index types to unsigned (std::size_t).

There's three cases where we need relative tag indices though:

 - Port::tags() provides the tags with indices relative to stream position. (these are of typestd::pair<index,map> anyway, so no extra type needed)
 - DataSet::timing_events are relative to the first sample in the set, and can
   be negative (e.g. for DataSink's snapshot acquisition mode, the trigger is
   in the past.
 - DataSink callback and pollers use tags relative to the current data chunk.

For the latter two, use RelativeIndexTag, which have a signed index type.

This doesn't actually handle overflow of the std::size_t, which I think would be an issue in the CircularBuffer impl (which relies on monotonously growing indices), so the question is if an overflow here is a realistic scenario we need to care about.

### DIFF
--- a/core/include/gnuradio-4.0/Tag.hpp
+++ b/core/include/gnuradio-4.0/Tag.hpp
@@ -65,18 +65,15 @@ concept PropertyMapType = std::same_as<std::decay_t<T>, property_map>;
  * may choose to chunk the data based on the MIN_SAMPLES/MAX_SAMPLES criteria only, or in addition break-up the stream
  * so that there is only one tag per scheduler iteration. Multiple tags on the same sample shall be merged to one.
  */
-struct alignas(hardware_constructive_interference_size) Tag {
-    using signed_index_type = std::make_signed_t<std::size_t>;
-    signed_index_type index{0};
-    property_map      map{};
+template<typename TIndex>
+struct alignas(hardware_constructive_interference_size) BasicTag {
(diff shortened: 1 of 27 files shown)


core/include/gnuradio-4.0/Tag.hpp
```diff
@@ -109,6 +106,21 @@ struct alignas(hardware_constructive_interference_size) Tag {
     void insert_or_assign(const std::string& key, const pmtv::pmt& value) { map[key] = value; }
 };
 
+using Tag              = BasicTag<std::size_t>;
+using RelativeIndexTag = BasicTag<std::make_signed_t<std::size_t>>;
+
+constexpr std::make_signed_t<std::size_t> tagIndexDifference(Tag::index_type x, Tag::index_type y) {
+    // assuming that tag indices and stream position are never further apart than half the range of the index type,
+    // assume integer overflow if they are.
```
frankosterfeld, line 114: This code only makes sense if we decide to care about std::size_t overflow, otherwise it should be simplified.

RalphSteinhagen (approved):
Thanks for incorporating the changes we discussed, and following and fixing the signed int overflow UB. Well done! :+1:

PR #441, opened 2024-10-10
Restore connect<"out", 0> / to<"in", 0> functionality for arrays

mattkretz:
With my last PR `array<PortOut, 2> out` turned into two named ports: "out0" and "out1". Consequently, existing `connect<"out", 0>` calls had to be changed to `connect<"out0">`.

This commit introduces the following changes:

1. The ports are now named "out#0" and "out#1" so as to never clash with another port member named "out0", etc.

2. `connect<"out", 0>` & `to<"in", 0>` will now check for a port with the given name. If that doesn't exist try `name + '#' + index`. If that also doesn't exist try `name + index`. Thus, the above array of ports can be connected both via `connect<"out0">` and `connect<"out", 0>`.

3. The fallback to `name + index` allows indexing of explicitly numbered port members.

4. `merge` will refuse to merge `vector<Port>` with a helpful error message.

Implementation detail:

`indexForName` moved from `meta::` into `traits::block::` because that's where all the related port reflections on blocks are implemented.

TODO: Non-`MergedGraph` blocks can never have duplicate port names by construction now. However, `MergedGraph` blocks can have arbitrary duplication. Since indexing ports of `MergedGraph`s is also fairly brittle (the numbering is not intuitive and could change if we ever try to make it more intuitive) we should consider a naming scheme that disambiguates port names.

### DIFF
--- a/blocks/soapy/src/soapy_example.cpp
+++ b/blocks/soapy/src/soapy_example.cpp
@@ -36,21 +36,21 @@ gr::Graph createGraph(std::string fileName1, std::string fileName2, gr::Size_t m
     if (fileName1.contains("null")) {
         fmt::println("write channel0 to NullSink");
         auto& fileSink1 = flow.emplaceBlock<testing::NullSink<TDataType>>();
-        expect(eq(gr::ConnectionResult::SUCCESS, flow.connect<"out0">(source).to<"in">(fileSink1))) << "error connecting NullSink1";
+        expect(eq(gr::ConnectionResult::SUCCESS, flow.connect<"out", 0UZ>(source).to<"in">(fileSink1))) << "error connecting NullSink1";
     } else {
         fmt::println("write to fileName1: {}", fileName1);
         auto& fileSink1 = flow.emplaceBlock<BasicFileSink<TDataType>>({{"file_name", fileName1}, {"mode", "multi"}, {"max_bytes_per_file", maxFileSize}});
-        expect(eq(gr::ConnectionResult::SUCCESS, flow.connect<"out0">(source).to<"in">(fileSink1))) << "error connecting BasicFileSink1";
+        expect(eq(gr::ConnectionResult::SUCCESS, flow.connect<"out", 0UZ>(source).to<"in">(fileSink1))) << "error connecting BasicFileSink1";
     }
 
     if (fileName2.contains("null")) {
         fmt::println("write channel1 to NullSink");
         auto& fileSink2 = flow.emplaceBlock<testing::NullSink<TDataType>>();
-        expect(eq(gr::ConnectionResult::SUCCESS, flow.connect<"out1">(source).to<"in">(fileSink2))) << "error connecting NullSink2";
+        expect(eq(gr::ConnectionResult::SUCCESS, flow.connect<"out", 1UZ>(source).to<"in">(fileSink2))) << "error connecting NullSink2";
     } else {
         fmt::println("write to fileName2: {}", fileName2);
         auto& fileSink2 = flow.emplaceBlock<BasicFileSink<TDataType>>({{"file_name", fileName2}, {"mode", "multi"}, {"max_bytes_per_file", maxFileSize}});
-        expect(eq(gr::ConnectionResult::SUCCESS, flow.connect<"out1">(source).to<"in">(fileSink2))) << "error connecting BasicFileSink2";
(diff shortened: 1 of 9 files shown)


wirew0rm:
Probably related to this i noticed when [updating gr-digitizer to current gr4](https://github.com/fair-acc/gr-digitizers/pull/148) that the processBulk signature for `std::span<Port>` has changed to enumerating all the ports in the arguments. For `std::vector<Port>` the existing signature with `processBulk(std::span<OutputSpanLike> &outputs)` still works.
Even if the number of ports is known and it's possible to write like this, it still makes iterating over the ports harder and in the Picoscope case , the process bulk is defined in a baseclass and different implementations define different numbers of ports. I have now worked around it by replacing the port spans with port vectors, but since this sounds related to this issue I wanted to check if it's possible to reenable the old signatures.

If it's unrelated I'd also be willing to follow this up myself, I just didn't want to get into the way in parts that are still changing.

mattkretz:
I didn't even know that a span of Ports makes sense. Can you point to an example that uses that? If it should be supported we probably want to have a test case in gr4.

mattkretz:
The reason I demoted this PR to draft is that apparently qa_Block never terminates. If I have to guess, I'd say some connection failed and thus some incomplete graph is waiting on a condition that can never happen. But I currently have no idea how this patch managed to introduce such an error (without a compile-time error).

wirew0rm:
> I didn't even know that a span of Ports makes sense. Can you point to an example that uses that? If it should be supported we probably want to have a test case in gr4.

Sorry, I meant `std::array<Port>`, corrected. If I interpret your PR description correctly, the changes in here might re-enable the old signature or at least make it a lot easier to make it available.

 I'll try to have a look at the qa_Block test failure with this PR.


Also small question: in your description you wrote
> Thus, the above array of ports can be connected both via connect<"out0"> and connect<"out", 0>

I assume you meant `connect<"out#0">` (with hash) for the first variant? otherwise there would have to be a lot of strange string logic.

mattkretz:
Ah yes. `array<Port, N>` came up when I was moving `duplicate<T, N>` from template argument ports to member ports. Since this block was used in merges, the member ports needed to be reflected *individually* — which was the case for the template argument.
There's a second reason: when I initially implemented member-based merging, I added the resulting ports of the `MergedGraph` as `std::tuple` to the new type. So I implemented reflection of tuples of Ports. And since `std::array` is tuple-like, there was another good reason to reflect over Ports in arrays individually (just like Ports in a tuple).

What I realized only later, was that array of Ports were passed as "`span` of `span`" into `processBulk`. For the Soapy block the resulting API change made the implementation cleaner (IMHO). But I can imagine cases where it's not necessarily better. In any case it's a breaking change (source needs to adjust / functionality isn't broken).

So while this PR is aiming to restore the connect API, the `processBulk` (and `processOne`, FWIW) API changes are currently not on my radar. I believe it could be possible to detect when a `processBulk` is defined with a "`span` of `span`" instead of a parameter list of `span`s and then call it accordingly. So if we want that flexibility on the Block implementers side, that'd be another (new) Issue/PR.

mattkretz:
> > Thus, the above array of ports can be connected both via connect<"out0">
> > and connect<"out", 0>
> 
> I assume you meant connect<"out#0"> (with hash) for the first variant?
> otherwise there would have to be a lot of strange string logic.

Ah, I believe wrote that part of the commit log before I introduced the '#' 
into the string. :) Will fix.

wirew0rm:
> So while this PR is aiming to restore the connect API, the `processBulk` (and `processOne`, FWIW) API changes are currently not on my radar. I believe it could be possible to detect when a `processBulk` is defined with a "`span` of `span`" instead of a parameter list of `span`s and then call it accordingly. So if we want that flexibility on the Block implementers side, that'd be another (new) Issue/PR.

I think for now we have a working solution without changing too much code by just using a `vector<Port>`. I also realized over the weekend that reflecting ports individually would also allow to use variadic teplated functions, instead of a span based API, with the advantage that we do not need to explicitly have some memory for the spans to point to, but on the other hand this makes writing the process bulk a lot harder, especially for people not familiar with handling parameter packs and without access to c++26 pack indexing.

Sometihing to revisit later.

mattkretz:
Also, parameter packs on `processBulk` won't work if you have two arrays of ports. You'd need two packs — which doesn't work. Or work with a single pack and then somehow untangle that pack. But yes, the span of span isn't the most efficient in terms of passing the function arguments.

RalphSteinhagen:
Will close this as obsoleted by PR https://github.com/fair-acc/gnuradio4/pull/549

PR #443, opened 2024-10-15
Add DataSetSink block

frankosterfeld:
Add sink to export DataSet<T> streams. This block is a simpler DataSink
(which operates of non-DataSet streams), basically handing out the
incoming data sets as is, allowing basic filtering.


drslebedev (approved):
Thanks a lot for your PR!

The new `DataSetSink` block implementation looks good to me.
The code is also well-tested for both poller and callback cases.

The PR can be merged once all tests have passed.

Issue #444, opened 2024-10-17
Tags forwarding when some ports are not connected or `Aync`


drslebedev:
Handling tags in a system with both synchronous and asynchronous ports presents challenges due to differences in data flow. 

Currently, our implementation does not check whether a port is connected or if it is synchronous or asynchronous. We calculate the minimum number of processed samples across input and output ports, and only if both `processedIn > 0` and `processedOut > 0` do we forward tags.

UPD In PR #223, we've introduced a check to determine if a port is connected, which partially addresses the issue.

### General Problem
The main challenge is determining how to forward tags to output asynchronous ports effectively:
- **Question 1:** Should we forward tags to output asynchronous ports if at least one sample is present on the asynchronous port?
Asynchronous ports may produce data independently including tags forwarding.

- **Question 2:** Should we cache tags for each port individually?
This allows for independent management of tags per port, ensuring that tags are accurately associated with the data on each specific port, whether synchronous or asynchronous.

PR #445, opened 2024-10-17
Check if a port is connected or not

drslebedev:
Currently, our implementation does not check whether a port is connected or if it is synchronous or asynchronous. We calculate the minimum number of processed samples across input and output ports, and only if both processedIn > 0 and processedOut > 0 do we forward tags.

This PR introduces a check to determine if a port is connected, which partially addresses the issue.

This PR helps to fix tag forwarding prolem for `Picoscope` block.

See https://github.com/fair-acc/gnuradio4/issues/444 for more explanations.


wirew0rm (approved):
Looks good as discussed and outlined in the linked issue :+1: 
Thanks for following this up!

PR #449, opened 2024-10-28
Fix ClockSource to work correctly in infinite mode.

drslebedev:
This PR addresses the issue with `ClockSource` not functioning correctly in infinite mode (`n_sample_max == 0`). Additionally, it resolves a problem where tags were not published properly when there was an early return due to insufficient slots in the output span.


RalphSteinhagen (approved):

drslebedev:
I've merged this PR despite the qa_Message test failing. The issue causing the test failure is not related to the changes in this PR and will be addressed in the next one.

PR #454, opened 2024-10-29
Fix qa_Message

drslebedev:
After making changes to ClockSource and fixing a bug where it wasn't functioning properly (see https://github.com/fair-acc/gnuradio4/pull/449), I noticed that the `qa_Message` test named "Settings handling via scheduler" started failing.

Previously, the test used ClockSource, which essentially produced no samples due to the bug. After my fix, the test began to fail because `ClockSource` started producing samples as expected.

Upon further investigation, I found that the issue isn't with the message handling but rather with BlockingIO—since `ClockSource` has the `BlockingIO` attribute. I created a small unit test that reproduces the error reliably.

The core problem is that` ClockSource` continues running in a separate thread for some time, and the detection mechanism to determine whether it has completed isn't always accurate.

This PR includes:
- Additional verification that a `BlockingIO` block is in `STOPPED` state.
- Add new unit test in `qa_Block`.

### DIFF
--- a/core/test/qa_Block.cpp
+++ b/core/test/qa_Block.cpp
@@ -8,6 +8,7 @@
 #include <gnuradio-4.0/Block.hpp>
 #include <gnuradio-4.0/Graph.hpp>
 #include <gnuradio-4.0/Scheduler.hpp>
+#include <gnuradio-4.0/basic/clock_source.hpp>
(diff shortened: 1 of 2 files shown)


RalphSteinhagen (approved):
Thanks for the quick-fix and documenation. LGTM :+1:

PR #455, opened 2024-10-31
qa_Messages: increase timeouts and allow retry

wirew0rm:
The qa_Messages::"message passing via scheduler" test reproducibly fails on the (relatively resource constrained) CI runners for the scheduler's blocking execution strategy.
From the test output it could be observed, that the first get would timeout, then the set would fail because it would unexpectedly find the reply of the get command and then the second get would fail because it receives the reply of the first get command.
Because `checkReply` failed with an expect assertion instead of returning false, the retry parameter of the test was not working and the test would fail.

This increases the timeout for the tests from 2 to 4 seconds, reduces the timeout for the test that expects to timeout, to not increase execution times unnecessarily and changes the assertion of the received message contents to returning false.

Also I noticed the test will actually pass regardless of whether the test was sucessful which makes diagnosing harder than needed so i added an additional `mayFail` test parameter and added an assertion that the test actually passed successfully.

I hope that this will allow us to have all tests working in the CI again, however even on resource constrained systems we should not require timeouts in the order of seconds, so at some point it would be good to reduce these again by fixing the scheduler/block logic handling this message mechanism. Since it only affects the blocking execution, my suspicion is, that the strategy does not take changes in settings and/or messages into account when determining whether to sleep for 10ms -> message processing is throttled by the sample rate and iirc there is some factor between process_work and process message calls.

Also added an improvement to the scheduler so it will process messages always immediately after returning from sleeping (see separate commit).

### DIFF
--- a/core/include/gnuradio-4.0/Scheduler.hpp
+++ b/core/include/gnuradio-4.0/Scheduler.hpp
@@ -271,8 +271,8 @@ protected:
         }
 
         [[maybe_unused]] auto currentProgress    = this->_graph.progress().value();
-        gr::Size_t            inactiveCycleCount = 0U;
-        gr::Size_t            msgToCount         = 0U;
+        std::size_t           inactiveCycleCount = 0UZ;
+        std::size_t           msgToCount         = 0UZ;
         auto                  activeState        = this->state();
         do {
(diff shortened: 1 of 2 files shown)


RalphSteinhagen (approved):
Thanks for spotting the bug and fixing it! Well done. :+1:

PR #456, opened 2024-11-01
Add messages operations for handling Context

dantti:
- Activate current context
- Get current Context
- Create new Context
- Get X Context
- Remove X Context

### DIFF
--- a/core/include/gnuradio-4.0/Settings.hpp
+++ b/core/include/gnuradio-4.0/Settings.hpp
@@ -192,6 +192,17 @@ struct SettingsBase {
     virtual void resetDefaults() = 0;
 
     /**
+     * @brief return the name of the active context
+     */
+    [[nodiscard]] virtual const SettingsCtx& activeContext() const noexcept = 0;
+
+    /**
+     * @brief removes the given context
+     * @return true on success
+     */
+    [[nodiscard]] virtual bool removeContext(SettingsCtx ctx) = 0;
+
+    /**
      * @brief Set new activate context and set staged parameters
      * @return best match context or std::nullopt if best match context is not found in storage
      */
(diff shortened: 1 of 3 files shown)


RalphSteinhagen (approved):
Nice! Can be merged once the CI passes. :+1:

core/include/gnuradio-4.0/Settings.hpp
```diff
@@ -526,8 +560,11 @@ class CtxSettings : public SettingsBase {
 #endif
         }
 
-        const std::optional<SettingsCtx> bestMatchSettingsCtx = findBestMatchSettingsCtx(ctx);
-        if (!bestMatchSettingsCtx || bestMatchSettingsCtx == _activeCtx) {
+        std::optional<SettingsCtx> bestMatchSettingsCtx = findBestMatchSettingsCtx(ctx);
+        if (!bestMatchSettingsCtx) {
+            _storedParameters[ctx.context] = {};
+            bestMatchSettingsCtx           = ctx;
+        } else if (bestMatchSettingsCtx == _activeCtx) {
```
drslebedev, line 567: I believe this code can be removed. Since the method returns a `std::optional`, it will return `std::nullopt` if `ctx` is not found. Additionally, the default context is always present and is initialized in the `init()` function. Therefore, extra checks are unnecessary.

core/include/gnuradio-4.0/Settings.hpp
```diff
@@ -518,6 +529,29 @@ class CtxSettings : public SettingsBase {
...
+            return false;
+        }
+
+        if (_activeCtx.context == ctx.context) {
+            activateContext(); // Activate default context
+        }
+        _storedParameters.erase(ctx.context);
+
+        return true;
+    }
```
drslebedev, line 553: We should compare the context for removal with an exact match by name and time. 
Additionally, we should not allow the default context to be removed.

Example of exact match

```cpp
bool findExactMatch(const SettingsCtx& ctx) const {
    auto it = _storedParameters.find(ctx.context);
    if (it != _storedParameters.end()) {
        const auto& vec = it->second;
        return std::ranges::any_of(vec, [&ctx](const auto& pair) {
            return pair.first.time == ctx.time;
        });
    }
    return false;
}
```

core/include/gnuradio-4.0/Settings.hpp
```diff
@@ -556,6 +593,11 @@ class CtxSettings : public SettingsBase {
                 _stagedParameters.insert(parameters.value().begin(), parameters.value().end());
                 _activeCtx = bestMatchSettingsCtx.value();
                 setChanged(true);
+            } else if (ctx.context == "") {
+                // Default context was requested
+                _stagedParameters = {};
+                _activeCtx        = {};
+                setChanged(true);
```
drslebedev, line 600: I believe this code can be removed. See comment above.

drslebedev (commented):
Thank you very much for your PR.

I’ve added some comments in the code, mainly concerning the `removeContext` function. Could you please take another look and review the code?

wirew0rm:
Took a quick look at the emscripten build failures, `Error 137` usually means that the build ran out of memory. I've restarted them and I don't really see why the changes in this PR would require significantly more memory to compile.

The only thing I noticed was the fmt calls in the exception messages, we had some instances before where these were expensive in terms of compile time for something that should rarely happen and not really need to be optimized to the last cycle at compile time. On the other hand most of them look like they have simple strings as arguments to be formatted. Maybe someone more familiar with the code has a better intuition if there is innocent looking code that might hide more complexity. Or it might just have been a busy friday on the CI.

dantti:
The way to retrieve all context return 2 vectors "contexts" and "times", is it ok? if not how can I glue them so that I can serialize it?

PR #457, opened 2024-11-01
Add .pre-commit-config.yaml

dantti:
Same file from OpenDigitizer


wirew0rm (approved):
looks good to me for an optional way to allow easy setup of recommended pre-commit hooks.

PR #459, opened 2024-11-01
extended BlockModel::draw(const property_map&) API

RalphSteinhagen:
needed to pass some optional non-global UI constraints to the specific block e.g. in case the same block is redrawn in two different UI views.

### DIFF
--- a/core/include/gnuradio-4.0/BlockModel.hpp
+++ b/core/include/gnuradio-4.0/BlockModel.hpp
@@ -356,7 +356,7 @@ public:
 
     [[nodiscard]] virtual work::Result work(std::size_t requested_work) = 0;
 
-    [[nodiscard]] virtual work::Status draw() = 0;
+    [[nodiscard]] virtual work::Status draw(const property_map& config = {}) = 0;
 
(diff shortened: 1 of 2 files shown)


wirew0rm (approved):
Looks good :+1:

I assume the intended use will become clearer on coming follow-up PRs, but just to get a rough picture, the idea is to allow overriding things like line-style/color and/or passing in additional layout constraints, right?

This means the ImGui code (just assuming imgui for now, might of course be another drawing framework) now has 3 sources of information:
- the global settings of the drawable Block
- the property map passed to the draw block (this PR)
- the global ImGui context information (things like currently active style and layout constraints)

With regard to the last 2, there might be some overlap, where we have to make sure that the priorities and usage between them is consistent between different blocks and parts of the code, but anyway having a generic way to pass additional information to the draw call looks very useful for future extension.

RalphSteinhagen:
> the idea is to allow overriding things like line-style/color and/or passing in additional layout constraints, right?

Yes, also to suppressing the drawing entirely e.g. if data is supposed to be processed and stored in the internal buffers but  not to be drawn.

Issue #452, opened 2024-10-28
Automatic Tag Propagation Policies and Forwarding Policy Updates


drslebedev:
This is a follow-up issue to https://github.com/fair-acc/gnuradio4/issues/407.

## Automatic Tag Propagation Policy
- **Relevant only for blocks with `non-N_in=1:M_out=1' chunked data constraints.**
- Introduce an optional block attribute (similar to `Resampling` or `BlockingIO`) to define tag propagation policies.
- Proposed policies:
  - **Forward-propagate (default):** Treat the tag as if it belongs to the first sample of the **next** `processBulk(..)` invocation.
  - **Backward-propagate:** Treat the tag as if it belongs to the first sample of the **current** `processBulk(..)` invocation.
  - **Ignore:** Tag propagation must be handled manually by the user within the `processBulk(..)` function (see `Forwarding Policy` below).
  - **Action Items:**
    - [ ] Implement proposed above `Automatic Tag Propagation Policy`

## Forwarding Policy
- Currently, we have `all-to-all` and `one-to-one` strategies defined, but only the **all-to-all** strategy is implemented.
- (The idea of providing a customization point for users to plug in predefined or custom strategies has been abandoned.)
- Instead, tag forwarding should be managed within the custom `processBulk(..)` function using the new `InputSpan` and `OutputSpan` API.
- **Action Items:**
  - [x] Implement a way to suppress automatic tag forwarding.
  - [x] Modify the existing `Selector` block to demonstrate how to explicitly forward the correct tags to the appropriate output samples.

## Related issues
- https://github.com/fair-acc/gnuradio4/issues/444
- https://github.com/fair-acc/gnuradio4/issues/439
- https://github.com/fair-acc/gnuradio4/issues/407

Issue #369, opened 2024-06-24
DOC: Semantic of the Sync and Async Ports


drslebedev:
With this issue we want to review the semantic of `Async` ports and clearly define the difference between `Sync` and `Async` cases.

#### Number of samples constraints
There are two types of constraints used to calculate the input and output number of samples to be processed:
1. **Port Constraints**: Min/max number of samples set individually for each port.
2. **Resampling Constraints**: Numerator, denominator, and stride set for each block.

- **Port constraints** apply to both `Sync` and `Async` ports.
- **Resampling constraints** apply only to `Sync` ports.

#### `Sync` and `Async` Port Behaviors

**Sync Ports:**
- Same number of samples for all `Sync` ports.
- Fixed `M:N` ratio between input and output samples. The `processBulk` is called with `k*M` inputs and `k*N` outputs.
- Both port constraints and resampling constraints are applied.

**Async Ports:**
- The `processBulk` function is called independently of the available samples on other `Sync` and/or `Async` ports.
- Only port constraints are applied; the input/output sample ratio is undefined.
- The user can change the sample ratio iteration-to-iteration using `input.consume()` and `output.publish()`.

#### Examples

1. **4` Sync` + 1 `Async` Inputs -> 1 `Sync` Output:**
    - All 4 `Sync` inputs have consistent input (same number of samples) and `Async` has some input samples (fulfilling port constraints) -> call `processBulk`.
    - All 4 `Sync` inputs have consistent input (fulfilling port + resampling constraints), `Async` has no input (or input does not fulfill port constraints) -> call `processBulk` (`Async` port has no input samples).
    - Samples do not fulfill `Sync` input constraints, `Async` has input (fulfilling port constraints) -> call `processBulk` (no samples for input `Sync` ports, no samples for output `Sync` port, only samples for input `Async` port).
    - No `Sync` input available and no `Async` data -> do not call `processBulk`.

2. **`Sync` and/or `Async` Inputs -> 4 `Sync` + 1 `Async` Outputs:**
    - If at least one `Async` output port is present, always call `processBulk` (unless `Async` output port constraints are not met). This is because the user can call `output.publish()` even if no input samples are present. Note: This can lead to performance issues due to busy polling. `processBulk` can be invoked upto 4G/s.
    - Consider introducing a policy to prevent this and only call `processBulk` for `Async` ports if input samples are available. This can be a port policy.

#### `consume()` / `publish()` Only for `Async` Ports
[...]

PR #462, opened 2024-11-07
Manual Tag Forwarding with InputSpan and OutputSpan API

### LINKED ISSUE #444: Tags forwarding when some ports are not connected or `Aync`

drslebedev:
The changes include:
* Removed TagPropagationPolicy and introduced NoDefaultTagForwarding
* Resolved tag forwarding for scenarios where some output ports are `Async` and lack samples (Closes #444)
* Updated Selector block as a test case and example for manual tag forwarding (#452)
* Added `synch_combined_ports` parameter and possibility to have item-by-item interleaved output for multi mapped output ports
* Added tag propagation test to qa_Selector
* Added back pressure test to qa_Selector
* Added unit test in qa_Block to handle cases where input_chunk_size > 1 and input range has multiple tags
* Fix: Port::consumeTags to make untilLocalIndex exclusive
* Fix: Port::mergedTag to make untilLocalIndex exclusive
* Fix: TagMonitor to limit nSamples to the remaining samples for processing

```cpp
/**
 * @brief Disable default tag forwarding.
 *
 * There are two types of tag forwarding: (1) default All-To-All, and (2) user-implemented.
 *
 * By default, tag forwarding operates as All-To-All. Before tags on input ports are forwarded, they are merged.
 * If a block has multiple ports and tags on these ports contain maps with identical keys, only one value for each key
 * will be retained in the merged tag. This may lead to potential information loss, as it’s not guaranteed which
 * value will be kept.
 *
 * This default behavior is generally sufficient. However, if it’s not suitable for your use case, you can disable it
 * by adding the `NoDefaultTagForwarding` attribute to the template parameters. In such cases, the block should implement
 * custom tag forwarding in the `processBulk` function. The `InputSpanLike` and `OutputSpanLike` APIs are available to simplify
 * with custom tag forwarding.
 */
struct NoDefaultTagForwarding {};
```

### DIFF
--- a/blocks/basic/test/qa_Selector.cpp
+++ b/blocks/basic/test/qa_Selector.cpp
@@ -19,9 +19,13 @@ struct TestParams {
     std::vector<std::pair<gr::Size_t, gr::Size_t>> mapping;
     std::vector<std::vector<double>>               inValues;
     std::vector<std::vector<double>>               outValues;
+    std::vector<std::vector<gr::Tag>>              inTags;
+    std::vector<std::vector<gr::Tag>>              outTags;
     gr::Size_t                                     monitorSource;
     std::vector<double>                            monitorValues;
     bool                                           backPressure;
+    std::vector<gr::Size_t>                        nSamplesSelectorInput; // check back pressure
+    bool                                           syncCombinedPorts{true};
     bool                                           ignoreOrder{false};
 };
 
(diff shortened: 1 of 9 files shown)


core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1483,15 +1482,15 @@ class Block : public lifecycle::StateMachine<Derived> {
             // See https://github.com/fair-acc/gnuradio4/issues/444
```
RalphSteinhagen, line 1482: I think the publishing/forwarding of samples/tags should be not related to how many samples we consumed at the inputs.

blocks/basic/test/qa_Selector.cpp
```diff
@@ -115,140 +122,177 @@ const boost::ut::suite SelectorTest = [] {
...
+        const Tag newTag1{6, tag1.map};
+        const Tag newTag2{10, tag2.map};
+        const Tag newTag3{13, tag3.map};
         execute_selector_test({.nSamples = 5,                                                       //
             .mapping                     = {{0, 1}, {1, 1}, {2, 1}},                                //
             .inValues                    = {{1}, {2}, {3}},                                         //
-            .outValues                   = {{}, {1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3}, {}}, //
-            .monitorSource               = -1U,                                                     //
-            .monitorValues               = {},                                                      //
+            .outValues                   = {{}, {1, 2, 2, 3, 3, 3, 1, 1, 1, 1, 2, 2, 2, 3, 3}, {}}, //
```
RalphSteinhagen, line 164: The default scheme should add a
```cpp
 Annotated<bool, "synch combined port", Doc<"true: input ports connected to the same output port are synchronised">> synch_combined_ports = true;
```
and produce
```cpp
.outValues                   = {{}, {1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3}, {}}, //
```

i.e. having an item-by-item interleaved output.

blocks/basic/test/qa_Selector.cpp
```diff
@@ -115,140 +122,177 @@ const boost::ut::suite SelectorTest = [] {
...
             .mapping                     = {{1, 1}},                  //
             .inValues                    = {{1}, {2}, {3}},           //
             .outValues                   = {{}, {2, 2, 2, 2, 2}, {}}, //
-            .monitorSource               = -1U,                       //
-            .monitorValues               = {},                        //
+            .inTags                      = {{tag1}, {tag2}, {tag3}},  //
+            .outTags                     = {{}, {tag2}, {}},
+            .monitorSource               = -1U, //
+            .monitorValues               = {},  //
             .backPressure                = true,
```
RalphSteinhagen, line 210: please check remaining samples in the unconnected ports, there should be '5' in this case. Similar for the cases below.

core/test/qa_Block.cpp
```diff
@@ -720,6 +720,40 @@ const boost::ut::suite<"Stride Tests"> _stride_tests = [] {
...
+        expect(eq(gr::ConnectionResult::SUCCESS, testGraph.connect<"out">(intDecBlock).to<"in">(sink)));
+
+        gr::scheduler::Simple sched{std::move(testGraph)};
+        expect(sched.runAndWait().has_value());
+
+        expect(eq(intDecBlock.status.process_counter, 2UZ));
+        expect(eq(intDecBlock.status.total_in, 20UZ));
+        expect(eq(intDecBlock.status.total_out, 20UZ));
+
+        // TODO: we need to check the output tags, but the proper policy should be first implemented
```
RalphSteinhagen, line 752: remove commented code and 'TODO'

RalphSteinhagen (approved):
Thanks. Looks good to me. Can be merged without further review once the above comments are addressed. :+1:

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1483,15 +1482,15 @@ class Block : public lifecycle::StateMachine<Derived> {
...
+                    if (in.isConsumeRequested() && in.isConnected && in.isSync) {
                         processedIn = std::min(processedIn, in.nRequestedSamplesToConsume());
                     }
                 },
                 inputSpans);
 
             for_each_writer_span(
                 [&processedOut](auto& out) {
-                    if (out.isPublishRequested() && out.isConnected) {
+                    if (out.isPublishRequested() && out.isConnected && out.isSync) {
```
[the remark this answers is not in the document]
drslebedev: The current logic requires this behavior; otherwise, if the output async port has no samples, tags won’t be forwarded across all ports.

We still need to address certain sync-async port semantics, as highlighted in issues like [#369](https://github.com/fair-acc/gnuradio4/issues/369) and [#444](https://github.com/fair-acc/gnuradio4/issues/444).

PR #465, opened 2024-11-19
Implement SyncBlock

drslebedev:
- This PR introduces a `SyncBlock` to synchronize data streams across multiple inputs that have the same sample rate. 
- It addresses key scenarios including initial time shifts, clock drift, and significant delays. 
- The `SyncBlock` synchronises input samples based on synchronization tags and publishes desynchronization tags with information about number of samples that were dropped.
- Add `bm_sync` for performance comparison

```
┌──────────────benchmark:───────────────┬──────┬──CPU branch misses───┬─ops/s─┬──mean──┬─<CPU-I>─┬───CPU cache misses───┬─stddev─┬─#N─┬─CTX-SW─┬─total time─┬─min─┬─median─┬─max─┐
│ src->gr::basic::SyncBlock<int>->sink  │ PASS │ 3.10M / 382M =  0.8% │  4.7G │ 843 ms │    930m │ 494M / 1.77G = 27.9% │        │    │     1  │     843 ms │     │        │     │
│ src->gr::basic::Selector<int>->sink   │ PASS │  394k / 363M =  0.1% │  4.7G │ 851 ms │    907m │ 409M / 1.86G = 22.0% │        │    │     2  │     851 ms │     │        │     │
│ src->copy->sink                       │ PASS │  403k / 125M =  0.3% │  4.8G │ 835 ms │    651m │ 337M / 1.94G = 17.3% │        │    │     2  │     835 ms │     │        │     │
└───────────────────────────────────────┴──────┴──────────────────────┴───────┴────────┴─────────┴──────────────────────┴────────┴────┴────────┴────────────┴─────┴────────┴─────┘
```

Related follow-up issue: #466

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/SyncBlock.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/SyncBlock.hpp
@@ -1,0 +1,363 @@
+#ifndef GNURADIO_SYNC_BLOCK_HPP
+#define GNURADIO_SYNC_BLOCK_HPP
+
+#include <gnuradio-4.0/Block.hpp>
+#include <gnuradio-4.0/HistoryBuffer.hpp>
+
+namespace gr::basic {
+
+using namespace gr;
+
+using SyncBlockDoc = Doc<R""(
+
+The SyncBlock synchronizes data streams across multiple inputs.
+It addresses key scenarios including initial time shifts, clock drift, and significant delays.
+The SyncBlock synchronises input samples based on synchronization tags and
+publishes desynchronization tags with information about number of samples that were dropped.
+
+### Important Prerequisites
+The sample rate must be the same across all input ports.
+
+For the examples provided, consider the following:
+1) Samples (`s1, s2, ...`) represent samples where the index indicates their sequence of arrival.
+2) Samples for different ports may vary (`s1` for `port1` may may not be equal to `s1` for `port2`)
+3) Illustrated with examples for two inputs (`in1`, `in2`) and two outputs (`out1`, `out2`) for simplicity:
+```text
+       +-------------------------------+
+in1 +-+|                               |+-+ out1
+    +-+|                               |+-+
+       |          SyncBlock            |
+in2 +-+|                               |+-+ out2
+    +-+|                               |+-+
+       +-------------------------------+
+```
+
+### Implementing `SyncBlock`: Key Scenarios
+When designing a `SyncBlock`, consider the following four scenarios:
+
+#### 1. Time Shift at Start
+Devices may initiate data transmission at different times, leading to a time shift in data arrival.
+
+Example:
+```text
+                 T1                         T1,TD1
+                  ↓                            ↓
+in1: s2-s3-s4-s5-s6      →  out1: s2-s3-s4-s5-s6
+
+                    T1                        T1
+                     ↓                         ↓
+in2: s1-s2-s3-s4-s5-s6   →  out2: s2-s3-s4-s5-s6
+```
+Both ports align with synchronization tag `T1` at sample `s6`.
+Samples received before `T1` are also included in the output, resulting in `out1` and `out2` both outputting `s2-s3-s4-s5-s6`.
+Note that the synchronization tags are forwarded further.
+In addition a desync tag (`TD1`) is published for `out1` which includes information about number of dropped samples.
+
+#### 2. Clock Drift
+This scenario accounts for a clock drift - minor timing discrepancy, such as several PPM (±1 sample per 100,000 samples).
+
+Example:
+```text
+     T1          T2               T1          T2
+      ↓           ↓                ↓           ↓
+in1: s1-s2-s3-s4-s5      →  out1: s1-s2-s3-s4-s5
+
+     T1             T2            T1         T2,TD2
+      ↓              ↓             ↓           ↓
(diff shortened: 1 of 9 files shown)


blocks/basic/include/gnuradio-4.0/basic/SyncBlock.hpp
```diff
@@ -0,0 +1,363 @@
...
+            fmt::print("{}: configuration changed: n_ports {} -> {}\n", this->name, oldSettings.at("n_ports"), newSettings.at("n_ports"));
+            inputs.resize(n_ports);
+            outputs.resize(n_ports);
+            _nDroppedSamples.resize(n_ports, 0UZ);
+        }
+
+        if (newSettings.contains("max_history_size")) {
+            // should be less than actual buffer size (better < 80%)
+            // The logic has to be implemented when the new Port API is available
+        }
```
RalphSteinhagen, line 159: As discussed in person: the block implementation should not assume nor use the Port's buffer API. 

I think it's OK to define a `max_history_size` that is used to define/communicate the required minimum buffer size to the graph and scheduler, and in turn be used implicitly by the block as the buffer having at least `max_history_size` slots.

However, the block should not depend on the size of the actually used Port-buffer. To note, we want to keep the buffer API slim and also connect blocks using other mechanisms than via buffers (<-> merge API).

blocks/basic/include/gnuradio-4.0/basic/SyncBlock.hpp
```diff
@@ -0,0 +1,363 @@
...
+            if (!ins[i].isConsumeRequested()) {
+                ins[i].consume(0UZ);
+                ins[i].consumeTags(0UZ);
+                outs[i].publish(0UZ);
+            }
+        }
+        return gr::work::Status::OK;
+    }
+
+private:
```
RalphSteinhagen, line 232: Consider removing the `private:` because having the functions below (most could be actually free-standing functions) public makes them easier to unit-test.

blocks/basic/include/gnuradio-4.0/basic/SyncBlock.hpp
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    template<InputSpanLike TInput>
+    [[nodiscard]] constexpr std::vector<SyncData> synchronize(const std::span<TInput>& ins) {
+        const std::uint64_t syncTime = findSyncTime(ins);
+        if (syncTime == std::numeric_limits<std::uint64_t>::max()) {
+            return {};
+        }
+
+        std::vector<SyncData> syncData;
```
RalphSteinhagen, line 255: At the risk that this might be a PMO: maybe we should consider making
```cpp
std::vector<SyncData> _syncData;
```
since this is called for every invocation of  `processBulk(..)? similarly for the `findSyncTime(...)` routine below. 

This may be likely a PMO but I am a bit concerned about two vectors being allocated everytime the block is invoked even if there are no synchronisation necessary. Maybe a basic benchmark (w/ and one w/o this block but a simple copy) that measures how much performance is lost might be useful. This way we could eliminate also future doubts and potential regressions.

In terms of benchmarks... guess we have plenty of `bm_,,,` to choose and copy-paste from. :-)

blocks/basic/include/gnuradio-4.0/basic/SyncBlock.hpp
```diff
@@ -0,0 +1,363 @@
...
+            }
+        }
+        return in.size();
+    }
+
+    [[nodiscard]] bool isTimeDifferenceWithinTolerance(std::uint64_t t1, std::uint64_t t2) { return ((t1 > t2) ? t1 - t2 : t2 - t1) < tolerance; }
+
+    [[nodiscard]] constexpr bool isSyncTag(const gr::Tag& tag) const {
+        const std::string keyTriggerName = gr::tag::TRIGGER_NAME.shortKey();
+        const std::string keyTriggerTime = gr::tag::TRIGGER_TIME.shortKey();
```
RalphSteinhagen, line 339: Why are these needed as a `std::string`? Doesn't the `at(..)` below work with string_views`

RalphSteinhagen (approved):
Thanks @drslebedev for this PR. This looks like a thorough, neat and concise implementation. I like it.

I think this could be already merged as-is but could benefit from some (optional) improvements outlined in the comments above.

Notably:
 1. the time-out and zero-padding isn't implemented, which is OK for now but should be marked by a `TODO` and followed up with a dedicated issue (we should discuss and meet for discussing the details since this is a wider issue/feature requirements)
  2. adding a benchmark to assess the cost of performing the synchronisation on streams that are already synchronised. This would help a lot when further refining the implementation and/or to detect potential performance regressions.
  
In any case: good job and well done! :+1:

blocks/basic/include/gnuradio-4.0/basic/SyncBlock.hpp
drslebedev: The idea here is to check if max_history_size is actually < buffer size, if not throw an gr::exception.

blocks/basic/include/gnuradio-4.0/basic/SyncBlock.hpp
drslebedev: `std::map::at` will not work with `std::sting_view`

PR #467, opened 2024-11-21
Bump Emscripten version to 3.1.66

drslebedev:
The Emscripten issue, likely caused by resource exhaustion, was introduced in [PR #456](https://github.com/fair-acc/gnuradio4/pull/456). 

This PR fixes the problem by bumping to the new Emscripten version and removing benchmark tests from compilation.

The `qa_Message` test failed due to the new tests added in [PR #456](https://github.com/fair-acc/gnuradio4/pull/456). This issue went unnoticed because the same PR also introduced the Emscripten resource exhaustion problem.

### DIFF
--- a/core/benchmarks/CMakeLists.txt
+++ b/core/benchmarks/CMakeLists.txt
@@ -1,14 +1,24 @@
-function(add_gr_benchmark BM_NAME)
+if(NOT EMSCRIPTEN)
+  function(add_gr_benchmark BM_NAME)
     add_benchmark(${BM_NAME})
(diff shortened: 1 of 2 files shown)


wirew0rm (approved):
Looks, good, thanks for following this up.

I think it's worthwhile to directly merge this even if the emscripten build still fails, but at least this now shows which of the tests are failing so on new PRs we can make sure that we don't get further emscripten specific test failures. Nevertheless, the failing tests should be fixed rather soon.

Can be merged after applying the reformatting.

PR #468, opened 2024-11-21
Implement YAML <-> pmtv::map_t generator and parser

frankosterfeld:
This implements serialization and deserialization of pmtv::map_t to a subset YAML 1.2.

It doesn't support the full YAML 1.2 standard, but rather focuses to serialize any possible pmtv::map_t value to valid YAML and back.

### DIFF
--- a/core/test/qa_YamlPmt.cpp
+++ b/core/test/qa_YamlPmt.cpp
@@ -133,8 +133,9 @@ void testYAML(std::string_view src, const pmtv::map_t expected) {
 }
 
 const boost::ut::suite YamlPmtTests = [] {
-#pragma GCC diagnostic push
+#ifndef __clang__
 #pragma GCC diagnostic ignored "-Wuseless-cast" // we want explicit casts for testing
+#endif
     using namespace boost::ut;
     using namespace pmtv;
     using namespace std::string_literals;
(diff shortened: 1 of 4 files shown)


jsallay:
By missing "maps" do you mean nested maps - since the whole file is a map?  I don't know that it is needed up front, but I think that it will be needed eventually.  In GR 3.0 there are a handful of common structs that are passed around kind of as object pointers.  They would be better expressed as simple yaml dictionaries.

Other than that it looks pretty good to me.

frankosterfeld:
> By missing "maps" do you mean nested maps - since the whole file is a map? I don't know that it is needed up front, but I think that it will be needed eventually. In GR 3.0 there are a handful of common structs that are passed around kind of as object pointers. They would be better expressed as simple yaml dictionaries.

It should have read "{flow-style} x {lists, maps}", focus on "flow-style" (`[1, 2, 3], {foo: 1, bar: 2}`, etc. I've added that by now, the description was out-of-date.

So if I don't miss anything, any possible pmtv::map_t value should now be able to be serialised to YAML and back, while in the other direction, not all of YAML is supported (unicode escapes I left out for now, assuming those can be encoded as UTF-8, there are more ways to encode multi-line strings, and there are some valid but weird ways to spread e.g. `key: scalar` and `key: [...]` over multiple lines that the parser currently rejects)

core/test/qa_YamlPmt.cpp
```diff
@@ -0,0 +1,722 @@
...
+#include <fmt/format.h>
+#include <fmt/ranges.h>
+#include <fmt/std.h>
+
+#include <gnuradio-4.0/YamlPmt.hpp>
+#include <limits>
+#include <variant>
+
+template<typename T>
+std::string_view typeName() {
```
RalphSteinhagen, line 14: We have a `gr::meta::type_name<T>()` definition that would provide a non-mangled name.

RalphSteinhagen (approved):
@frankosterfeld nicely done and to the point of what is needed and agreed upon limited functionality for the GR4 internal use i.e. read-write identity, sending `property_map`s from/to remote sources, replaying the GRC yaml loading ... :+1: 

The unit-tests are comprehensive and -- if we need to expand upon this implementation -- a good starting point.

jsallay (approved):
LGTM

PR #474, opened 2024-11-29
Return a message when a Settings Context is removed

dantti:
Also always return current active context time when set/get


dantti:
it's finished but requires https://github.com/fair-acc/gnuradio4/pull/474 to be merged

RalphSteinhagen (approved):

PR #475, opened 2024-12-02
DataSink updates and bug fixing

drslebedev:
This pull request introduces the following updates and bug fixes to the DataSink block:

* Implemented `BaseListener` for `DataSetPoller`: Refactored the code to introduce a BaseListener class for DataSetPoller. This change reduces code duplication across listener implementations.
* Switched to blocking `reserve()` for DataSetPoller.

Notes:

We should consider revisiting the DataSink block, especially concerning the handling and terminology of the blocking mode:
* The current terminology for the blocking mode can be misleading. When set to `NonBlocking`, samples are dropped if there aren’t enough slots available for writing. This behavior might not be apparent to users, potentially leading to confusion. Consider renaming to  `enable_sample_dropping`.
* Blocking Behavior Impact: The current implementation uses a blocking call to `Writer::reserve()`, which can result in all flowgraph threads being completely blocked if the output buffer doesn’t have enough space for writing. This can cause significant bottlenecks, particularly when `DataSink` consumers process data slower than the producers generate it.
* The desired behavior could be achieved using non-blocking `tryReserve` calls combined with backpressure mechanisms. This method would allow the scheduler to become aware when `DataSink` consumers are lagging behind, enabling it to adjust the data rate accordingly.

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
@@ -25,7 +25,6 @@ template<typename T>
 class DataSetSink;
 
 // Until clang-format can handle concepts
-// clang-format off
 
 template<typename T, typename V>
 concept DataSetCallback = std::invocable<T, DataSet<V>>;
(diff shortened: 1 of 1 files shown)


wirew0rm (approved):
Looks good as a limited scope improvement for the DataSink. Thanks for the improvements and additional comments and documentation.

PR #480, opened 2024-12-05
PmtYaml: Fix parsing of colons

frankosterfeld:
Don't interpret ":" as key/value separator when followed by a non-whitespace character.


RalphSteinhagen (approved):

dantti (approved):
Works for me

PR #482, opened 2024-12-06
CI: Cancel existing builds when PR branch is updated

frankosterfeld:


wirew0rm (approved):
This kind of got forgotten before christmas, but it looks like it is working very nicely and will help to not waste unnecessary CI cycles.

Thanks for following this up and coming up with a very concise solution that will save quite a bit of manual CI interactions whenever we run out of CI runners

PR #484, opened 2024-12-09
updated build system

RalphSteinhagen:
* set relative path for exprtk
* set optimisation flags to max '-O2' to alleviate CI build times
* set absolute hash for pmt dependency

_N.B. performance difference between `-O2` vs. '-03' proved to be insignificant. ExprTk has about 30% ops/s compared to the natively executed function expression. See chart below for details:_

gr4_exptk_perf


RalphSteinhagen:
shaved off ~30" especially from gcc14-release build:

before:
image

after:
image

wirew0rm (approved):
Looks good :+1:

PR #485, opened 2024-12-09
cmake: use PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR

daniestevez:
Use PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR, since when gnuradio4 is built as a dependency using FetchContent_Declare(), CMAKE_SOURCE_DIR points to the source of the application that depends on gnuradio4 instead of gnuradio4's source.


daniestevez:
Just seen this change is included as part of https://github.com/fair-acc/gnuradio4/pull/484, so disregard this PR. Closing.

PR #489, opened 2024-12-18
ClockSource T -> std::uint8_t fix

RalphSteinhagen:
as discussed: changed ClockSource output port to always return 'std::uint8_t' as default data type for timing-only, clock-type and other usages where the vertical signal value is only of secondary importance.


wirew0rm (approved):
Looks good :+1:  can be merged once the CI passed.

PR #490, opened 2024-12-24
XMas-Day-1: HistoryBuffer: added push_front method, iterator constraints, and resize

RalphSteinhagen:
## On the first day of Christmas :christmas_tree: :angel: :one: :christmas_tree: , my code base gave to me...

*A ring buffer with push_front and pop—how jolly can that be?*

This PR **refines** the `HistoryBuffer` by introducing or improving:

1. **\`push_front(...)\`**: Lets you store elements so the *oldest* ends up at index \[0\].  
2. **\`pop_front()\` & \`pop_back()\`**: Remove elements from the logical front/back of the ring.  
3. **\`front()\` & \`back()\`**: Simple, direct access to the first/last elements in logical order.  
4. **\`resize(size_t)\`** (for dynamic-extent): Resize capacity while retaining existing items.  
5. **Documentation**: We’ve updated our doxygen comments to show how these new methods work, plus concise code examples illustrating “newest-at-\[0\]” vs. “oldest-at-\[0\].”

Along the way, we trimmed some unneeded modulo operations for single-step increments, simplified loops with `std::copy`, and tidied up our tests.

Feel free to test it out and give feedback.  
**Happy coding—and happy holidays!**

**Code Example**:
```cpp
gr::history_buffer<int, 5> hb_newest;   // Use push_back -> index[0] is newest
hb_newest.push_back(1); // [1]
hb_newest.push_back(2); // [2, 1]
hb_newest.push_back(3); // [3, 2, 1]
// => index[0] == 3, newest item

gr::history_buffer<int, 5> hb_oldest;   // Use push_front -> index[0] is oldest
hb_oldest.push_front(10); // [10]
hb_oldest.push_front(20); // [10, 20]
hb_oldest.push_front(30); // [10, 20, 30]
// => index[0] == 10, oldest item

hb_newest.pop_front();  // remove newest => now hb_newest: [2, 1]
hb_oldest.pop_front();  // remove oldest => now hb_oldest: [20, 30]

auto val = hb_newest.front();  // val == 2
auto last = hb_newest.back();  // last == 1
```

### DIFF
--- a/core/include/gnuradio-4.0/HistoryBuffer.hpp
+++ b/core/include/gnuradio-4.0/HistoryBuffer.hpp
@@ -15,31 +15,44 @@
 namespace gr {
 
 /**
- * @brief A simple circular history buffer that supports contiguous ranged access across the wrap-around point.
+ * @brief A single-threaded circular history buffer with a double-mapped array,
+ * allowing either newest-at-[0] (`push_back`) or oldest-at-[0] (`push_front`) usage.
  *
- * This buffer is similar to the `circular_buffer<T>` and uses a double-mapped memory approach.
- * It is optimised for single-threaded use and does not provide thread-safety or multi-writer/multi-reader support.
+ * ### Storage
+ * - Dynamic (`N == std::dynamic_extent`): uses `std::vector<T>` (size = 2 * capacity).
+ * - Fixed (`N != std::dynamic_extent`): uses `std::array<T, N * 2>`.
  *
- * Example usage:
- * gr::history_buffer<int, 5> buffer;
- * buffer.push_back(1);  // buffer: [1] (size: 1, capacity: 5)
- * buffer.push_back(2);  // buffer: [2, 1] (size: 2, capacity: 5)
- * buffer.push_back(3);  // buffer: [3, 2, 1] (size: 3, capacity: 5)
- * buffer.push_back(4);  // buffer: [4, 3, 2, 1] (size: 4, capacity: 5)
- * buffer.push_back(5);  // buffer: [5, 4, 3, 2, 1] (size: 5, capacity: 5)
- * buffer.push_back(6);  // buffer: [6, 5, 4, 3, 2] (size: 5, capacity: 5)
+ * ### Key Operations
+ * - `push_back(const T&)`: Add new item, drop oldest if full, index `[0]` is newest.
+ * - `push_front(const T&)`: Add new item, drop oldest if full, index `[0]` is oldest.
+ * - `pop_front()`, `pop_back()`: Remove from logical front/back of the ring.
+ * - `front()`, `back()`: Returns the first/last item in logical order.
+ * - `operator[](i)`, `at(i)`: Unchecked/checked access.
+ * - `resize(...)` (dynamic-only): Adjust capacity, preserving existing data.
+ * - `get_span(...)`: Obtain a contiguous view across wrap boundaries.
  *
- * // :
- * buffer[0];     // value: 6 - unchecked access of last/actual sample
- * buffer[1];     // value: 5 - unchecked access of previous sample
- * ...
- * buffer.at(0);  // value: 6 - checked access of last/actual sample
- * buffer.at(1); // value: 5 - checked access of last/actual sample
- * ...
- * buffer.get_span(0, 3);  // span: [6, 5, 4]
- * buffer.get_span(1, 3);  // span: [6, 3]
- * buffer.get_span(0);     // span: [6, 5, 4, 3, 2]
- * buffer.get_span(1);     // span: [5, 4, 3, 2s]
+ * ### Examples
+ * \code{.cpp}
+ * gr::history_buffer<int, 5> hb_newest;   // Use push_back -> index[0] is newest
+ * hb_newest.push_back(1); // [1]
+ * hb_newest.push_back(2); // [2, 1]
+ * hb_newest.push_back(3); // [3, 2, 1]
+ * // => index[0] == 3, newest item
+ *
+ * gr::history_buffer<int, 5> hb_oldest;   // Use push_front -> index[0] is oldest
+ * hb_oldest.push_front(10); // [10]
+ * hb_oldest.push_front(20); // [10, 20]
+ * hb_oldest.push_front(30); // [10, 20, 30]
+ * // => index[0] == 10, oldest item
+ *
+ * hb_newest.pop_front();  // remove newest => now hb_newest: [2, 1]
+ * hb_oldest.pop_front();  // remove oldest => now hb_oldest: [20, 30]
+ *
+ * auto val = hb_newest.front();  // val == 2
+ * auto last = hb_newest.back();  // last == 1
+ * \endcode
+ *
+ * This class is not thread-safe. For concurrency, use external synchronization.
  */
 template<typename T, std::size_t N = std::dynamic_extent, typename Allocator = std::allocator<T>>
 class HistoryBuffer {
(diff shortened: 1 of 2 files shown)


core/include/gnuradio-4.0/HistoryBuffer.hpp
```diff
@@ -100,23 +113,176 @@ class HistoryBuffer {
...
+    template<std::input_iterator Iter>
+    constexpr void push_back(Iter cbegin, Iter cend) noexcept {
         const auto nSamplesToCopy = std::distance(cbegin, cend);
         Iter       optimizedBegin = static_cast<std::size_t>(nSamplesToCopy) > _capacity ? std::prev(cend, static_cast<std::ptrdiff_t>(_capacity)) : cbegin;
         for (auto it = optimizedBegin; it != cend; ++it) {
             push_back(*it);
         }
     }
 
+    template<std::input_iterator Iter>
```
ivan-cukic, line 125: Why a separate name for the new push_back?

IMO, it should be named `append_range` as in `std::vector`, though that one takes a range and not a pair of iterators. We could have `append_range(begin, end)` and `append_range(range)` -- if the functions below also become `append_range`?

core/include/gnuradio-4.0/HistoryBuffer.hpp
```diff
@@ -100,23 +113,176 @@ class HistoryBuffer {
...
+        push_back_bulk(range.cbegin(), range.cend());
+    }
+
+    template<std::ranges::range Range>
     constexpr void push_back_bulk(const Range& range) noexcept {
         push_back_bulk(range.cbegin(), range.cend());
     }
 
+    /**
+     * @brief Adds an element to the front expiring the oldest element beyond the buffer's capacities.
```
ivan-cukic, line 144: Not sure what this means.

core/include/gnuradio-4.0/HistoryBuffer.hpp
```diff
@@ -100,23 +113,176 @@ class HistoryBuffer {
...
+        _buffer[insertPos]             = value;
+        _buffer[insertPos + _capacity] = value;
+    }
+
+    /**
+     * @brief Inserts a contiguous range [first, last) so that the new items become
+     *        the "newest" in the ring, with operator[](0) always the oldest.
+     */
+    template<std::input_iterator Iter>
+    constexpr void push_front(Iter first, Iter last) noexcept {
```
ivan-cukic, line 169: Following the previous comment, this would be `prepend_range`.

core/include/gnuradio-4.0/HistoryBuffer.hpp
```diff
@@ -100,23 +113,176 @@ class HistoryBuffer {
...
+        return push_back(cbegin, cend);
+    }
+
     /**
      * @brief Adds a range of elements the end expiring the oldest elements beyond the buffer's capacities.
      */
-    template<typename Range>
+    template<std::ranges::range Range>
+    constexpr void push_back(const Range& range) noexcept {
+        push_back_bulk(range.cbegin(), range.cend());
```
ivan-cukic, line 135: `std::ranges::begin(range)`, same for end.

core/include/gnuradio-4.0/HistoryBuffer.hpp
```diff
@@ -100,23 +113,176 @@ class HistoryBuffer {
...
+        }
+    }
+
+    /**
+     * @brief Adds a range of elements the end expiring the oldest element so that the new items become
+     *        the "newest" in the ring, with operator[](0) always the oldest.
+     */
+    template<std::ranges::range Range>
+    constexpr void push_front(const Range& r) noexcept {
+        push_front(std::begin(r), std::end(r));
```
ivan-cukic, line 216: `std::ranges::begin(range)`, same for end.

core/include/gnuradio-4.0/HistoryBuffer.hpp
```diff
@@ -100,23 +113,176 @@ class HistoryBuffer {
...
+
+        std::vector<T, Allocator> newBuf(newCapacity * 2);
+
+        const std::size_t copyCount = std::min(_size, newCapacity);
+        const auto        oldFirst  = cbegin();
+        std::copy(oldFirst, oldFirst + static_cast<std::ptrdiff_t>(copyCount), newBuf.begin());                                                        // copy first half
+        std::copy(newBuf.begin(), newBuf.begin() + static_cast<std::ptrdiff_t>(copyCount), newBuf.begin() + static_cast<std::ptrdiff_t>(newCapacity)); // mirror second half
+
+        // update members
+        _buffer         = std::move(newBuf); // destroys old buffer
```
ivan-cukic, line 280: `std::swap` might be cheaper here, as newBuf (which would then contain old _buffer data) is anyhow destroyed on end-of-scope.

ivan-cukic (changes requested):

core/include/gnuradio-4.0/HistoryBuffer.hpp
RalphSteinhagen: I didn't want to break existing code, thus keeping the "old" `push_back_bulk(...)`. From a templated API point of view there is no need to distinguish to distinguish between single sample or bulk operation. The implementation is different (for efficiency reasons) but the API should stay the same. 

`push_back` and `push_front' keep the symmetry between reverse (nominal) and forward (new optional) behaviour.

core/include/gnuradio-4.0/HistoryBuffer.hpp
RalphSteinhagen: I went for the [std::deque<T>](https://en.cppreference.com/w/cpp/container/deque) API syntax with some sprinkles of `std::vector<T>`. Hope this make sense. Didn't want to break too much of the existing API/use elsewhere.

:-)

core/include/gnuradio-4.0/HistoryBuffer.hpp
RalphSteinhagen: It's still a circular/limited buffer size. If you add a sample (either front or back) the oldest is being dropped once the maximum size is reached. This is the difference to the `std::deque` behaviour.

ivan-cukic: No, I get that, just wanted to say the comment itself is not that clear :)

core/include/gnuradio-4.0/HistoryBuffer.hpp
ivan-cukic: `std::deque` also has single-value push_back/push_front and append_range/prepend_range for multiple values.

core/include/gnuradio-4.0/HistoryBuffer.hpp
ivan-cukic: > I didn't want to break existing code, thus keeping the "old" push_back_bulk(...).

I was hoping we're still in the 'we can break API things' as we didn't have a release. (and we do break behaviour in other places from time to time)

RalphSteinhagen: Generally yes, but didn't want to change the whole code base and do the follow-up in gr-digitizer and OpenDigitizer ... yet.

core/include/gnuradio-4.0/HistoryBuffer.hpp
RalphSteinhagen: Didn't know about 'append_range/prepend_range' yet (it's C++23 newish). Shall we add this? I like the overloaded `push_back/push_front` alone because semantically there isn't much difference other than that the `..._range` also describe the type of function argument.

core/include/gnuradio-4.0/HistoryBuffer.hpp
```diff
@@ -15,31 +15,44 @@
...
+ * hb_oldest.push_front(10); // [10]
+ * hb_oldest.push_front(20); // [10, 20]
+ * hb_oldest.push_front(30); // [10, 20, 30]
+ * // => index[0] == 10, oldest item
+ *
+ * hb_newest.pop_front();  // remove newest => now hb_newest: [2, 1]
+ * hb_oldest.pop_front();  // remove oldest => now hb_oldest: [20, 30]
+ *
+ * auto val = hb_newest.front();  // val == 2
+ * auto last = hb_newest.back();  // last == 1
```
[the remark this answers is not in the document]
wirew0rm: discussed offline, this will be fixed in a followup change that will change this in all affected repos at once, together with removing unused functions.

PR #496, opened 2025-01-15
HistoryBuffer<T> flipped push_back<->push_front naming

RalphSteinhagen:
should (hopefully) simplify and provide a more intuitive the naming.

### Storage
 - Dynamic (`N == std::dynamic_extent`): uses `std::vector<T>` (size = 2 * capacity).
 - Fixed (`N != std::dynamic_extent`): uses `std::array<T, N * 2>`.

### Key Operations
 - `push_front(const T&)`: Add new item, drop oldest if full, index `[0]` is newest.
 - `push_back(const T&)`: Add new item, drop oldest if full, index `[0]` is oldest.
 - `pop_front()`, `pop_back()`: Remove from logical front/back of the ring.
 - `front()`, `back()`: Returns the first/last item in logical order.
 - `operator[](i)`, `at(i)`: Unchecked/checked access.
 - `resize(...)` (dynamic-only): Adjust capacity, preserving existing data.
 - `get_span(...)`: Obtain a contiguous view across wrap boundaries.

 ### Code Examples
 ```cpp
 gr::history_buffer<int, 5> hb_newest;   // use push_front -> index[0] is newest
 hb_newest.push_front(1); // [1]
 hb_newest.push_front(2); // [2, 1]
 hb_newest.push_front(3); // [3, 2, 1]
 // => index[0] == 3, newest item

gr::history_buffer<int, 5> hb_oldest;   // use push_back -> index[0] is oldest
hb_oldest.push_back(10); // [10]
hb_oldest.push_back(20); // [10, 20]
hb_oldest.push_back(30); // [10, 20, 30]
// => index[0] == 10, oldest item

hb_newest.pop_front();  // remove newest => now hb_newest: [2, 1]
hb_oldest.pop_front();  // remove oldest => now hb_oldest: [20, 30]

auto val = hb_newest.front();  // val == 2
auto last = hb_newest.back();  // last == 1
 ```

 Also added/extended the `HistoryBuffer<T>` performance benchmark:
 
image

*N.B. the performance for `push_front(..)` -- used by IIR, FIR filter etc. storing the newest sample in the first (`0`) index -- is slightly asymmetric and ~30% faster w.r.t. `push_back(..)` which stores the newest sample in the last `size() - 1UZ` index.*

### DIFF
--- a/algorithm/include/gnuradio-4.0/algorithm/SchmittTrigger.hpp
+++ b/algorithm/include/gnuradio-4.0/algorithm/SchmittTrigger.hpp
@@ -103,7 +103,7 @@ struct SchmittTrigger {
         }
 
         if constexpr (Method == BASIC_LINEAR_INTERPOLATION) {
-            _historyBuffer.push_back(input);
+            _historyBuffer.push_front(input);
             if (_historyBuffer.size() < 2) {
                 return EdgeDetection::NONE;
             }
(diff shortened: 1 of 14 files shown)


wirew0rm (approved):
Looks good and will make usage more consistent. Thanks for following this through!
Will merge once the last Job has run through.

PR #502, opened 2025-02-05
Updates in DataSetSink

drslebedev:
- Sets `DataSet` output as the default for the `StreamToDataSet` block.
- Updates `DataSetSink` to process the entire `DataSet` based on `signal_name`.

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
@@ -24,8 +24,6 @@ class DataSink;
 template<typename T>
 class DataSetSink;
 
-// Until clang-format can handle concepts
-
 template<typename T, typename V>
 concept DataSetCallback = std::invocable<T, DataSet<V>>;
 
@@ -37,6 +35,12 @@ concept StreamCallback = std::invocable<T, std::span<const V>> || std::invocable
 
 template<typename T, typename V>
 concept DataSetMatcher = std::invocable<T, gr::DataSet<V>> && std::is_same_v<std::invoke_result_t<T, gr::DataSet<V>>, bool>;
+
+template<typename T>
+concept DataSinkOrDataSetSinkLike = gr::meta::is_instantiation_of<T, DataSink> || gr::meta::is_instantiation_of<T, DataSetSink>;
+
+static_assert(DataSinkOrDataSetSinkLike<DataSink<float>>);
+static_assert(DataSinkOrDataSetSinkLike<DataSetSink<float>>);
 
 namespace detail {
 constexpr std::size_t data_sink_buffer_size          = 65536;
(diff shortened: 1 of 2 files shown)


blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
```diff
@@ -128,68 +132,50 @@ class DataSinkRegistry {
         return s_instance;
     }
 
-    template<typename T>
-    void registerSink(DataSink<T>* sink) {
+    template<DataSinkOrDataSetSinkLike TSink>
+    void registerSink(TSink* sink) {
         std::lock_guard lg{_mutex};
+
+        if (sink->signal_name == "") {
```
frankosterfeld, line 139: Or `sink->signal_name.empty()`

frankosterfeld (commented):
LGTM.

Is it clear how GR worker subscriptions (in opendigitizer) should work with this? subscribing to a whole dataset vs. the payload there (`Acquisition`) only supporting one dimension.

blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
```diff
@@ -38,6 +36,12 @@ concept StreamCallback = std::invocable<T, std::span<const V>> || std::invocable
 template<typename T, typename V>
 concept DataSetMatcher = std::invocable<T, gr::DataSet<V>> && std::is_same_v<std::invoke_result_t<T, gr::DataSet<V>>, bool>;
 
+template<typename T>
+concept DataSinkOrDataSetSinkLike = gr::meta::is_instantiation_of<T, DataSink> || gr::meta::is_instantiation_of<T, DataSetSink>;
```
RalphSteinhagen, line 40: This is not really a 'concept' in terms of 'duck-typing' but basically forcing that only one of two specific implementation/instantiations are allowed. A `requires (...)` clause or template specialisation would have IMO been better since the templates below cannot be fulfilled by any other than the two given implementations.

RalphSteinhagen: OK for now, but 'DataSinkOrDataSetSinkLike' should be removed in one of the upcoming PRs.

RalphSteinhagen (approved):
OK for now. This PR correctly implements that the registered signal name is taken from the signal name of the corresponding sink and not internal DataSet signal names. 


To note: As discussed in person the new `DataSinkOrDataSetSinkLike' isn't a good concept because it matches only two specific implementations and does not support 'duck-typing' or any other implementation to fit this concepts. Should be removed/cleaned-up in one of the upcoming PRs.

blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
drslebedev: I will change it in the upcoming PR, just to merge this PR faster.

PR #506, opened 2025-02-07
Fix DataSinkRegistry

drslebedev:
The PR includes the following changes:

- DataSink cannot be registered with an empty string; it must be set either by the user or via a tag.
- The registration with the same signal name is not allowed.
- Fix qa_DataSink

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp
@@ -133,15 +133,15 @@ public:
     }
 
     template<DataSinkOrDataSetSinkLike TSink>
-    void registerSink(TSink* sink) {
+    void registerSink(TSink* sink, std::source_location location = std::source_location::current()) {
         std::lock_guard lg{_mutex};
 
-        if (sink->signal_name == "") {
-            throw gr::exception("Failed to register sink. Sink signal_name is an empty string.");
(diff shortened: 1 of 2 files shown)


RalphSteinhagen (approved):
This quick fix is OK, as discussed, and I’m happy with the overall direction.

Before merging, there a few minor points that could and maybe should be addressed (no further review necessary though):

1. Improve the `gr::exception` messages (re-)thrown in `settingsChanged(..)`:
   Both `updateSignalName(...)` and `registerSink(...)` correctly throw a `gr::exception` in case of a naming clash. However, it’s not immediately clear which block is causing the error. It might help to add a `std::source_location` parameter to those functions and intercept the exception, augmenting the message with the block’s name (and unique name, to disambiguate blocks where users may have given the same name) to avoid confusion when multiple sinks raise the same error.

2. Consolidate `DataSink` and `DataSetSink` -- *if possible w/ reasonable effort*:
   The `DataSink<T>` and `DataSetSink<T>` classes are very similar and primarily differ only in their storing data type. In principle, one could share or unify these to avoid duplication and keep the maintenance simpler. 
   We are having the same issue in the OpenDigitizer UI w.r.t. `ImPlotData[Set]Sink`. I'd take care of the latter one.

Otherwise, the fix looks good to me. @drslebedev thanks for the quick turnaround on this!

PR #509, opened 2025-02-17
Add multi output ports test.

drslebedev:
When a block has multiple synchronous output ports, any unconnected port defaults to a buffer size of 4096. 
This default value can unintentionally limit all ports’ buffer sizes within `workInternal()` , resulting in chunk sizes constrained to 4096, even if larger buffers were explicitly configured for other ports. The issue is that the user may not notice it since no errors are printed.

A test has been included to replicate this behavior, similar to the Picoscope block, where one unconnected output port causes all output to be reduced to chunks of size 4096.

Current workaround is to connect unused ports to `NullSink`.

### DIFF
--- a/core/test/qa_Graph.cpp
+++ b/core/test/qa_Graph.cpp
@@ -6,6 +6,40 @@
 #include <gnuradio-4.0/testing/NullSources.hpp>
 #include <gnuradio-4.0/testing/TagMonitors.hpp>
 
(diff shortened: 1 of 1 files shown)


RalphSteinhagen (approved):
Nice. Did I mentioned that I like unit-tests?!!

Can be merged as soon as the CI passed.

PR #510, opened 2025-02-18
Add missing signal_quanity default tag

drslebedev:


wirew0rm (approved):
Shouldn't this also be added here so that it is consistently propagated? Or is this tag semantically different from the tags there?
https://github.com/fair-acc/gnuradio4/blob/main/blocks/basic/include/gnuradio-4.0/basic/DataSink.hpp#L372-L380

Otherwise looks good and could be merged.

PR #512, opened 2025-02-25
fix/standardised CONTEXT position in tag property_map

RalphSteinhagen:
* CONTEXT needs to be always in the property-root, since this is also used outside the trigger logic. `TRIGGER_META_INFO` sub-property-map is kept (default: {}) for user-defined/hw-specific info.
  * replaced `"context"` and `"sample_rate"` with `CONTEXT` and `SAMPLE_RATE` definition -> should simplify later refactoring
  * FunctionGenerator block changes:
    * added additional `trigger_...`diagnostics property fields
    * added 'signal_trigger' parameter matcher/criteria
  * `Packet<T>`, `Tensor<T>`, `DataSet<T>` changes:
     *  changed to using straight `pmtv::map_t` type (easier move/copy operation)
     * added `default_value` field for ZOH, range padding, interpolation purposes etc.

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
@@ -15,7 +15,7 @@ using namespace gr;
 namespace function_generator {
 enum class SignalType : int { Const, LinearRamp, ParabolicRamp, CubicSpline, ImpulseResponse };
 // TODO: Enum values should be capitalized CamelStyle, for the moment they are the same as FunctionGenerator class members
-enum class ParameterType : int { signal_type, start_value, final_value, duration, round_off_time, impulse_time0, impulse_time1 };
+enum class ParameterType : int { signal_trigger, signal_type, start_value, final_value, duration, round_off_time, impulse_time0, impulse_time1 };
 
 using enum SignalType;
 using enum ParameterType;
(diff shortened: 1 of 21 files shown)


blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
```diff
@@ -148,9 +148,14 @@ The parameters will automatically update when a Tag containing the "context" fie
     Annotated<T, "impulse_time0", Doc<"Specific to ImpulseResponse, in sec">> impulse_time0  = T(0.);
     Annotated<T, "impulse_time1", Doc<"Specific to ImpulseResponse, in sec">> impulse_time1  = T(0.);
 
+    Annotated<std::string, "trigger name">       trigger_name;
+    Annotated<std::uint64_t, "trigger time">     trigger_time;
+    Annotated<float, "trigger offset">           trigger_offset;
+    Annotated<std::string, "context name">       context;
     Annotated<property_map, "trigger_meta_info"> trigger_meta_info{};
```
drslebedev, line 155: Probably `trigger_meta_info`is no longer needed. Auto update will be triggered for `trigger_name`, `trigger_time`, `trigger_offset `, `context`

drslebedev (approved):
Thanks a lot for the PR!

The changes look good to me.
I left just one minor comment in the code, but the PR can be merged as is.

PR #514, opened 2025-02-26
`-Werror` fixes

RalphSteinhagen:
Added explicit cast where necessary to get the GR4 code-base warning free. 

Also added explicit `-DWARNINGS_AS_ERRORS=<ON|OFF>` (default: ON) cmake flag to control the warning-to-error conversion. For development, you may disable this but for production code (i.e. those that shall be merged to main) this should remain always `ON`.


cmake/CompilerWarnings.cmake
```diff
@@ -52,8 +50,9 @@ function(set_project_warnings project_name)
   )
 
   if(WARNINGS_AS_ERRORS)
-    set(CLANG_WARNINGS ${CLANG_WARNINGS} -Werror)
+    set(CLANG_WARNINGS ${CLANG_WARNINGS} -Werror)  # avoid warnings since they are often indicative of immature API and/or potential sources of bugs
```
frankosterfeld, line 53: Maybe use CMAKE_COMPILE_WARNING_AS_ERROR (https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_WARNING_AS_ERROR.html)?

wirew0rm (approved):

PR #519, opened 2025-03-07
Ccache followup

wirew0rm:
Improves the ccache integration in the CI. See individual commits for details.


wirew0rm:
image

Compile times for a direct rebuild with no changes at all look good, except for Debug, where something seems to break caching. See: https://github.com/fair-acc/gnuradio4/actions/runs/13723214080/job/38383555636?pr=519#step:15:7

RalphSteinhagen (approved):

PR #523, opened 2025-03-11
fixed FunctionGenerator

RalphSteinhagen:
-- non matching trigger should not trigger a function parameter change.

This is a workaround. We may need to look into a user-definable settings pre-condition check that are used to reject settings changes based on -- for example -- required trigger, internal block state, ctx etc.

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
@@ -165,19 +165,27 @@ The parameters will automatically update when a Tag containing the 'context' fie
     function_generator::SignalType _signalType = function_generator::parse<function_generator::SignalType>(signal_type);
     T                              _timeTick   = T(1.) / static_cast<T>(sample_rate);
 
-    void settingsChanged(const property_map& /*old_settings*/, const property_map& new_settings) {
-        if (new_settings.contains(function_generator::toString(function_generator::signal_type))) {
+    void settingsChanged(const property_map& oldSettings, const property_map& newSettings) {
+        if (newSettings.contains(function_generator::toString(function_generator::signal_type))) {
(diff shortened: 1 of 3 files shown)


blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
```diff
@@ -165,19 +165,27 @@ The parameters will automatically update when a Tag containing the 'context' fie
...
                 } else {
-                    // trigger does not match required signal_trigger
+                    // trigger does not match required signal_trigger -- revert signal_type and others to previous ones
+                    signal_type    = function_generator::toString(_signalType);
+                    start_value    = std::get<T>(oldSettings.at("start_value"));
+                    final_value    = std::get<T>(oldSettings.at("final_value"));
+                    duration       = std::get<T>(oldSettings.at("duration"));
+                    round_off_time = std::get<T>(oldSettings.at("round_off_time"));
+                    impulse_time0  = std::get<T>(oldSettings.at("impulse_time0"));
+                    impulse_time1  = std::get<T>(oldSettings.at("impulse_time1"));
```
drslebedev, line 186: Probably we need to check if a oldSettings contains a certain parameter?

drslebedev (approved):
Thanks for the PR!

The implementation looks good. We might consider whether this type of filter should be implemented directly in `Settings`, but for now, your solution already seems solid.

I left a comment about checking whether the parameter exists in `oldSettings`. I believe this check is necessary; otherwise, it may throw an error if the parameter is missing.

blocks/basic/include/gnuradio-4.0/basic/FunctionGenerator.hpp
RalphSteinhagen: The `oldSettings` always contains a copy of all block parameters. So no extra check is needed.

PR #531, opened 2025-03-20
Integrate blocklib generator

ivan-cukic:


blocklib_generator/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
+cmake_minimum_required(VERSION 3.28)
+project(MyProject LANGUAGES CXX)
+
+set(CMAKE_CXX_STANDARD 23)
+
+option(ENABLE_SPLIT "Enable splitting of generated files (pass --split to the parser)" ON)
```
RalphSteinhagen, line 6: Need a better name for `ENABLE_SPLIT` not clear without the CMake file context.

blocks/libs/test/qa_KnownSharedLibBlocks.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+extern "C" {
+bool gr_blocklib_init_module_GrFilterBlocks(gr::BlockRegistry& registry);
+bool gr_blocklib_init_module_GrFourierBlocks(gr::BlockRegistry& registry);
+bool gr_blocklib_init_module_GrElectricalBlocks(gr::BlockRegistry& registry);
+bool gr_blocklib_init_module_GrHttpBlocks(gr::BlockRegistry& registry);
+bool gr_blocklib_init_module_GrFileIoBlocks(gr::BlockRegistry& registry);
+bool gr_blocklib_init_module_GrTestingBlocks(gr::BlockRegistry& registry);
+bool gr_blocklib_init_module_GrBasicBlocks(gr::BlockRegistry& registry);
+}
```
RalphSteinhagen, line 20: Is this needed just for this unit-tests or expected for all user programs?

RalphSteinhagen: Thanks ... yes, would be better if the block-lib user/writer doesn't need to know/care about this.

blocklib_generator/CMakeLists.txt
ivan-cukic: Ideas? I can not think of anything that is not really verbose.

RalphSteinhagen: What about:

```cmake
option(ENABLE_PARALLEL_BLOCK_INSTANTIATIONS 
"Split specific block-type template instantiation into multiple files. Improves parallel compilation, reducing overall build time" 
ON)
```

RalphSteinhagen (approved):
Thanks for fleshing-out and integrating the original block PoC into GR4!

PR #534, opened 2025-03-21
DataSink: Create consistent DataSet for Trigger, Multiplexed and Snap…

drslebedev:
Ensure a consistent `DataSet` structure for `Trigger`, `Multiplexed`, and `Snapshot` modes.

- `DataSet::axis_values` and `DataSet::extents` are now correctly initialized.
- Additional tests were implemented.

### DIFF
--- a/blocks/basic/test/qa_DataSink.cpp
+++ b/blocks/basic/test/qa_DataSink.cpp
@@ -243,6 +243,26 @@ bool spinUntil(std::chrono::milliseconds timeout, auto fnc) {
     return false;
 }
 
+template<typename T>
+void testBasicsDataSet(const DataSet<T>& ds, std::source_location location = std::source_location::current()) {
+    using namespace boost::ut;
+
+    expect(eq(ds.extents.size(), 1UZ)) << fmt::format("Error line: {}\n", location.line());
+    if (ds.extents.size() >= 1) {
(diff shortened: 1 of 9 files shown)


blocks/basic/test/qa_DataSink.cpp
```diff
@@ -243,6 +243,26 @@ bool spinUntil(std::chrono::milliseconds timeout, auto fnc) {
     return false;
 }
 
+template<typename T>
```
RalphSteinhagen, line 246: There are already two existing functions in `DataSetHelper.hpp` that do exactly that (and more):
[gr::dataset::checkDataSetConsistency(..)](https://github.com/fair-acc/gnuradio4/blob/94083cf1f7a2f4fe49384a9efc02e2387ca0f9a2/algorithm/include/gnuradio-4.0/algorithm/dataset/DataSetHelper.hpp#L43C46-L83) and [gr::dataset::verify](https://github.com/fair-acc/gnuradio4/blob/94083cf1f7a2f4fe49384a9efc02e2387ca0f9a2/algorithm/include/gnuradio-4.0/algorithm/dataset/DataSetHelper.hpp#L225-L290).

It would be better to re-use one of these (and perhaps even merge them into one).

RalphSteinhagen (approved):

PR #538, opened 2025-03-25
Fix formatters build on F41, make fmt11-proof

marcusmueller:
- **soapy QA: make Range unambiguous**
- **formatters: Range format() should be const**
- **formatters: only implement std::expected formatter if fmt < 11.0.0**

The build of qa_formatters failed for reasons of lacking constness; so fixed that.

While I was at it, fixed related alias resolution ambiguity making soapy stuff ftb.

And since at that point it was worth trying: with this changeset, GR4 also builds with fmt 11, not only 10.

### DIFF
--- a/blocks/soapy/test/qa_Soapy.cpp
+++ b/blocks/soapy/test/qa_Soapy.cpp
@@ -25,6 +25,7 @@ const boost::ut::suite<"basic SoapySDR API "> basicSoapyAPI = [] {
     using namespace boost::ut;
     using namespace gr;
     using namespace gr::blocks::soapy;
+    using gr::blocks::soapy::Range;
 
     "helper functions"_test = [] { "range printer"_test = [] { expect(eq(fmt::format("{}", gr::blocks::soapy::Range{1.0, 10.0, 0.5}), std::string("Range{min: 1, max: 10, step: 0.5}"))); }; };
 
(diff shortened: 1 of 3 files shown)


RalphSteinhagen (approved):
@marcusmueller thanks for your PR. Concise and to the point ... that's how we like it.

marcusmueller:
as requested, rebased and pushed. if restyled continues to fail with a git error, I'd request ignoring that check.

RalphSteinhagen:
Restyled passed now: :+1: :smile: 

image

Will wait for the other CI step to pass and then merge. Code-coverage may fail because of non-sharable integration secrets, but I'll ignore that until we find a safe fix.

marcusmueller:
wonderful. Really strange that Restyled failed – 4 times – on the original. But nothing I'll lose sleep over :)

RalphSteinhagen:
> wonderful. Really strange that Restyled failed – 4 times – on the original. But nothing I'll lose sleep over :)

I changed the restyled CI config to allow for PRs from public remote repos. Was too restrictive initially

marcusmueller:
ah cool, thanks!

marcusmueller:
the gcc 13 debug build seems to fail because sonar can't find the project.

RalphSteinhagen:
> the gcc 13 debug build seems to fail because sonar can't find the project.

It's GCC14, and your user account doesn't have the GitHub `sonar.projectKey` and `sonar.organization` credentials to upload the result. OK in this case.

RalphSteinhagen:
@all-contributors please add @marcusmueller for infra

PR #539, opened 2025-03-26
Tags forwarding

drslebedev:
This PR  addresses the issue with tag propagation when decimation is used, it includes the following changes: 
* Initialise `autoForwardParameters` with `kDefaultTags`.
* Correctly forward tags when decimation is on:
* Introduce `BackwardTagForwarding` attribute
* Forward and backward tag propagation is supported
* Update `sample_rate` if the block performs decimation or interpolation

### DIFF
--- a/core/include/gnuradio-4.0/Block.hpp
+++ b/core/include/gnuradio-4.0/Block.hpp
@@ -353,6 +353,7 @@ public:
 
     constexpr static bool blockingIO             = std::disjunction_v<std::is_same<BlockingIO<true>, Arguments>..., std::is_same<BlockingIO<false>, Arguments>...>;
     constexpr static bool noDefaultTagForwarding = std::disjunction_v<std::is_same<NoDefaultTagForwarding, Arguments>...>;
+    constexpr static bool backwardTagForwarding  = std::disjunction_v<std::is_same<BackwardTagForwarding, Arguments>...>;
 
     constexpr static block::Category blockCategory = block::Category::NormalBlock;
 
@@ -714,11 +715,15 @@ public:
     /**
      * Merge tags from all sync ports into one merged tag, apply auto-update parameters
      */
-    constexpr void updateMergedInputTagAndApplySettings(auto& inputSpans, std::size_t untilLocalIndex = 1UZ) noexcept {
+    void updateMergedInputTagAndApplySettings(auto& inputSpans, std::size_t untilLocalIndex = 1UZ) noexcept {
         for_each_reader_span(
             [this, untilLocalIndex](auto& in) {
                 if (in.isSync) {
-                    for (const auto& [key, value] : in.getMergedTag(untilLocalIndex).map) {
+                    std::size_t untilLocalIndexAdjusted = untilLocalIndex;
+                    if constexpr (!backwardTagForwarding) {
+                        untilLocalIndexAdjusted = 1UZ;
+                    }
+                    for (const auto& [key, value] : in.getMergedTag(untilLocalIndexAdjusted).map) {
                         _mergedInputTag.map.insert_or_assign(key, value);
                     }
                 }
@@ -760,10 +765,10 @@ public:
                     using enum gr::SpanReleasePolicy;
                     if constexpr (std::remove_cvref_t<Port>::kIsInput) {
                         if constexpr (std::remove_cvref_t<Port>::kIsSynch) {
-                            return std::forward<Port>(port).template get<ProcessAll>(nSyncSamples);
+                            return std::forward<Port>(port).template get<ProcessAll, !backwardTagForwarding>(nSyncSamples);
                         } else {
                             // return std::forward<Port>(port).template get<ProcessNone>(std::max(port.min_samples, std::min(port.streamReader().available(), port.max_samples)));
-                            return std::forward<Port>(port).template get<ProcessNone>(port.streamReader().available());
+                            return std::forward<Port>(port).template get<ProcessNone, !backwardTagForwarding>(port.streamReader().available());
                         }
                     } else if constexpr (std::remove_cvref_t<Port>::kIsOutput) {
                         if constexpr (std::remove_cvref_t<Port>::kIsSynch) {
@@ -1612,7 +1617,8 @@ protected:
         auto inputSpans  = prepareStreams(inputPorts<PortType::STREAM>(&self()), processedIn);
         auto outputSpans = prepareStreams(outputPorts<PortType::STREAM>(&self()), processedOut);
 
-        updateMergedInputTagAndApplySettings(inputSpans);
+        updateMergedInputTagAndApplySettings(inputSpans, processedIn);
+
         applyChangedSettings();
 
         // Actual publishing occurs when outputSpans go out of scope. If processedOut == 0, the Tags will not be published.
--- a/core/include/gnuradio-4.0/Port.hpp
+++ b/core/include/gnuradio-4.0/Port.hpp
@@ -462,7 +462,7 @@ struct Port {
     template<SpanReleasePolicy spanReleasePolicy>
     using ReaderSpanType = decltype(std::declval<ReaderType>().template get<spanReleasePolicy>());
 
-    template<SpanReleasePolicy spanReleasePolicy>
+    template<SpanReleasePolicy spanReleasePolicy, bool consumeOnlyFirstTag = false>
     struct InputSpan : public ReaderSpanType<spanReleasePolicy> {
         TagReaderSpanType rawTags;
         std::size_t       streamIndex;
@@ -490,11 +490,15 @@ struct Port {
                     return;
                 }
 
-                if (ReaderSpanType<spanReleasePolicy>::isConsumeRequested()) {
-                    consumeTags(ReaderSpanType<spanReleasePolicy>::nRequestedSamplesToConsume());
+                if constexpr (consumeOnlyFirstTag) {
+                    consumeTags(1UZ);
                 } else {
-                    if (ReaderSpanType<spanReleasePolicy>::spanReleasePolicy() == SpanReleasePolicy::ProcessAll) {
-                        consumeTags(ReaderSpanType<spanReleasePolicy>::size());
+                    if (ReaderSpanType<spanReleasePolicy>::isConsumeRequested()) {
+                        consumeTags(ReaderSpanType<spanReleasePolicy>::nRequestedSamplesToConsume());
+                    } else {
+                        if (ReaderSpanType<spanReleasePolicy>::spanReleasePolicy() == SpanReleasePolicy::ProcessAll) {
+                            consumeTags(ReaderSpanType<spanReleasePolicy>::size());
+                        }
                     }
                 }
             }
@@ -532,8 +536,8 @@ struct Port {
             return reader.get(n);
         }
     }; // end of InputSpan
-    static_assert(ReaderSpanLike<InputSpan<gr::SpanReleasePolicy::ProcessAll>>);
-    static_assert(InputSpanLike<InputSpan<gr::SpanReleasePolicy::ProcessAll>>);
+    static_assert(ReaderSpanLike<InputSpan<gr::SpanReleasePolicy::ProcessAll, false>>);
+    static_assert(InputSpanLike<InputSpan<gr::SpanReleasePolicy::ProcessAll, false>>);
 
     template<SpanReleasePolicy spanReleasePolicy>
     using WriterSpanType = decltype(std::declval<WriterType>().template reserve<spanReleasePolicy>(1UZ));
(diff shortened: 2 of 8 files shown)


core/include/gnuradio-4.0/Settings.hpp
```diff
@@ -304,8 +304,6 @@ struct SettingsBase {
 
     [[nodiscard]] virtual std::set<std::string> autoUpdateParameters(SettingsCtx ctx = {}) noexcept = 0;
 
-    [[nodiscard]] virtual std::set<std::string>& autoForwardParameters() noexcept = 0;
```
RalphSteinhagen, line 307: I think this parameter is still needed albeit needs to be initially populated with the default tags.

The block needs to be able to suppress auto-forwarding of specific tags because some blocks -- notably the decimating blocks -- need to apply and swallow the upstream `sample_rate` related parameter and issue a new 'sample_rate' with the modified (reduced) one.

There needs to be a unit-test for this behaviour..

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -714,11 +715,18 @@ class Block : public lifecycle::StateMachine<Derived> {
...
      */
-    constexpr void updateMergedInputTagAndApplySettings(auto& inputSpans, std::size_t untilLocalIndex = 1UZ) noexcept {
+    void updateMergedInputTagAndApplySettings(auto& inputSpans, std::size_t untilLocalIndex = 1UZ) noexcept {
         for_each_reader_span(
             [this, untilLocalIndex](auto& in) {
                 if (in.isSync) {
-                    for (const auto& [key, value] : in.getMergedTag(untilLocalIndex).map) {
+                    std::size_t untilLocalIndex2 = untilLocalIndex;
+                    if constexpr (backwardTagForwarding) {
+                        in.setConsumeOnlyFirstTag(false);
```
RalphSteinhagen, line 724: This is a static information. Why making it a dynamic setting in Port? Should be linked to -- for example -- NTTP parameter in the tag reader.

core/include/gnuradio-4.0/Port.hpp
```diff
@@ -466,8 +466,9 @@ struct Port {
     struct InputSpan : public ReaderSpanType<spanReleasePolicy> {
         TagReaderSpanType rawTags;
         std::size_t       streamIndex;
-        bool              isConnected = true; // true if Port is connected
-        bool              isSync      = true; // true if  Port is Sync
+        bool              isConnected         = true; // true if Port is connected
+        bool              isSync              = true; // true if  Port is Sync
+        bool              consumeOnlyFirstTag = false;
```
RalphSteinhagen, line 471: Not clear whether the 'consumeOnlyFirstTag' is actually needed in the Port class. It's a property function of the Block<T> and static rather a dynamic port info. Probably the wrong place to implement this.

core/include/gnuradio-4.0/Settings.hpp
```diff
@@ -409,9 +406,6 @@ class CtxSettings : public SettingsBase {
 
                 // detect whether field has one of the DEFAULT_TAGS signature
                 if constexpr (settings::isWritableMember<Type, MemberType>()) {
-                    if constexpr (std::ranges::find(gr::tag::kDefaultTags, refl::data_member_name<TBlock, kIdx>.view()) != gr::tag::kDefaultTags.cend()) {
-                        _autoForwardParameters.emplace(memberName);
```
RalphSteinhagen, line 413: `_autoForwardParameters` need to be initialised with the `gr::tag::kDefaultTags` list, and keys that are being set explicitely via the users removed from the auto-forwarding list to avoid having parameters propagated that are statically set.

core/test/qa_Tags.cpp
```diff
@@ -240,6 +287,56 @@ const boost::ut::suite TagPropagation = [] {
...
+        expect(equal_tag_lists(sink._tags, expectedTags));
+    };
+
+    "Tag propagation with decimation - Forward policy"_test = [&runPolicyTest]() {
+        std::vector<Tag>       expectedTags = std::vector<Tag>{                             //
+            {0, {{"key", "value@0"}, {"key0", "value@0"}}},                           //
+            {1, {{"key", "value@5"}, {"key4", "value@4"}, {"key5", "value@5"}}},      //
+            {2, {{"key", "value@20"}, {"key15", "value@15"}, {"key20", "value@20"}}}, //
+            {3, {{"key", "value@25"}, {"key25", "value@25"}}}};
+        runPolicyTest.template operator()<DecimatorForward<float>>(expectedTags);
```
RalphSteinhagen, line 329: the kustom 'key', 'key0' should not be forwarded w/o being explicitely set in the autoForward set (in Settings.hpp)

RalphSteinhagen (changes requested):
Needs changes as discussed and indicated in the comments above.

core/include/gnuradio-4.0/Port.hpp
drslebedev: The actual consumption happens in the destructor of `Port::InputSpan`, so at this point we need to determine whether we want to consume just a single tag or all tags within the range.  

One possible way to achieve the desired behavior would be to update the tag indices in `InputSpan::rawTags`, but that’s not feasible since they are read-only.

core/include/gnuradio-4.0/Settings.hpp
drslebedev: Done

core/test/qa_Tags.cpp
drslebedev: Currently, all tags are forwarded without considering `Settings::autoForwardTParameters`. I completely agree that this functionality should be implemented. However, integrating it would require significant changes across many unit tests, which is likely beyond the scope of this PR.

I would suggest keeping the current behavior for now. In a future PR, we can introduce support for `Settings::autoForwardParameters`, ensure it is properly applied to all forwarded tags, and then update the relevant unit tests accordingly—rather than changing only this test prematurely.

core/test/qa_Settings.cpp
```diff
@@ -131,8 +123,8 @@ const boost::ut::suite SettingsTests = [] {
         expect(eq(src.sample_rate, 42.f)) << "check map constructor";
         expect(eq(src._nSamplesProduced, gr::Size_t(0))) << "default value";
         expect(eq(src.settings().getNAutoUpdateParameters(), 1UZ));
-        expect(eq(src.settings().autoUpdateParameters().size(), 3UL));  // 3 base + 0 derived
-        expect(eq(src.settings().autoForwardParameters().size(), 1UL)); // sample_rate
+        expect(eq(src.settings().autoUpdateParameters().size(), 3UL));   // 3 base + 0 derived
+        expect(eq(src.settings().autoForwardParameters().size(), 16UL)); // all gr::tag::kDefaultTags
```
RalphSteinhagen, line 127: Using `gr::tag::kDefaultTags.size()` would be more expressive, avoid the code comment, and simplify later refactoring.

core/test/qa_Settings.cpp
```diff
@@ -141,24 +133,24 @@ const boost::ut::suite SettingsTests = [] {
...
 
         auto& sink = testGraph.emplaceBlock<Sink<float>>();
         expect(eq(sink.settings().defaultParameters().size(), 9UZ)); // 7 base + 2 derived
         expect(eq(sink.settings().getNStoredParameters(), 1UZ));
         expect(eq(sink.settings().getStored().value().size(), 9UZ));
         expect(eq(sink.settings().getNAutoUpdateParameters(), 1UZ));
-        expect(eq(sink.settings().autoUpdateParameters().size(), 5UL));  // 3 base + 2 derived
-        expect(eq(sink.settings().autoForwardParameters().size(), 1UL)); // sample_rate
+        expect(eq(sink.settings().autoUpdateParameters().size(), 5UL));   // 3 base + 2 derived
+        expect(eq(sink.settings().autoForwardParameters().size(), 16UL)); // all gr::tag::kDefaultTags
```
RalphSteinhagen, line 145: Using `gr::tag::kDefaultTags.size()` would be more expressive, avoid the code comment, and simplify later refactoring.

core/test/qa_Settings.cpp
```diff
@@ -141,24 +133,24 @@ const boost::ut::suite SettingsTests = [] {
...
 
         // need to add 'n_samples_max' to forwarding list for the block to automatically forward it as the 'n_samples_max' tag is not part of the canonical 'gr::tag::kDefaultTags' list
         block1.settings().autoForwardParameters().emplace("n_samples_max");
-        expect(eq(block1.settings().autoForwardParameters().size(), 3UL));
+        expect(eq(block1.settings().autoForwardParameters().size(), 17UL)); // all gr::tag::kDefaultTags + n_samples_max
```
RalphSteinhagen, line 149: Using `gr::tag::kDefaultTags.size() + 1UZ` would be more expressive, avoid the code comment, and simplify later refactoring.

core/test/qa_Settings.cpp
```diff
@@ -141,24 +133,24 @@ const boost::ut::suite SettingsTests = [] {
...
         block2.settings().autoForwardParameters().emplace("n_samples_max");
-        expect(eq(block2.settings().autoForwardParameters().size(), 3UL));
+        expect(eq(block2.settings().autoForwardParameters().size(), 17UL)); // all gr::tag::kDefaultTags + n_samples_max
```
RalphSteinhagen, line 151: idem

core/test/qa_Settings.cpp
```diff
@@ -141,24 +133,24 @@ const boost::ut::suite SettingsTests = [] {
...
         sink.settings().autoForwardParameters().emplace("n_samples_max");
-        expect(eq(sink.settings().autoForwardParameters().size(), 2UL));
+        expect(eq(sink.settings().autoForwardParameters().size(), 17UL)); // all gr::tag::kDefaultTags + n_samples_max
```
RalphSteinhagen, line 153: idem

core/test/qa_Tags.cpp
```diff
@@ -240,6 +287,56 @@ const boost::ut::suite TagPropagation = [] {
         expect(eq(sink._nSamplesProduced, 1008U)) << "sinkOne did not consume enough input samples"; // default policy is to drop epilogue samples
         expect(eq(sink._tags.size(), 3UZ));                                                          // default policy is to drop epilogue samples
     };
+
+    auto runPolicyTest = []<typename TDecimator>(const std::vector<Tag>& expectedTags) {
```
RalphSteinhagen, line 291: please use the UT API for parameterised tests, e.g. https://github.com/boost-ext/ut/blob/master/example/parameterized.cpp

Rationale: exceptions/expect throw correct line-numbers and associated tests where they are being called.

core/include/gnuradio-4.0/Settings.hpp
```diff
@@ -873,6 +870,13 @@ class CtxSettings : public SettingsBase {
 
             updateActiveParametersImpl();
 
+            // Update sample_rate if the block performs decimation or interpolation
+            if (result.forwardParameters.contains(gr::tag::SAMPLE_RATE.shortKey()) && (_block->input_chunk_size != 1ULL || _block->output_chunk_size != 1ULL)) {
```
RalphSteinhagen, line 874: This 'if' should be guarded by a constexpr-if on the condition that the input->output rate can actually change (i.e. `input_chunk_size` and `output_chunk_size` not being `const` as defined by `ResamplingControl<...>`).

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -353,6 +353,7 @@ class Block : public lifecycle::StateMachine<Derived> {
 
     constexpr static bool blockingIO             = std::disjunction_v<std::is_same<BlockingIO<true>, Arguments>..., std::is_same<BlockingIO<false>, Arguments>...>;
     constexpr static bool noDefaultTagForwarding = std::disjunction_v<std::is_same<NoDefaultTagForwarding, Arguments>...>;
```
RalphSteinhagen, line 355: There is no strong need for a static bool definition for `noDefaultTagForwarding` and `backwardTagForwarding`. 

It would be better if these are defined via `using ...` statements. This way they do not occupy space in the final binary representation/tables and just compile away to nothing.

core/test/qa_Tags.cpp
RalphSteinhagen: This should not be dropped but documented as a follow-up. We do not want people to base their implementation based on the assumption used in the unit-test.

core/test/qa_Tags.cpp
drslebedev: The issue is that it's not straightforward to parametrize tests when you have two parameters: one representing a block type and another containing an array of expected values. In such cases, an additional struct must be created to handle this clearly.

The line numbers and related test outputs will appear correctly regardless of which approach is chosen. Here's an example:

```text
Running test "Tag propagation with decimation - Backward policy"... FAILED
in: /home/slebedev/Dev/gnuradio4/core/test/qa_Tags.cpp:318 - test condition:  [4 == 46]
```

RalphSteinhagen (approved):
Thanks for making the changes.

PR #542, opened 2025-03-28
USE_CCACHE=On instead of using compiler launcher

ivan-cukic:


RalphSteinhagen (approved):
OK, as discussed. The cmake ccache flag seems to automatically also enable the caching for linking which seems to be problematic with generating/linking static`.a` files.

Will merge this once the CI passes (possibly a qa_Message flake).

PR #544, opened 2025-04-02
extended kIsArithmeticLikeValueType for gr::UncertainValues types

RalphSteinhagen:


drslebedev (approved):
Thanks for the quick fix!
Looks good to me.

PR #546, opened 2025-04-04
Don't build blocklib when block registry is disabled

ivan-cukic:


RalphSteinhagen (approved):
OK. 

I was wondering: couldn't the frequent ` if(ENABLE_BLOCK_REGISTRY)` conditions be absorbed directly into the `gr_add_block_library(..)` calls?

Issue #161, opened 2023-09-13
List of GR 3.10 blocks sorted by categories and priority


drslebedev:
With this issue we want to create a list of  blocks from GR 3.10 which have to be implemented in graph-prototype.
The list is sorted by categories and priority for GSI/FAIR.

Harishankar14:
Hey, i just wanted to know, some of the high priority blocks,which needed to be migrated ?

RalphSteinhagen:
@Harishankar14, great and thanks for your interest and willingness to contribute. 👍 

The choice is yours and depends on your preference and expertise. Let's start with something simple (e.g. the boolean or byte operators) to get acquainted with the new block API and then move on to something more complex. 

If you are into HAM then the AM or FM blocks might also be a good start. We deliberately left them unimplemented so that others -- like perhaps you -- could grow into the block lib concepts.

If you have some more C++ experience and thorough understanding of DFT and FFT algorithms, there are also some other `std::simd`-related project aspects that would greatly help.

As always, feel free to ask questions on the GR channels (https://matrix.to/#/#gnuradio:gnuradio.org and https://matrix.to/#/#development:gnuradio.org). 
For GR4 specifically: https://matrix.to/#/#architecture:gnuradio.org
That's often a good and perhaps better place to start. :smile: 

<small><small>*to note: 'porting' and 'migrating' implies functionality, not just copying code, and may require some implementing from scratch. We are trying to keep the core LGPLv3, so beware that your ideas/resources are compatible (most are, except GPLv3) .*</small></small>

Harishankar14:
yeah, sure, I will look into it and provide the help that the community wants ! Btw, I am there in the GNU Channels, I go by the name **OnlyDPCanSaveMe**

jameshu15869:
Would these be OOT modules based on the tutorial on the GNURadio website for [GNU Radio 3.x](https://wiki.gnuradio.org/index.php?title=Creating_C%2B%2B_OOT_with_gr-modtool)? 

I took a look at #395 and the structure looks to be different from the OOT modules on tutorial.

RalphSteinhagen:
@jameshu15869 great and thanks for your interest in contributing.

[...]

PR #550, opened 2025-04-09
Add Complex Conjugate  block with tests 

Harishankar14:
Refers #161 
This PR basically is meant for the implementation of Complex Conjugate Block Under Math. 

The Complex Conjugate block processes complex input signals, producing their conjugates by retaining the real part and inverting the sign of the imaginary part. For an input of the form x + yi, where x represents the real component and y the imaginary component, the output is x - yi. This operation is vital for complex signal manipulation in digital signal processing applications.

Changes Made:

1.  Implemented and tested the Complex Conjugate block locally to ensure correct functionality. 

2. Added comprehensive unit test cases in qa_Math.cpp to validate the block’s behavior for float and  double inputs.

3. Verified successful execution of all tests with make && ./qa_Math, confirming no regressions in existing math block tests.
 
### Test Case Example
Screenshot From 2025-04-09 23-59-48

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,7 +231,9 @@ endif()
 
 # include header-only libraries that have been inlined to simplify builds w/o requiring access to the internet
 if(NOT (TARGET magic_enum))
-  add_library(magic_enum INTERFACE)
+  add_library(magic_enum INTERFACE
+          blocks/basic/include/gnuradio-4.0/basic/ClockSource.hpp
+          blocks/basic/include/gnuradio-4.0/basic/ClockSource.hpp)
   target_include_directories(magic_enum ${CMAKE_EXT_DEP_WARNING_GUARD}
                              INTERFACE ${PROJECT_SOURCE_DIR}/third_party/magic_enum/)
 endif()
@@ -239,7 +241,10 @@ endif()
 # include exprtk header-only libraries available as a statically linked library to simplify/speed-up builds
 add_library(
   exprtk STATIC "${CMAKE_CURRENT_SOURCE_DIR}/third_party/exprtk.hpp"
-                "${CMAKE_CURRENT_SOURCE_DIR}/third_party/exprtk.cpp" # dummy source file
+                "${CMAKE_CURRENT_SOURCE_DIR}/third_party/exprtk.cpp"
+        blocks/basic/include/gnuradio-4.0/basic/ClockSource.hpp
+        blocks/basic/include/gnuradio-4.0/basic/ClockSource.hpp
+        blocks/basic/include/gnuradio-4.0/basic/ClockSource.hpp # dummy source file
 )
 target_include_directories(exprtk PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/third_party"
                                          "${CMAKE_CURRENT_SOURCE_DIR}/third_party/exprtk")
@@ -286,11 +291,15 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "(GNU|Clang)") # WIP
   find_package(SoapySDR CONFIG)
 endif()
 
-add_library(pmtv INTERFACE)
+add_library(pmtv INTERFACE
+        blocks/basic/include/gnuradio-4.0/basic/ClockSource.hpp
+        blocks/basic/include/gnuradio-4.0/basic/ClockSource.hpp)
 target_include_directories(pmtv INTERFACE ${pmt_SOURCE_DIR}/include/)
 target_link_libraries(pmtv INTERFACE)
 
-add_library(vir INTERFACE)
+add_library(vir INTERFACE
+        blocks/basic/include/gnuradio-4.0/basic/ClockSource.hpp
+        blocks/basic/include/gnuradio-4.0/basic/ClockSource.hpp)
 target_include_directories(vir INTERFACE ${vir-simd_SOURCE_DIR}/)
 
 # FFTW3 is build 2 times for float and double precisions
@@ -371,7 +380,9 @@ ExternalProject_Add(
   INSTALL_COMMAND ""
   LOG_DOWNLOAD ON)
 
-add_library(fftw INTERFACE)
+add_library(fftw INTERFACE
+        blocks/basic/include/gnuradio-4.0/basic/ClockSource.hpp
+        blocks/basic/include/gnuradio-4.0/basic/ClockSource.hpp)
 target_link_libraries(
   fftw
   INTERFACE fftw3
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,15 +5,19 @@ set(CMAKE_CXX_STANDARD 23)
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
 
-add_library(gnuradio-options INTERFACE)
+# Mainly for FMT
+set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
 
 if(CMAKE_CXX_COMPILER MATCHES "/em\\+\\+(-[a-zA-Z0-9.])?$") # if this hasn't been set before via e.g. emcmake
   message(" Transpiling to WASM: using: Emscripten (${CMAKE_CXX_COMPILER})")
   set(EMSCRIPTEN ON)
 endif()
 
-option(ENABLE_BLOCK_PLUGINS "Enable building the plugin system" ON)
-option(ENABLE_BLOCK_REGISTRY "Enable building the block registry" ON)
+# This option affects the GR build, as well as the behavior of BlockLib cmake functions that are used in out-of-tree
+# modules
+option(GR_ENABLE_BLOCK_REGISTRY "Enable building the block registry" ON)
+
+option(INTERNAL_ENABLE_BLOCK_PLUGINS "Enable building the plugin system" ON)
 option(EMBEDDED "Enable embedded mode" OFF)
 if(EMSCRIPTEN)
   option(WARNINGS_AS_ERRORS "Enable -Werror flags" OFF) # some ubiquous warnings in UT need to be fixed first.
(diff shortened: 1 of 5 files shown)


blocks/math/test/CMakeLists.txt
```diff
@@ -1,3 +1,14 @@
+
+find_package(Volk REQUIRED)
```
jsallay, line 2: If you switch to the process one function, then we won't need to depend on volk.

blocks/math/test/CMakeLists.txt
```diff
@@ -1,3 +1,14 @@
...
+find_package(Boost REQUIRED COMPONENTS unit_test_framework)
```
jsallay, line 3: We don't want to add a dependency on Boost to the core of gnuradio.

jsallay (changes requested):

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -108,13 +116,64 @@ struct MathOpMultiPortImpl : Block<MathOpMultiPortImpl<T, op>> {
...
+        }
+        else if constexpr (std::is_same_v<T, std::complex<double>>) {
+            std::transform(input_span.begin(), input_span.end(), output_span.begin(),
+                [](const std::complex<double>& z) {
+                    return std::conj(z);
+                });
+        }
+        return gr::work::Status::OK;
+    }
+};
```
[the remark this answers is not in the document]
Harishankar14: @jsallay Sure, why not, I did have an intuition to change it to processOne, but I thought, I give a go using processBulk.. anyways, I  changed it to processOne, and tested it out, it works and passes all the testcases,will push those changes.

blocks/math/test/CMakeLists.txt
Harishankar14: Yep, understood,

blocks/math/test/CMakeLists.txt
Harishankar14: Yep

Harishankar14:
@jsallay Made the changes, which u said to do so.

blocks/math/include/gnuradio-4.0/math/Math.hpp
jsallay, line 1: It looks like the `GR_REGISTER_BLOCK` macros were moved around from where they were.  Can you move them back?  I would like to have a minimal set of changes to merge in.

blocks/math/test/qa_Math.cpp
```diff
@@ -1,11 +1,12 @@
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wconversion"
```
jsallay, line 2: Rather than ignore warnings, we should fix them in the code.  This is likely fixed by doing a `static_cast<>`.

blocks/math/test/qa_Math.cpp
```diff
@@ -1,11 +1,12 @@
...
 #include <boost/ut.hpp>
-
 #include <gnuradio-4.0/math/Math.hpp>
-
 #include <gnuradio-4.0/Block.hpp>
 #include <gnuradio-4.0/Graph.hpp>
 #include <gnuradio-4.0/Scheduler.hpp>
 #include <gnuradio-4.0/testing/TagMonitors.hpp>
+#include <volk/volk.h>
```
jsallay, line 9: Please remove this header.  It is not needed.

blocks/math/test/qa_Math.cpp
```diff
@@ -23,10 +24,14 @@ void test_block(const TestParameters<T> p) {
...
-    auto& block = graph.emplaceBlock<BlockUnderTest>({{"n_inputs", n_inputs}});
+    //auto& block = graph.emplaceBlock<BlockUnderTest>({{"n_inputs", n_inputs}});
+    auto& block = std::is_same_v<BlockUnderTest,ConjugateImpl<T>>
+        ? graph.emplaceBlock<BlockUnderTest>()
+        : graph.emplaceBlock<BlockUnderTest>({{"n_inputs", n_inputs}});
     for (Size_t i = 0; i < n_inputs; ++i) {
         auto& src = graph.emplaceBlock<TagSource<T>>({{"values", p.inputs[i]}, {"n_samples_max", static_cast<Size_t>(p.inputs[i].size())}});
-        expect(eq(graph.connect(src, "out"s, block, "in#"s + std::to_string(i)), ConnectionResult::SUCCESS)) << fmt::format("Failed to connect output port of src {} to input port 'in#{}' of block", i, i);
+        std::string input_port=std::is_same_v<BlockUnderTest,ConjugateImpl<T>> ? "in"s : "in#" + std::to_string(i);
+        expect(eq(graph.connect(src, "out"s, block, input_port), ConnectionResult::SUCCESS)) << fmt::format("Failed to connect output port of src {} to input port {}", i, input_port);
```
jsallay, line 34: I'm not really sure why the special logic is needed for the `ConjugateImpl` block.  Is there a reason it can't follow the same format as the other blocks?

jsallay (changes requested):

blocks/math/include/gnuradio-4.0/math/Math.hpp
Harishankar14: Yes. Fixed it.

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -5,9 +5,15 @@
 #include <gnuradio-4.0/BlockRegistry.hpp>
 #include <gnuradio-4.0/DataSet.hpp>
 #include <gnuradio-4.0/meta/UncertainValue.hpp>
-
+#include <vector>
+#include <algorithm>
+#include <complex>
+#include <functional>
+#include <span>
+#include <volk/volk.h>
```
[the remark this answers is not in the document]
Harishankar14: sorry :)

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -108,13 +116,47 @@ struct MathOpMultiPortImpl : Block<MathOpMultiPortImpl<T, op>> {
...
+    using Description = Doc<R"("@brief This block returns the conjugate by inverting the imaginary part of the input signal")">;
+
+    // Ports
+    PortIn<T> in{};
+    PortOut<T> out{};
+
+    GR_MAKE_REFLECTABLE(ConjugateImpl, in, out);   // macros used for the reflection class 
+    
+    static std::shared_ptr<ConjugateImpl<T>> make() {     // returns a smart pointer to ConjugateImpl
+        return std::make_shared<ConjugateImpl<T>>();      // new instance of ConjugateImpl dynamically and wraps it ina shrared_ptr.
```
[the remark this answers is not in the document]
Harishankar14: Yep, removed.

blocks/math/test/qa_Math.cpp
Harishankar14: Yes, fixed it.

blocks/math/test/qa_Math.cpp
Harishankar14: The special logic for ConjugateImpl is necessary because it has a fixed single input port ("in"), unlike other blocks that can have a variable number of input ports configured by n_inputs. When emplacing the block, we don’t pass n_inputs for ConjugateImpl because it doesn’t accept this parameter—it’s hardcoded to have one input. When connecting sources, we always use the "in" port for ConjugateImpl, while for other blocks, we connect to distinct "in#i" ports based on the index i. This reflects the fundamental difference in how ConjugateImpl is designed (likely for a single-input operation like conjugation) compared to other blocks that support multiple inputs.Making ConjugateImpl follow the same format as other blocks (i.e., accepting n_inputs and using "in#i" ports) would only make sense if ConjugateImpl were redesigned to support multiple input ports. However, if its purpose is to apply a conjugate operation to a single stream, this wouldn’t align with its intended functionality. Alternatively, we could enforce that all blocks use a single input port and remove the n_inputs flexibility, but that would limit the test’s ability to handle multi-input blocks. The current approach—special casing ConjugateImpl—is the most practical given these constraints.

blocks/testing/include/gnuradio-4.0/testing/TagMonitors.hpp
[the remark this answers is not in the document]
Harishankar14: Yeah, sorry, this was by mistake,

CMakeLists.txt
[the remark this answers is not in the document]
Harishankar14: Woah !!!  I haven't even done or touched anything in the CMakeLists.txt !!!!!! That's really weirdd

blocks/math/test/qa_Math.cpp
Harishankar14: @jsallay Major error occurs in TagMontors.hpp

Harishankar14: Yeah, fixed it, warning persisted in TagMonitors.hpp file.
return mark_tag ? (tagGenerated ? static_cast<T>(1) : static_cast<T>(0)) : static_cast<T>(static_cast<float>(_nSamplesProduced));  made the difference

blocks/math/test/qa_Math.cpp
Harishankar14: @jsallay 
The special logic for ConjugateImpl arises because the Conjugate block inherently has only one input by design. Unlike other blocks that can accept multiple inputs (n_inputs), the ConjugateImpl block always has exactly one input port named "in".

Other blocks use an indexed input naming pattern ("in#0", "in#1", etc.) depending on the number of inputs (n_inputs). However, for a block like ConjugateImpl, there's no indexed input—it’s fixed to a single, non-indexed "in" port. Hence, the logic:

ConjugateImpl has exactly one input ("in"), not parameterizable by n_inputs.

Other blocks support multiple inputs (parameterized by n_inputs) and thus use indexed input ports ("in#0", "in#1", ...).

RalphSteinhagen:
@Harishankar14 do you plan to continue working on this PR? If yes, you may need to rebase against the latest main commit.

Harishankar14:
@RalphSteinhagen yes, okay, I actually had to reset my entire system, so can I like start a fresh PR, with the updated changes?

PR #553, opened 2025-04-15
Sergio/246

iamsergio:
Add kGraphGRC message for setting/getting a YAML graph

For  #246


iamsergio:
It's done, but will undraft only after https://github.com/fair-acc/gnuradio4/pull/555 is merged, so I can rebase (change required for tests to work).

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -343,6 +350,7 @@ class Graph : public gr::Block<Graph> {
         propertyCallbacks[graph::property::kEmplaceEdge]        = std::mem_fn(&Graph::propertyCallbackEmplaceEdge);
         propertyCallbacks[graph::property::kRemoveEdge]         = std::mem_fn(&Graph::propertyCallbackRemoveEdge);
         propertyCallbacks[graph::property::kGraphInspect]       = std::mem_fn(&Graph::propertyCallbackGraphInspect);
+        propertyCallbacks[graph::property::kGraphGRC]           = std::mem_fn(&Graph::propertyCallbackGraphGRC);
```
wirew0rm, line 353: I feel like the semantics of the graph receiving a message and then replacing itself are a bit weird. For my mental model, the scheduler getting that message as the owner of the graph would make a lot more sense?
This might also be that my model of of the interplay between the components is a bit outdated and the graph has become a lot more dynamic lately.

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -701,6 +709,25 @@ class Graph : public gr::Block<Graph> {
...
 
+    std::optional<Message> propertyCallbackGraphGRC([[maybe_unused]] std::string_view propertyName, Message message) {
+        assert(propertyName == graph::property::kGraphGRC);
+
+        if (message.cmd == message::Command::Get) {
+            message.data = property_map{{"value", gr::saveGrc(*_pluginLoader, *this)}};
+        } else if (message.cmd == message::Command::Set) {
+            const auto& data        = message.data.value();
+            auto        yamlContent = std::get<std::string>(data.at("value"s));
+            *this                   = gr::loadGrc(*_pluginLoader, yamlContent);
```
wirew0rm, line 720: If I read this correctly, this will just replace the current graph with a new one which has been read from the yaml, right? It might do some cleanup in the old graph's destructor, but it cannot do blocking actions like correctly stopping the current graph, before removing it.

Question: Should it be/Is it possible to do this on a running graph, or does it have to be in stopped state? On the other hand this would make sending the message more complicated, since effectively you would need to trigger and wait for the state change before.

wirew0rm (commented):
Looks good in general, just some questions on how this will interact with the lifecycle state of the graph and its blocks, see below. Might all be fine as it is and just need some more explanation for me to update my understanding of how we are going to use this.

core/include/gnuradio-4.0/Graph.hpp
iamsergio: While integrating in OD, I'm seeing a crash. I'll first add a unit-test for "Running Graph" case, so we capture the crash.

@ivan-cukic can comment on the rest

iamsergio:
being folded into #570

PR #557, opened 2025-04-22
Fix build failure if `clang-scan-deps` is not found. 

drslebedev:
This PR includes the following:

- Resolves build failure when `clang-scan-deps` is missing.
- Enhances CMake logic to automatically locate the `clang-scan-deps` executable matching the Clang compiler version and sets `CMAKE_CXX_COMPILER_CLANG_SCAN_DEPS`.
- Adds the `clang-tools` package to Docker to ensure the availability of `clang-scan-deps`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,24 @@ if(NOT DEFINED GR_TOPLEVEL_PROJECT)
   endif()
 endif()
 
+# clang-scan-deps in required by ccache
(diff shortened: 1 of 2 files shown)


wirew0rm (approved):
Looks good and comprehensive.

I don't really like that we have to put this chunk of code inside of our CMake, imo this is working around packaging issues/inadequate tool discovery on cmake's side... but it's definitely better than failing to compile on a common system. Ideally this can be removed with future cmake versions and cmake c++ module tooling becoming more mature.

PR #565, opened 2025-05-03
chrisjohgorman gcc-15 compiler errors

chrisjohgorman:
Arch just updated me to gcc-15.  I realize this is not one of the build environments, but I thought I should fix the build errors and submit the changes for review.

I am not a c++ expert, so please review this patch carefully, although I believe I have created a good patch.  It works under both gcc 14.2.0 and 15.1.1.

What follows are the notes from the commit.
CMakeList.txt
- vir-smid has a space between quotes and suffix which is deprecatd in
  c++23.  gcc-14 lets this go, but gcc-15 catches it.
- add patch to vir-smid via FetchContent

qa_ImChart.cpp
- with gcc-15, yMin and yMax are seen as not used, label them as
  [[maybe_unused]].

YamlPmt.hpp
- we are using auto where the return type cannot be deduced at the 
  point of use.  mark as inline to ensure the compiler knows it's a
  header-only inline template.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -315,6 +315,8 @@ endif()
 set_source_files_properties(third_party/exprtk.cpp PROPERTIES COMPILE_OPTIONS "-fvisibility=hidden")
 # include exprtk header-only - END
 
+set(vir-simd_patch git apply ${CMAKE_CURRENT_SOURCE_DIR}/patches/0001-vir-smid-src_constexpr_wrapper_h.patch)
+
 include(FetchContent)
 FetchContent_Declare(
   fmt
@@ -336,7 +338,10 @@ FetchContent_Declare(
 FetchContent_Declare(
   vir-simd
   GIT_REPOSITORY https://github.com/mattkretz/vir-simd.git
-  GIT_TAG v0.4.0)
+  GIT_TAG v0.4.0
+  PATCH_COMMAND ${vir-simd_patch}
+  UPDATE_DISCONNECTED 1
+)
 
 FetchContent_Declare(
   cpp-httplib
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -299,6 +299,13 @@ target_link_libraries(pmtv INTERFACE)
 add_library(vir INTERFACE)
 target_include_directories(vir INTERFACE ${vir-simd_SOURCE_DIR}/)
 
+# the script configure won't run on win32 without asking bash to do it
+if(WIN32)
+  set(BASH "bash.exe")
+else()
+  set(BASH "")
+endif()
+
 # FFTW3 is build 2 times for float and double precisions
 set(FFTW_PREFIX ${CMAKE_CURRENT_BINARY_DIR}/fftw)
 if(EMSCRIPTEN)
@@ -335,6 +342,7 @@ if(EMSCRIPTEN)
 else()
   set(FFTW_CFLAGS "${CFLAGS} -fPIC -w -O3 -march=native -mtune=native")
   set(FFTW_CONFIG
+      ${BASH}
       ${FFTW_PREFIX}/src/configure
       --enable-silent-rules
       --quiet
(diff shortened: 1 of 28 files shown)


iamsergio:
I'm on Archlinux as well, so very appreciate you got it running, I might cherry-pick locally.

@RalphSteinhagen to decide here, but some comments from me:

- Seems your Microsoft MSVC patches got included in this gcc-15 change ?
- The vir-smid is only a warning, which you can workaround with `-DWARNINGS_AS_ERRORS=OFF` (please confirm). It's preferable to upstream the warning fix and bump sha1, instead of carrying `*.patch` files
- We can add a *nightly* CI that tests gcc-15 (too much load otherwise ?)

chrisjohgorman:
@iamsergio thank you for looking at this.  Indeed it appears that the changes I made to get a build running on windows were included in this PR.  That was not my intent!  Thank you for catching this.  All I intended to push was the branch `chrisjohgorman-gcc-15` or 26e11a1.  I guess I messed it up when merging `main` into my `chrisjhogorman-buiid-ucrt64-win32` branch.  

The branch `chrisjhogorman-build-ucrt64-win32` is almost all the changes I need to get gr4 built under windows.  (Windows msys2 updated to gcc-15 too and I need another change to get gr4 building with gcc-15, which I haven't committed yet.)

I will test today and confirm for you that the `vir-smid` is a warning not an error.  While I like working with `-DWARNINGS_AS_ERRORS=ON`, there is another reason to do so.  `qa_Soapy` includes a header from `soapysdr`, `Config.hpp`. `Config.hpp` includes `ciso646`, which is deprecated and causes a warning in gcc-15.  I did try to submit a PR upstream for that one, but I'm not sure at this time how they're going to get around it.  If I have time today, I'll make a PR for the `vir-smid` folks.

A nightly CI for gcc-15 sounds like a good idea too.

To @RalphSteinhagen, probably best to reject this PR and I will do a new one for gcc-15 that doesn't include my win32 changes.  Thanks again @iamsergio

chrisjohgorman:
@iamsergo, I can confirm that `vir-smid` will build with `-DWARNINGS_AS_ERRORS=OFF`.  I'll make a PR for upstream.

RalphSteinhagen:
Hi @chrisjohgorman — no worries, and thanks again for your effort here.

I'm sorry for not getting back to you sooner. To clarify, GCC 15 isn't officially supported yet, primarily because no Ubuntu LTS version currently uses it. and we didn't want to upset the community. We generally rely on this as a compatibility baseline for GNU Radio builds. That said, we are actively working on a GCC 15 test build.

One of the primary motivations for moving to GCC 15 is to finally drop the `{fmt}` dependency in favour of the standard `<format>` and `<print>` headers — which are supported adequately from GCC 14 onward. This shift helps us avoid ABI issues caused by mismatched `{fmt}` versions. We’ve aimed to maintain compatibility with GCC 13 until recently, but this constraint will be lifted shortly.

What’s still missing:
  * We need a reliable way to install/use GCC 15 in GitHub CI runners.
  * A few minor compatibility fixes.
  * Ensuring everything remains compatible with Clang 18–20.

Thanks again for putting together this PR — and also for identifying that the Win32-specific changes crept in. 

Getting this working for Windows sounds promising, and we'd be happy to accept your support/PRs in that respect.

We are also available on the GNU Radio 'Development' and 'Architecture' matrix channels.

chrisjohgorman:
@iamsergio `vir-simd` already has a patch for -Wdeprecated-literal-operator warnings.  It's at https://github.com/mattkretz/vir-simd/commit/4d865bcb92abc283dffd46cae2106da26cd7feeb.patch .  Not sure what fix will be preferred.  Applying the patch to 0.4.0, or updating to a newer version of `vir-simd`.  Anyway just fyi.

RalphSteinhagen:
@chrisjohgorman you may want to have a look at the PR https://github.com/fair-acc/gnuradio4/pull/571

and perhaps rebase. This includes the latest bump of the SIMD compatibility helper wrapper.

chrisjohgorman:
Sorry for the long delay in replying @RalphSteinhagen and thank you for the info on PR #571.  I have been working on getting Visual Studio Community Edition up and running on my Windows VM.  There were some issues I had and it took longer than I anticipated.  I am now trying to debug a build, but as expected I am running into different errors than with the minimalist gnu for windows I was originally using.  We'll see where this goes.  At the very least I can make some bug reports and who knows maybe even a PR.  Thanks again for the pointer to the new SIMD and ut.

PR #566, opened 2025-05-06
Make unit tests running again.

drslebedev:
This PR includes 3 commits:
1) Make unit tests running again.
Fix ‑ move `enable_testing()` and include(CTest) before any directory that adds tests, right after the option(ENABLE_TESTING …) block and before the first add_subdirectory().

2) Fix qa_DataSink test, add additinal checks.
    Fix in StreamToDataSet: add timing_events only after pre samples are added.
    
 3) Fix gcc-14 Coverage: workaround GCC #68080 by adding --gcov-ignore-parse-errors=negative_hits.warn_once_per_file to gcovr args

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,8 +27,8 @@ jobs:
         uses: dorny/paths-filter@v3
         with:
           filters: |
-             docker:
-               - 'docker/Dockerfile'
+            docker:
+              - 'docker/Dockerfile'
 
       - name: Set up Docker Label
         id: docker-label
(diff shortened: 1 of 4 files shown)


wirew0rm (approved):

dantti:
Does this fix the wrong capture in OD? I started investigating and oddly ctest produced wrong results, but manually running the test produced the correct one, I rebased my PR but I think the sha there was not bumped, should I bump it?

drslebedev:
This commit addresses the specific issue of gnuradio4 tests not running in the CI.

PR #567, opened 2025-05-07
BlockRegistry: do not add duplicated block types.

drslebedev:
BlockRegistry: do not add duplicated block types:

* Rename BlockRegistry methods similar to std::flat_map: insert, contains, create, keys
* Create vector of available blocks on demand from map


ivan-cukic:
Maybe keeping a sorted list of known block types would be better than what we currently do. If it is of any importance, as the things here are not in the hot path. Other than that, the patch looks Ok.

blocks/libs/test/qa_KnownSharedLibBlocks.cpp
```diff
@@ -31,54 +31,55 @@ const boost::ut::suite TagTests = [] {
...
-        expect(registry.isBlockKnown("gr::basic::DataSink<float32>"sv));
-        expect(registry.isBlockKnown("gr::blocks::basic::SchmittTrigger<float32, (gr::trigger::InterpolationMethod)0>"sv));
-        expect(registry.isBlockKnown("gr::electrical::PowerMetrics<float32, 3ul>"sv));
-        expect(registry.isBlockKnown("gr::http::HttpBlock<float32>"sv));
-        expect(registry.isBlockKnown("gr::filter::fir_filter<float32>"sv));
-        expect(registry.isBlockKnown("gr::blocks::fft::FFT<float32>"sv));
+    "CheckAvailableBlocks"_test = [&] {
+        expect(gt(registry.availableBlocks().size(), 20UZ));
+
+        expect(registry.isBlockAvailable("gr::basic::ClockSource"sv));
```
RalphSteinhagen, line 37: what about calling it simply `isAvailable(std::string_view blockName)`? More terse and the method/parameter name makes it self-explanatory.

blocks/libs/test/qa_apptest_KnownStaticLibBlocks.cpp
```diff
@@ -11,7 +11,7 @@ using namespace std::string_literals;
 int main() {
     gr_blocklib_init_module_GrBasicBlocks(gr::globalBlockRegistry());
 
-    auto known = gr::globalBlockRegistry().knownBlocks();
+    auto known = gr::globalBlockRegistry().availableBlocks();
```
RalphSteinhagen, line 14: same as above: perhaps just `availableBlock()`  ... but I am unsure on this one.

core/include/gnuradio-4.0/BlockRegistry.hpp
```diff
@@ -111,18 +107,19 @@ class BlockRegistry {
...
     }
 
-    [[nodiscard]] auto knownBlocks() const { return _blockTypes; }
+    [[nodiscard]] std::vector<std::string> availableBlocks() const {
+        auto view = _blockTypeHandlers | std::views::keys;
+        return {view.begin(), view.end()};
+    }
 
-    bool isBlockKnown(std::string_view block) const { return _blockTypeHandlers.find(block) != _blockTypeHandlers.end(); }
+    [[nodiscard]] bool isBlockAvailable(std::string_view block) const { return _blockTypeHandlers.contains(block); }
```
RalphSteinhagen, line 122: proposed API naming change `isAvailable(std::string_view blockName)`

PR #568, opened 2025-05-07
Don't mix libc++ with libstdc++ in the same executable

iamsergio:
Executables linked with `libc++` where being linked against `gnuradio-blocklib-core.so` which pulled in libstdc++

Something I stumbled on while debugging an unrelated linker error.
I'm surprised tests are not crashing more often


ivan-cukic (approved):

RalphSteinhagen (approved):

RalphSteinhagen:
@all-contributors please add @iamsergio for code

PR #576, opened 2025-05-21
TagMonitors: use InputSpan for TagSink

wirew0rm:
This allows to process more than one tag on the same index. Since tag forwarding is still based on merged_tag, this will only work for blocks immediately following the emitting block until the necessary block and port logic follow-up is done.


RalphSteinhagen (approved):

wirew0rm:
Closing this for now since there were unforseen issues with signed-ness of the tag indices.

PR #591, opened 2025-06-01
unify radix-2+Bluestein FFT support, prep SIMD/SYCL integration

RalphSteinhagen:
… and re-enabled/added benchmarks.

* added pretty-fast FFT support as FFTpf.hpp
* merged radix-2 Cooley-Tukey and Bluestein FFT paths into unified `FFT<T>` implementation supporting arbitrary input lengths.
* introduced per-stage `twiddleFactors`, `bitReverseTable`, `bluesteinExpTable`, `bluesteinChirpFFT` caches
* implemented constexpr `fft_stage_kernel` with specialisation for size-2/4/8—optimised for future `std::simd` and SYCL backends.
* Re-enabled core benchmarks and added a dedicated benchmark suite for all FFT variants (`FFTw`, `FFTpf`, `FFT`), including prime-sized input cases.

This lays the groundwork for native vectorisation (`<simd>`) and transparent SYCL/executor-based offload in upcoming scheduler integrations.

**Benchmark results**:
image


algorithm/CMakeLists.txt
wirew0rm, line 1: For now i think we have tried to keep all dependencies at one point (cmake/Dependencies[Native].cmake). I think the documentation also recommends to only call MakeAvailable once after all FetchContent_Declare calls have been issued.

I would for now merge this as is and tackle reorganization of dependencies separately.

wirew0rm (approved):
Looks good, nice benchmark results and the dependency looks not too complicated.

PR #595, opened 2025-06-06
75% of MATH headers were ported 

KrxGu:
75% of MATH headers were ported and are now compiling/building under GR-4 (AddConst, SubtractConst, MultiplyConst, DivideConst, Add, Subtract, Multiply, Divide, Max, Min, And, Or, Xor, Negate, Not, Abs, Integrate, Argmax. ) Along with their vast test cases. Tried my best to keep it optimized and minimal so that we don't need many templates for header files.


blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -68,6 +68,11 @@ GR_REGISTER_BLOCK("gr::blocks::math::Add", gr::blocks::math::MathOpImpl, ([T], s
...
 GR_REGISTER_BLOCK("gr::blocks::math::Multiply", gr::blocks::math::MathOpImpl, ([T], std::multiplies<[T]>), [ uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, float, double, gr::UncertainValue<float>, gr::UncertainValue<double>, std::complex<float>, std::complex<double> ])
 GR_REGISTER_BLOCK("gr::blocks::math::Divide", gr::blocks::math::MathOpImpl, ([T], std::divides<[T]>), [ uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, float, double, gr::UncertainValue<float>, gr::UncertainValue<double>, std::complex<float>, std::complex<double> ])
+GR_REGISTER_BLOCK("gr::blocks::math::Max", gr::blocks::math::MathOpImpl, ([T], std::max<[T]>), [ uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, float, double, gr::UncertainValue<float>, gr::UncertainValue<double>, std::complex<float>, std::complex<double> ])
+GR_REGISTER_BLOCK("gr::blocks::math::Min", gr::blocks::math::MathOpImpl, ([T], std::min<[T]>), [ uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, float, double, gr::UncertainValue<float>, gr::UncertainValue<double>, std::complex<float>, std::complex<double> ])
```
jsallay: I don't think that `std::complex<T>` works for `std::max`.  We could create a version, but I think it is reasonable to leave it off the list of types.

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -68,6 +68,11 @@ GR_REGISTER_BLOCK("gr::blocks::math::Add", gr::blocks::math::MathOpImpl, ([T], s
...
+GR_REGISTER_BLOCK("gr::blocks::math::And", gr::blocks::math::MathOpImpl, ([T], std::bit_and<[T]>), [ uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, float, double, gr::UncertainValue<float>, gr::UncertainValue<double>, std::complex<float>, std::complex<double> ])
+GR_REGISTER_BLOCK("gr::blocks::math::Or", gr::blocks::math::MathOpImpl, ([T], std::bit_or<[T]>), [ uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, float, double, gr::UncertainValue<float>, gr::UncertainValue<double>, std::complex<float>, std::complex<double> ])
+GR_REGISTER_BLOCK("gr::blocks::math::Xor", gr::blocks::math::MathOpImpl, ([T], std::bit_xor<[T]>), [ uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, float, double, gr::UncertainValue<float>, gr::UncertainValue<double>, std::complex<float>, std::complex<double> ])
```
jsallay: I think that the bit level operations should only be defined for `std::integral` types.

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -115,6 +125,221 @@ using Multiply = MathOpMultiPortImpl<T, std::multiplies<T>>;
...
+};
+
+template<typename T>
+using Max = MathOpMultiPortImpl<T, max<T>>;
+
+template<typename T>
+using Min = MathOpMultiPortImpl<T, min<T>>;
+
+GR_REGISTER_BLOCK("gr::blocks::math::Negate", gr::blocks::math::MathOpImpl, ([T], std::negate<[T]>), [ uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, float, double, gr::UncertainValue<float>, gr::UncertainValue<double>, std::complex<float>, std::complex<double> ])
+GR_REGISTER_BLOCK("gr::blocks::math::Not", gr::blocks::math::MathOpImpl, ([T], std::bit_not<[T]>), [ uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, float, double, gr::UncertainValue<float>, gr::UncertainValue<double>, std::complex<float>, std::complex<double> ])
```
jsallay: Again, I think it should be only defined for `std::integral` types.

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -115,6 +125,221 @@ using Multiply = MathOpMultiPortImpl<T, std::multiplies<T>>;
...
+GR_REGISTER_BLOCK("gr::blocks::math::Abs", gr::blocks::math::MathOpImpl, ([T], std::abs_op<[T]>), [ uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, float, double ])
```
jsallay: I know that the GR 3 block doesn't work for complex types, but your `abs_op` function below does.  I would complex types in.

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -115,6 +125,221 @@ using Multiply = MathOpMultiPortImpl<T, std::multiplies<T>>;
...
+/* ------------------------------------------------------------------ *
+ *  Integrate : running sum over N samples, output 1 value, decimate   *
+ * ------------------------------------------------------------------ */
+
+ GR_REGISTER_BLOCK("gr::blocks::math::Integrate",
+                  gr::blocks::math::Integrate,
+                  ([T]),
+                  [ uint8_t, uint16_t, uint32_t, uint64_t,
+                    int8_t,  int16_t,  int32_t,  int64_t,
+                    float, double ])
```
jsallay: Can we add complex types here as well.  The GR block supports `std::complex<float>`

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -115,6 +125,221 @@ using Multiply = MathOpMultiPortImpl<T, std::multiplies<T>>;
...
+        const auto dec     = decim.value;
+
+        for (auto v : ins) {
+            _acc += v;
+            if (++_seen == dec) {
+                *outIt++ = _acc;
+                _acc  = T(0);
+                _seen = 0;
+            }
+        }
```
jsallay: I think that you can use a `chunk_view` (https://en.cppreference.com/w/cpp/ranges/chunk_view.html) and `std::accumulate` to iterate over this more cleanly.

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -115,6 +125,221 @@ using Multiply = MathOpMultiPortImpl<T, std::multiplies<T>>;
...
+              Visible,
+              Doc<"decimation / integration length">,
+              Limits<1UZ, (1UZ << 20)> >    // ← space before last '>'
+        decim = 1UZ;
+
+    GR_MAKE_REFLECTABLE(Integrate, in, out, decim);
+
+    /* ---------- State --------------------------------------------- */
+    T           _acc  = T(0);
+    std::size_t _seen = 0;
```
jsallay, line 278: I'm not sure what the `_seen` variable is doing,  Looking at the code it looks like it is always 0.  Could you explain it?

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -115,6 +125,221 @@ using Multiply = MathOpMultiPortImpl<T, std::multiplies<T>>;
...
+    /* ---------- Ports & settings ---------------------------------- */
+    PortIn<T>  in;
+    PortOut<T> out;
+
+    Annotated<gr::Size_t,
+              "decim",
+              Visible,
+              Doc<"decimation / integration length">,
+              Limits<1UZ, (1UZ << 20)> >    // ← space before last '>'
+        decim = 1UZ;
```
jsallay: I would double check with @RalphSteinhagen .  I believe (but I'm struggling to remember how to do it) that there is a way to set up the block to use the decimation as a block size.  Which would guarantee that process_bulk always has an integer number of blocks to process.

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -115,6 +125,221 @@ using Multiply = MathOpMultiPortImpl<T, std::multiplies<T>>;
...
+            auto        max_val = *in_it;          // first element as initial max
+
+            for (std::size_t i = 1; i < L; ++i) {
+                const auto& v =
+                    *(in_it + static_cast<std::ptrdiff_t>(i));
+                if (v > max_val) {
+                    max_val = v;
+                    max_idx = i;
+                }
+            }
```
jsallay: Consider using `std::max_element` with `std::distance` to find the arg max.

jsallay (changes requested):
There is a lot of good stuff here.  My comments are for pretty minor things.  I would also add to double check your test cases match the supported types of the blocks.  I think that for example, if you tested `xor` with `std::complex` that it would fail to compile.

blocks/math/include/gnuradio-4.0/math/Math.hpp
KrxGu: it is not always zero: it counts samples across successive processBulk calls until decim is reached, then resets. Hence it's required to keep the code logic intact.

blocks/math/include/gnuradio-4.0/math/Math.hpp
KrxGu: okay so the header file still builds without std::complex<T> , but kindly confirm that u only need it to be removed from Max/Min header file right?

blocks/math/include/gnuradio-4.0/math/Math.hpp
KrxGu: gotcha, also i would then Tighten the concept in the alias 
so basically istead of
`template<typename T>
using And = MathOpMultiPortImpl<T, std::bit_and<T>>;`
it will be something like 
`template<std::integral T>
using And = MathOpMultiPortImpl<T, std::bit_and<T>>;`

same will be done for Or, Xor, and the single-port Not. should i ?

blocks/math/include/gnuradio-4.0/math/Math.hpp
KrxGu: okay, but i had a doubt what's bad in covering the possibilities of other data types cases too like float, double , complex ,etc.?

blocks/math/include/gnuradio-4.0/math/Math.hpp
KrxGu: gotcha!

blocks/math/include/gnuradio-4.0/math/Math.hpp
KrxGu: tried 2-3 different approach to do so, but the build is failing. So, we can discuss it for further porting of next header files later

blocks/math/include/gnuradio-4.0/math/Math.hpp
KrxGu: @RalphSteinhagen kindly help us with this

blocks/math/include/gnuradio-4.0/math/Math.hpp
KrxGu: Gotcha!! cleaner

blocks/math/include/gnuradio-4.0/math/Math.hpp
KrxGu: will keep it mind porting next header files, however not necessary for this PR i think.

i think it would look something like
`for (auto chunk : ins | std::views::chunk(decim.value)) {
    *outIt++ = std::accumulate(chunk.begin(), chunk.end(), T{0});
}
`

KrxGu (commented):
addressed

blocks/math/include/gnuradio-4.0/math/Math.hpp
jsallay: Yes.  You made the changes I was expecting.

blocks/math/include/gnuradio-4.0/math/Math.hpp
jsallay: For negate, you are correct.  We should cover the floating point/complex types.  Sorry about my mistake.  Can you add the types back in.

For Not, when dealing with floating point numbers, there can be confusing behavior.  When converting an integer to bool, it is easy to understand and consistent on all systems.  When converting a floating point number, a number can be really close to 0 without being 0.  (Or if the floating point had infinite precision, it would be 0, but with finite precision it is not).  It generally preferred that a user handle these cases such as:
```
bool x;
float y = ...;
if (std::abs(y) < 1e-6) x = false;
else x = true;
```

blocks/math/include/gnuradio-4.0/math/Math.hpp
mormj: @KrxGu - can you elaborate on what is failing in the build with complex?

blocks/math/include/gnuradio-4.0/math/Math.hpp
KrxGu: makes sense, will undo it for negate.

jsallay (approved):

jsallay:
I think the code is in good shape, but we need to get the CI jobs passing now.  Look compiler errors.

blocks/math/test/qa_Math_ext.cpp
```diff
@@ -0,0 +1,233 @@
+/*
+ * qa_Math_ext.cpp  ── “extended” QA for math blocks.   Split out of the
+ * original qa_Math.cpp so that large CI jobs can be parallelised.
+ */
```
RalphSteinhagen: This is not correct... the parallelisation is controlled via the plugin generator's `--split` flag. No need to split the definition files.

RalphSteinhagen: N.B., without `--split`, the generator will create one compilation unit per block template type (x N types).
With `--split`, the generator will create one compilation unit per block type instantiation.

On GitHub, this split isn't enabled because the normal runner has only two cores available; thus, the parallelisation is meager (i.e., `-j 2` equivalent).

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -115,6 +138,277 @@ using Multiply = MathOpMultiPortImpl<T, std::multiplies<T>>;
...
+                  gr::blocks::math::Log10,
+                  ([T]),
+                  [ float, double ])
+
+template<typename T>
+requires std::is_floating_point_v<T>
+struct Log10 : gr::Block<Log10<T>>
+{
+    using Description = Doc<R""(
+@brief Compute \(n \cdot \log_{10}(|x|) + k\)
```
RalphSteinhagen: This is probably not what you want. This creates an empty line on the description before the '@brief'.

blocks/math/include/gnuradio-4.0/math/Math.hpp
RalphSteinhagen: Have a look at this [example](https://github.com/fair-acc/gnuradio4/blob/5c36c55d2c0fe8fb17fe1bbb6bebb4fec8c9abb8/core/test/qa_Block.cpp#L370-L392) implementation and corresponding [unit-test](https://github.com/fair-acc/gnuradio4/blob/5c36c55d2c0fe8fb17fe1bbb6bebb4fec8c9abb8/core/test/qa_Block.cpp#L547-L571).

blocks/math/test/qa_Math_core.cpp
```diff
@@ -0,0 +1,125 @@
+/*
+ * Core math-block QA – covers the blocks that were present _before_
+ * the 2025-06 extensions (Add/Sub/Multiply/Divide and their *Const
+ * variants).  Splitting the test lowers the peak memory GCC needs.
+ */
+
```
RalphSteinhagen: Please do not keep what seesm to be journal- or decorator-like comments: 
strive for good naming of blocks/methods/variables, conciseness and let the code speak for itself.

As yourself the question: is this comment functionally absolutely needed or provides additional information beyond what the code implements? To be kept in mind: code is read more often than written.

blocks/math/test/qa_Math_ext.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+#include <boost/ut.hpp>
+
+// We avoid <format> because wasm/libc++ on CI doesn’t ship it.
```
RalphSteinhagen: ??? this has been part of the `libc++` (which WASM/Clang uses since long).

blocks/math/test/qa_Math_ext.cpp
```diff
@@ -0,0 +1,233 @@
...
+    constexpr auto kArith = std::tuple<
+        uint8_t,uint16_t,uint32_t,uint64_t,
+        int8_t,int16_t,int32_t,int64_t,
+        float,double>();
+
+    constexpr auto kLogic = std::tuple<
+        uint8_t,uint16_t,uint32_t,uint64_t,
+        int8_t,int16_t,int32_t,int64_t>();
+
+    /* ---------- Max / Min ----------------------------------------- */
```
RalphSteinhagen: you can embed the code in `"Max / Min subtests"_test = [] {...}"`. This would make this code more expressive and avoids the comments.

blocks/math/include/gnuradio-4.0/math/Math.hpp
```diff
@@ -115,6 +138,277 @@ using Multiply = MathOpMultiPortImpl<T, std::multiplies<T>>;
...
+                  ([T], gr::blocks::math::abs_op<[T]>),
+                  [ uint8_t, uint16_t, uint32_t, uint64_t,
+                    int8_t,  int16_t,  int32_t,  int64_t,
+                    float,   double ])
+
+template<typename T, typename op>
+requires(std::is_arithmetic_v<T>)
+struct MathOpSinglePortImpl : public gr::Block<MathOpSinglePortImpl<T, op>> {
+    using Description = Doc<R""(
+    @brief Math block transforming a single input to a single output with a given operation
```
RalphSteinhagen: Why the additional new line?

RalphSteinhagen:
@KrxGu 

image

Please rebase and bump.

Also, please keep the code together, structured and as concise as possible. Splitting the math header files isn't a good optimisation. See my other comments above. You are nearly there ...

blocks/math/test/qa_Math_ext.cpp
KrxGu: > This is not correct... the parallelisation is controlled via the plugin generator's `--split` flag. No need to split the definition files.

Splitting the qa_Math.cpp was done yesterday before that it was intact and in very good shape but the problem was  some CI tests were failing for test file and i reported the issue to @jsallay and when he tried to debug the CI fail by reproducing them locally he faced OOM issue hence he suggested to split the qa_Math.cpp . We are actively trying to debug the issue with the CI actions. I will first try to solve the conflict in the merge so that CI actions start running on the split files and then if that doesn't work will bring back qa_Math.cpp , is that approach correct or should i just bring back qa_Math.cpp and then try to figure out why they are failing.

blocks/math/test/qa_Math_core.cpp
KrxGu: **Gotcha!!** will take care. Thank you for the suggestions.

blocks/math/include/gnuradio-4.0/math/Math.hpp
KrxGu: Gotcha!! will have a look

blocks/math/test/qa_Math_ext.cpp
KrxGu: before, my code followed the legacy <format> but while debugging CI tests errors @jsallay and i came to a conclusion that <format> might be the reason for failing CI tests hence removed it, however i am still open to use <format> if CI actions doesn't fail.

blocks/math/test/qa_Math.cpp
KrxGu, line 1: @RalphSteinhagen kindly have a look how qa_math.cpp looked before me and @jsallay started to debug the failing CI tests

mormj:
Closing to pursue integrating into gr4-incubator for further refinement

PR #596, opened 2025-06-13
core: add .pc and update install targets

mormj:
The goal of these changes is to be able to create a standalone install prefix for gnuradio4 such that an OOT can be defined without gnuradio4 as a subproject.

There will be more steps involved, but this is the minimal amount of changes that I needed to compile a basic OOT block in a separate repo while finding the appropriate include and lib using the installed .pc file

Ideally a follow on PR will install generated cmake files that publish all the targets

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -293,6 +293,10 @@ if(NOT (TARGET magic_enum))
   add_library(magic_enum INTERFACE)
   target_include_directories(magic_enum ${CMAKE_EXT_DEP_WARNING_GUARD}
                              INTERFACE ${PROJECT_SOURCE_DIR}/third_party/magic_enum/)
+
+  set(magic_enum_public_headers third_party/magic_enum/magic_enum.hpp third_party/magic_enum/magic_enum_utility.hpp)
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -274,6 +274,17 @@ if(NOT (TARGET magic_enum))
...
+  set(magic_enum_public_headers
+    third_party/magic_enum/magic_enum.hpp
+    third_party/magic_enum/magic_enum_utility.hpp
+  )                             
+  set_target_properties(magic_enum
+                      PROPERTIES PUBLIC_HEADER "${magic_enum_public_headers}"
+                      )
+  install(
+    TARGETS magic_enum
+    PUBLIC_HEADER DESTINATION include)
```
wirew0rm: For now this is probably not a problem since most users will not install this to /usr/lib but to some custom prefix, but once this is used for packaging for distributions we can probably not just globally install third_party headers. Need to look into how this is done in other projects. Prefixing, relying on upstream packaging of the headers, something else?

wirew0rm (approved):
Thanks for getting this started, the changes look good. I also expect that this will need some refinement and follow-ups, but these are easier once the foundations are there and it is exposed to external usage.

Issue #585, opened 2025-05-23
win32: qa_AtomicBitset fails due to win32 unavailability of aborts


chrisjohgorman:
Here again I have had to edit qa_AtomicBitset.cpp and exclude a bit of code due to boost/ut's aborts() not being available in Windows.  I'm not sure this is the reason for failure, but I suspect it probably is.  There is a conditional block  I had to add `&& not defined(_WIN32)` to.  The block follows.

```
#if not defined(__EMSCRIPTEN__) && not defined(NDEBUG) && not defined(_WIN32)
    expect(aborts([&] { bitset.set(bitsetSize); })) << "Setting bit should throw an assertion.";
    expect(aborts([&] { bitset.reset(bitsetSize); })) << "Resetting bit should throw an assertion.";
    expect(aborts([&] { bitset.test(bitsetSize); })) << "Testing bit should throw an assertion.";
    // bulk operations
    expect(aborts([&] { bitset.set(100UZ, 200UZ); })) << "Setting bulk bits should throw an assertion.";
    expect(aborts([&] { bitset.reset(100UZ, 200UZ); })) << "Resetting bulk bits should throw an assertion.";
    expect(aborts([&] { bitset.set(200UZ, 100UZ); })) << "Setting bulk begin > end should throw an assertion.";
    expect(aborts([&] { bitset.reset(200UZ, 100UZ); })) << "Resetting bulk begin > end should throw an assertion.";
#endif
```
I'm attaching the output from qa_AtomicBitset.exe as [qa_AtomicBitset.txt](https://github.com/user-attachments/files/20419091/qa_AtomicBitset.txt).  I don't know if it's useful, given the code edit, but it might give someone in the know a good idea on what's wrong with the program.

chrisjohgorman:
Hi again @RalphSteinhagen.  I  have fixed the bug in my tree, which is why I closed it.  I guess it was a little premature as I haven't submitted my tree as a PR yet.  I would like some feedback on my proposed changes before submitting it as a PR, if you or someone else has a moment to look at my tree and see if anything needs to be addressed, it would be greatly appreciated.  It can be found at https://github.com/chrisjohgorman/gnuradio4/tree/chrisjohgorman-build-ucrt64-win32 .  If my tree looks good I'll submit it for inclusion in main.

PR #597, opened 2025-06-15
win32: fix for qa_AtomicBitset issue #585

chrisjohgorman:
- exclude attempts to use boost::ut's aborts call in qa_AtomicBitset.cpp on win32.
- fixed AtomicBitset.hpp to use ULL instead of UL on win32 as std::size_t is unsigned long long on win32.

### DIFF
--- a/core/include/gnuradio-4.0/AtomicBitset.hpp
+++ b/core/include/gnuradio-4.0/AtomicBitset.hpp
@@ -35,7 +35,7 @@ public:
     requires(!isSizeDynamic)
     {
         for (auto& word : _bits) {
-            word.store(0UL, std::memory_order_relaxed);
+            word.store(0UZ, std::memory_order_relaxed);
         }
     }
 
@@ -44,7 +44,7 @@ public:
         : _size(size), _bits(std::vector<std::atomic<std::size_t>>(size)) {
         // assert(size > 0UZ);
         for (std::size_t i = 0; i < _size; i++) {
-            _bits[i].store(0UL, std::memory_order_relaxed);
+            _bits[i].store(0UZ, std::memory_order_relaxed);
         }
     }
 
@@ -52,7 +52,7 @@ public:
         assert(bitPosition < _size);
         const std::size_t wordIndex = bitPosition / _bitsPerWord;
         const std::size_t bitIndex  = bitPosition % _bitsPerWord;
-        const std::size_t mask      = 1UL << bitIndex;
+        const std::size_t mask      = 1UZ << bitIndex;
 
         std::size_t oldBits;
         std::size_t newBits;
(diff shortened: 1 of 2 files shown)


core/include/gnuradio-4.0/AtomicBitset.hpp
```diff
@@ -35,7 +35,11 @@ class AtomicBitset {
     requires(!isSizeDynamic)
     {
         for (auto& word : _bits) {
+#if defined(_WIN32)
+            word.store(0ULL, std::memory_order_relaxed);
```
RalphSteinhagen: Correct portability bug identified (i.e. `std::size_t` has platform-specific backing types), but rather than using an if-def macro, we should use the `UZ` instead of the `UL` or `ULL` literal.

core/test/qa_AtomicBitset.cpp
```diff
@@ -72,7 +72,7 @@ void runAtomicBitsetTest(TBitset& bitset, std::size_t bitsetSize) {
         }
     }
 
-#if not defined(__EMSCRIPTEN__) && not defined(NDEBUG)
+#if not defined(__EMSCRIPTEN__) && not defined(NDEBUG) && not defined(_WIN32)
```
RalphSteinhagen, line 75: For Emscripten (WASM), this test is (IIRC) suppressed because exceptions are usually not enabled out of the box in WASM code.

Is this suppression also necessary for Windows?

RalphSteinhagen (changes requested):
@chrisjohgorman thanks for identifying the issue and your PR.

As indicated above: maybe through using the `UZ` literal could avoid having the if-else compile-time macro. If you could change and test that. Much appreciated!

Otherwise OK and I'd merge it afterwards. Thanks again!

core/include/gnuradio-4.0/AtomicBitset.hpp
RalphSteinhagen: *N.B. for info: the original code was written for C++17/20 and older gcc/clang versione where `UZ` wasn't available yet.*

core/test/qa_AtomicBitset.cpp
chrisjohgorman: Hi @RalphSteinhagen we do need to exclude boost::ut::aborts() function on win32.  On windows there is no easy way to handle a SIGABRT from a child process.  We could throw and catch exceptions, but we can't rely on boost::ut::aborts() to fork() and capture POSIX signals as it's just not implemented.

We could do something like ...
```
try 
{ 
  bitset.set(bitsetSize); // Setting bit should throw an assertion
}
catch (const std::exception& e) {
  std::println("caught exception {}", e.what());
}
```

chrisjohgorman: I didn't note this in my first reply, but I made the requested changes.  All instances of using ULL / UL was replaced by UZ.  I tested the results on both Windows and Linux and they work.  Let me know if you require further changes.

RalphSteinhagen: Fine for the time being! :+1:

RalphSteinhagen:
@chrisjohgorman could you fill the DCO in your commit. See the explaination [here](https://github.com/fair-acc/gnuradio4/pull/597/checks?check_run_id=44198725193).

Will merge this afterwards.

RalphSteinhagen:
@all-contributors please add @chrisjohgorman  for code and platform

chrisjohgorman:
@RalphSteinhagen sorry for missing the DCO.  I'm going to set git up to do it by default.  Thanks for all the help.

Issue #584, opened 2025-05-23
win32: qa_ImChart.exe doesn't appear to draw with the right font


chrisjohgorman:
Hello All,

First off let me say it's really nice how the tty's have graphs in them like the ones drawn by qa_ImChart.  On Windows the test cases pass, but the font or characters drawn are wrong.

Looking at the first graph of qa_ImChart, all the bits are there.  The color of the graph are correct, the x and y axes are there and the numbers are there.  But...

The lines aren't drawn properly, what should be the x-axis, eg.
```
 ├─────────────┬─────────────┬
```
turns out to be 
```
Γö£ΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓöÇΓ
```
The sine and cosine curves are there, but in the same way the x-axis isn't drawn properly, the braille font used to draw the curves shows up as ...
```
ΓóÇΓúáΓáñΓáñΓúäΓíÇ
```
instead of ...
```
⢀⡴⠚⠙⠲⢤
```
When the curves are drawn using 'X' characters, that is drawn well.

The y-axis suffers from the same problem as the x-axis.  I'm including a text file showing the output of qa_ImChart.exe.  [qa_ImChart.txt](https://github.com/user-attachments/files/20418023/qa_ImChart.txt)

I'm wondering if I could fix this by specifying a specific font in windows?

RalphSteinhagen:
Hi @chrisjohgorman — thanks a lot for all your Win32-related reports! These are genuinely helpful, and we appreciate you taking the time to document and share them. 👍

As you may have guessed, Windows hasn’t been our primary focus so far. We're a small team and don’t have any dedicated Windows development expertise in-house. That said, we absolutely don’t want to discourage contributions in this area — quite the opposite. Your input is helping to improve on this, and we’d love to see them tackled with community help.

Would it make sense to collect the various Win32-specific bugs into a single "Windows Compatibility EPIC" issue to keep things manageable? That way, we could list known issues and link individual PRs or sub-issues as fixes come in. It would also help keep Win32-specific discussions separate from cross-platform or toolchain-related topics (e.g. GCC, Clang, Emscripten).

Let us know if you have a better idea for structuring this or if you know other Windows developers 👷 who might be interested in helping out. @mormj @dkozel, do you have any thoughts?

[...]

PR #601, opened 2025-06-19
Added support for ImChart.hpp in windows

chrisjohgorman:
- added windows.h include to ImChart.hpp
- added win32 call to SetConsoleOutputCP(UTF-8) to cpp programs built with ImChart.hpp
- added setlocale(C.UTF8) call to example_ImChart.cpp
- changed UL size specification to UZ, for std::min, std::max and std::clamp calls
- fixes https://github.com/fair-acc/gnuradio4/issues/584

### DIFF
--- a/blocks/testing/test/qa_UI_Integration.cpp
+++ b/blocks/testing/test/qa_UI_Integration.cpp
@@ -11,6 +11,13 @@
 #include "gnuradio-4.0/testing/ImChartMonitor.hpp"
 #include "gnuradio-4.0/testing/TagMonitors.hpp"
 
+#ifdef _WIN32
+[[maybe_unused]] const auto _ = [] {
+    SetConsoleOutputCP(CP_UTF8);
+    return 0;
+}();
+#endif
+
 const boost::ut::suite TagTests = [] {
     using namespace boost::ut;
     using namespace gr;
--- a/blocks/testing/test/qa_UI_Integration.cpp
+++ b/blocks/testing/test/qa_UI_Integration.cpp
@@ -8,15 +8,9 @@
 #include "gnuradio-4.0/algorithm/ImChart.hpp"
 #include "gnuradio-4.0/basic/ClockSource.hpp"
 #include "gnuradio-4.0/basic/FunctionGenerator.hpp"
+#include "gnuradio-4.0/meta/UnitTestHelper.hpp"
 #include "gnuradio-4.0/testing/ImChartMonitor.hpp"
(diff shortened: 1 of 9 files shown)


blocks/testing/test/qa_UI_Integration.cpp
```diff
@@ -11,6 +11,13 @@
 #include "gnuradio-4.0/testing/ImChartMonitor.hpp"
 #include "gnuradio-4.0/testing/TagMonitors.hpp"
 
+#ifdef _WIN32
+[[maybe_unused]] const auto _ = [] {
+    SetConsoleOutputCP(CP_UTF8);
+    return 0;
+}();
+#endif
+
```
RalphSteinhagen: @chrisjohgorman I wonder whether this could be placed into the common [UnitTestHelper.hpp](https://github.com/fair-acc/gnuradio4/blob/main/meta/include/gnuradio-4.0/meta/UnitTestHelper.hpp) file and included in each unit test. This would avoid the platform-specific config duplication. 

Also, maybe add an `inline` to avoid ODR issues later.

What do you think?

*P.S. I restarted the clang CI check. I think this might be an unrelated flaky test.*

RalphSteinhagen (approved):
@chrisjohgorman, thanks for the fixes. They look good to me. 

See whether my minor (perhaps improvement) suggestion above helps. Otherwise, this could be merged as-is once the (flaky!?) test passes.

blocks/testing/test/qa_UI_Integration.cpp
chrisjohgorman: Hi @RalphSteinhagen thank you for retesting the clang CI.  I got notification that the clang18 | Release passed, so you're right about the test case failure being unrelated to the proposed commit.  I moved the `#ifdef` to `UnitTestHelper.hpp` and included it in the `.cpp` files.  The only exception to this is `example_ImChart`, which I left as it was.  I think your changes were a good suggestion, and it helped me explore more of the code base, since I wasn't aware of `UnitTestHelper.hpp` before today.  Let me know if this is what you had in mind or if there are some places that could be improved.  Thanks again for your time and help.

RalphSteinhagen: @chrisjohgorman excellent! Your help and improvements are much appreciated!

Sorry for not getting back to you sooner. As you mentioned, the way you approach things is very good way to get acquainted with the code base. I'd wish/hope more will follow your example! :+1:

PR #603, opened 2025-06-26
updated bm_Scheduler

RalphSteinhagen:
* added `SimCompute<T>` block to simulate compute delay and model latency effects for a given complexity order
 * added unit-tests for Null[Source, Sink], Counting[Sink, Source], ...
 * renamed _jobList to _executionOrder (better naming)
 * corrected latency statistic print-out in benchmark.hpp
 * added DFS scheduler, and added benchmark added cases
   * A - linear graph
   * B - forked graph
   * C - split graph
 * fixed bug by removing _blockList class member field in BFS and DFS scheduler (block list got duplicated on each init())
 * added property_map constructor to SchedulerBase
 
 Example output (N.B. 'first-out -> first-in' latency):
 
image
image
image

### DIFF
--- a/core/include/gnuradio-4.0/Scheduler.hpp
+++ b/core/include/gnuradio-4.0/Scheduler.hpp
@@ -96,7 +96,7 @@ protected:
         doForNestedBlocks(doForNestedBlocks, _graph);
     }
 
-    void registerPropertyCallBacks() noexcept {
+    void registerPropertyCallbacks() noexcept {
         this->propertyCallbacks[scheduler::property::kEmplaceBlock] = std::mem_fn(&SchedulerBase::propertyCallbackEmplaceBlock);
         this->propertyCallbacks[scheduler::property::kRemoveBlock]  = std::mem_fn(&SchedulerBase::propertyCallbackRemoveBlock);
         this->propertyCallbacks[scheduler::property::kRemoveEdge]   = std::mem_fn(&SchedulerBase::propertyCallbackRemoveEdge);
@@ -124,12 +124,12 @@ public:
 
     explicit SchedulerBase(gr::Graph&& graph, std::string_view defaultPoolName, const profiling::Options& profiling_options = {}) //
         : _graph(std::move(graph)), _profiler{profiling_options}, _profilerHandler{_profiler.forThisThread()}, _pool(gr::thread_pool::Manager::instance().get(defaultPoolName)) {
-        registerPropertyCallBacks();
+        registerPropertyCallbacks();
     }
 
     explicit SchedulerBase(gr::Graph&& graph, property_map initialSettings, std::string_view defaultPoolName = gr::thread_pool::kDefaultCpuPoolId) //
         : base_t(initialSettings), _graph(std::move(graph)), _profiler{{}}, _profilerHandler{_profiler.forThisThread()}, _pool(gr::thread_pool::Manager::instance().get(defaultPoolName)) {
-        registerPropertyCallBacks();
+        registerPropertyCallbacks();
         std::ignore = this->settings().set(initialSettings);
         std::ignore = this->settings().activateContext();
         std::ignore = this->settings().applyStagedParameters();
@@ -990,7 +990,7 @@ private:
         // depth first traversal — use stack
         std::set<block_t> visited;
 
-        std::function<void(block_t)> dfs = [&](block_t node) {
+        auto dfs = [&]<typename Self>(Self&& self, block_t node) -> void {
             if (visited.contains(node)) {
                 return;
             }
@@ -1000,14 +1000,14 @@ private:
 
             if (_adjacency_list.contains(node)) {
                 for (auto& dst : _adjacency_list.at(node)) {
-                    dfs(dst);
+                    self(self, dst); // recursive call
                 }
             }
         };
 
         // launch from all source blocks
         for (auto sourceBlock : _source_blocks) {
-            dfs(sourceBlock);
+            dfs(dfs, sourceBlock);
         }
 
         // batching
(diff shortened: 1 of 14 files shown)


core/test/qa_Scheduler.cpp
```diff
@@ -704,11 +704,6 @@ const boost::ut::suite<"SchedulerTests"> SchedulerTests = [] {
         scheduler.timeout_ms               = 100U; // also dynamically settable via messages/block interface
         scheduler.timeout_inactivity_count = 10U;  // also dynamically settable via messages/block interface
 
-        expect(scheduler.graph().reconnectAllEdges());
```
ivan-cukic, line 707: Why are these checks removed?

core/test/qa_SchedulerMessages.cpp
```diff
@@ -25,13 +25,23 @@ TestContext* context = new TestContext(paths{}, // plugin paths
...
-public:
-    TestScheduler(gr::Graph graph) : scheduler_(std::move(graph)) {
-        using namespace gr::testing;
+    using TScheduler = gr::scheduler::Simple<gr::scheduler::ExecutionPolicy::singleThreaded>;
+    std::expected<void, gr::Error> schedulerRet_;
+    std::thread                    schedulerThread_;
 
-        scheduler_.graph().emplaceBlock<gr::testing::SlowSource<float>>();
-        scheduler_.graph().emplaceBlock<gr::testing::CountingSink<float>>();
+    gr::Graph& addSourceSink(gr::Graph& graph) const noexcept {
```
ivan-cukic: `gr::Graph&& addSourceSink(gr::Graph&& graph) const noexcept {` (rvalue references)

Also, the name could have been nicer -- `withTestingSourceAndSink` or similar. `addSourceSink` sounds like it adds a source-sink :)

core/test/qa_SchedulerMessages.cpp
```diff
@@ -188,18 +189,18 @@ const boost::ut::suite TopologyGraphTests = [] {
     };
 
     "Block removal tests"_test = [] {
-        TestScheduler scheduler(gr::Graph(context->loader));
-
-        auto& testGraph = scheduler.graph();
+        gr::Graph graph = gr::Graph(context->loader);
```
ivan-cukic: `gr::Graph graph(context->loader);`

core/test/qa_SchedulerMessages.cpp
```diff
@@ -449,15 +449,15 @@ const boost::ut::suite MoreTopologyGraphTests = [] {
     using namespace gr::testing;
     using enum gr::message::Command;
 
-    gr::scheduler::Simple scheduler{gr::Graph(context->loader)};
+    gr::Graph graph  = gr::Graph(context->loader);
```
ivan-cukic: Same as above

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -96,29 +96,43 @@ class SchedulerBase : public Block<Derived> {
         doForNestedBlocks(doForNestedBlocks, _graph);
     }
 
+    void registerPropertyCallBacks() noexcept {
```
ivan-cukic: lowercase B for Call**b**acks as in other parts of the code.

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -129,9 +143,12 @@ class SchedulerBase : public Block<Derived> {
             }
         }
         waitDone();
-        _jobLists.reset(); // force earlier crashes is accessed after destruction
+        _executionOrder.reset(); // force earlier crashes if this is accessed after destruction (e.g. from thread that was kept running)
     }
 
+    [[nodiscard]] const gr::Graph& graph() const noexcept { return _graph; }
```
ivan-cukic, line 149: Should we have a non-const getter as well?

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -896,20 +928,108 @@ detecting cycles and blocks which can be reached from several source blocks.)"">
...
+            _source_blocks.push_back(e._sourceBlock);
+            block_reached.insert(e._destinationBlock);
+        }
+
+        _source_blocks.erase(std::remove_if(_source_blocks.begin(), _source_blocks.end(), [&block_reached](auto currentBlock) { return block_reached.contains(currentBlock); }), _source_blocks.end());
+
+        // depth first traversal — use stack
+        std::set<block_t> visited;
+
+        std::function<void(block_t)> dfs = [&](block_t node) {
```
ivan-cukic: We can now use `this auto self` for recursive lambdas for better speed than std::function. (or, this can be a Y combinator-inspired function and take itself as the first argument for older C++ compilers).

core/test/qa_Scheduler.cpp
RalphSteinhagen: For life-time reasons, the `scheduler.graph() -> const Graph&` returns a const-reference that only the scheduler can (and should) execute/change. Access to the interna is neat for testing but opens the door for unsafe user-level API access. I thus changed that.

core/test/qa_SchedulerMessages.cpp
RalphSteinhagen: Done.

core/include/gnuradio-4.0/Scheduler.hpp
RalphSteinhagen: I think not. It would be unsafe to add/remove blocks and edges outside the scheduler knowing about this. I thus made this explicitely a const-reference to harden the user-level API against unsafe use.

PR #605, opened 2025-07-02
Thread instrumentation

RalphSteinhagen:
* added thread name diagnostics
* added atomic blocking + watchdog to scheduler


core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -56,7 +56,7 @@ class SchedulerBase : public Block<Derived> {
     TProfiler                           _profiler;
     decltype(_profiler.forThisThread()) _profilerHandler;
     std::shared_ptr<TaskExecutor>       _pool;
-    std::atomic_size_t                  _nRunningJobs{0UZ};
+    std::shared_ptr<std::atomic_size_t> _nRunningJobs = std::make_shared<std::atomic_size_t>(0UZ);
```
wirew0rm: I'm not sure i understand this correctly. The original purpose of the _nRunningJobs atomic was to make sure that the Scheduler stays alive for the whole lifetime of all jobs, so the jobs could directly access `_nRunningJobs` (and `_Progress`). There is a call to `waitDone` in the scheduler's destructor, which should block until all jobs are done.

Is this because the old logic did not work and produced invalid reads or because we now expect these values to be accessed from outside of the scheduler controlled jobs?

My feeling is that this introduces a bit of a weird in-between state, where the scheduler is gone, but some threads stilll need(?) to access a select set of scheduler members.
I also see the danger of that set of members expanding in the future, making it even more difficult to argue about lifetimes. At some point it might be cleaner to just make the whole scheduler shared.

Or might be that I just need a bit more explanation.

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -438,14 +448,50 @@ class SchedulerBase : public Block<Derived> {
...
+            }
+            counter = 0; // check and increase progress if there hasn't been none.
+
+            std::size_t progress = localProgress->value();
+            if (progress == currentProgress) {
+                nWarnings++;
+                currentProgress = localProgress->incrementAndGet(); // watchdog triggered manual update
+                localProgress->notify_all();
+                if (nWarnings >= timeOut_count) {
+                    std::println("trigger watchdog update {} of {} in {}", nWarnings, timeOut_count, this->unique_name);
```
wirew0rm, line 511: If I have a flowgraph that is expected to not make progress over longer periods of time, is there a better way to disable this than setting the timouts and/or counter tresholds to huge values?

Sidenote: at least from grc we probably cannot set the top-level scheduler's settings, right? Not for this PR but should be kept in mind if we add more settings here that at some point we or someone else is going to want to change them.

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -345,25 +346,34 @@ class SchedulerBase : public Block<Derived> {
...
-                _nRunningJobs.wait(0UZ, std::memory_order_acquire); // waits until at least one pool worker started
+                _nRunningJobs->wait(0UZ, std::memory_order_acquire); // waits until at least one pool worker started
             }
         }
     }
 
     void poolWorker(const std::size_t runnerID, std::shared_ptr<std::vector<std::vector<BlockModel*>>> jobList) noexcept {
-        _nRunningJobs.fetch_add(1UZ, std::memory_order_acq_rel);
-        _nRunningJobs.notify_all();
+        std::shared_ptr<gr::Sequence>       localProgress = _graph._progress; // life-time guaranteed
```
wirew0rm: Can i suggest dropping the `local` here from the name? To me this says that we have some different copy of progress in this local context, when actually this is a reference to the scheduler-global value. 
Open for different suggestions as well.

```suggestion
        std::shared_ptr<gr::Sequence>       progress = _graph._progress; // life-time guaranteed
```

core/include/gnuradio-4.0/Scheduler.hpp
RalphSteinhagen: The `_nRunningJobs`and `_Progress` need to be shared_ptr because they need to outlive the poolWorker and scheduler that may terminate/destroy before the -- for example -- watchdog is shut down. Similary for when exchanging graphs (second-next PR). This adds a bit more thread safety.

core/include/gnuradio-4.0/Scheduler.hpp
RalphSteinhagen: This is needed because otherwise, no messages will be processed either. We have a design where only one thread handles messages (and part of) the graph work execution, and that thread may be put on hold waiting for the atomic operation. 

Contrary to mutex, atomic waits cannot time out. The watchdog is a soft time-out to allow the poolWorker loops to regularly check if there are other pending scheduler messages.

Issue #587, opened 2025-05-24
win32: qa_KnownSharedLibBlocks failure with gr::electrical::PowerMetrics call


chrisjohgorman:
This one's a bit clearer to me. `qa_KnownSharedLibBlocks.exe` fails unless I comment line 79 out.  The statement ...
```
expect(registry.contains("gr::electrical::PowerMetrics<float32, 3ul>"sv));
```
causes the test to fail under windows.  Without line 79, the test builds and passes.

chrisjohgorman:
The problem here is the expect block is different in windows.  The pattern that should be searched for is 

```
expect(registry.contains("gr::electrical::PowerMetrics<float32, 3ull>"sv));
```

as std::size_t is unsigned long long in windows.

PR #607, opened 2025-07-02
win32: fix for qa_KnownSharedLibBlocks.exe

chrisjohgorman:
- in win32 std::size_t is unsigned long long and in linux it's unsigned long.  So the string retured by the registry under windows for gr::electrical::PowerMetrics is gr::electrical::PowerMetrics<float32, 3ull> rather than gr::electrical::PowerMetrics<float32, 3ul>. Used if-def macro to test the different expected string values.
- fixes issue https://github.com/fair-acc/gnuradio4/issues/587

### DIFF
--- a/blocks/libs/test/qa_KnownSharedLibBlocks.cpp
+++ b/blocks/libs/test/qa_KnownSharedLibBlocks.cpp
@@ -76,7 +76,11 @@ const boost::ut::suite TagTests = [] {
         expect(registry.contains("gr::blocks::type::converter::ScalingConvert<float32, float64>"sv));
         expect(registry.contains("gr::basic::DataSink<float32>"sv));
         expect(registry.contains("gr::blocks::basic::SchmittTrigger<float32, (gr::trigger::InterpolationMethod)0>"sv));
+#if defined(_WIN32)
+        expect(registry.contains("gr::electrical::PowerMetrics<float32, 3ull>"sv));
+#else
         expect(registry.contains("gr::electrical::PowerMetrics<float32, 3ul>"sv));
+#endif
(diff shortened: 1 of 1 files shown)


blocks/libs/test/qa_KnownSharedLibBlocks.cpp
```diff
@@ -76,7 +76,11 @@ const boost::ut::suite TagTests = [] {
         expect(registry.contains("gr::blocks::type::converter::ScalingConvert<float32, float64>"sv));
         expect(registry.contains("gr::basic::DataSink<float32>"sv));
         expect(registry.contains("gr::blocks::basic::SchmittTrigger<float32, (gr::trigger::InterpolationMethod)0>"sv));
+#if defined(_WIN32)
+        expect(registry.contains("gr::electrical::PowerMetrics<float32, 3ull>"sv));
```
RalphSteinhagen, line 80: I wonder whether defining

```cpp
template<typename T, std::size_t nPhases>
requires(std::floating_point<T> or std::is_arithmetic_v<meta::fundamental_base_value_type_t<T>>)
struct PowerMetrics : Block<PowerMetrics<T, nPhases>, Resampling<100U, 1UZ, false>> { /*...*/}
```

as

```cpp
template<typename T, std::uint32_t nPhases>
requires(std::floating_point<T> or std::is_arithmetic_v<meta::fundamental_base_value_type_t<T>>)
struct PowerMetrics : Block<PowerMetrics<T, nPhases>, Resampling<100U, 1UZ, false>> { /*...*/}
```

and to do the cast to `std::size_t` internally might be better. 

*N.B. this is unfortunately the same issue of `std::size_t` not being a portable type. @chrisjohgorman thanks for the fix.*

PR #613, opened 2025-07-07
thread_pool::Manager: initialize default pools lazyly

wirew0rm:
Otherwise they cannot be initialized with custom implementations, because the default pools will be initialized with the first access to the manager.
On emscripten this resulted in requesting to many threads from the scheduler.


drslebedev (approved):
Thanks a lot for the PR. 

Changes look good to me, PR can be merged.

PR #616, opened 2025-07-07
Fixing various win32 errors that occur when trying to build

chrisjohgorman:
- Soapy.hpp needs a static_cast<long int> to build.
- bm_Buffer.cpp needs to conditionally exclude _WIN32 for setCpuAffinity even though we have the required headers.  We also need a UL to UZ type change.
- HistoryBuffer.cpp needs a UL to UZ type change.
- Scheduler.hpp need to unset ERROR defined in windows.h so the macro is not substituted in the work::status::ERROR method.
- MemoryMonitor.cpp windows.h needs to be included before psapi.h. psapi.h has alot of windows specific types that we don't know about before including windows.h
- thread_pool.hpp replace unsigned long type with std::size_t.  This squares with the type for numThreads() and will allow ULL under windows and UL on linux.

- tested on windows and linux

### DIFF
--- a/blocks/soapy/include/gnuradio-4.0/soapy/Soapy.hpp
+++ b/blocks/soapy/include/gnuradio-4.0/soapy/Soapy.hpp
@@ -145,7 +145,7 @@ This block supports multiple output ports and was tested against the 'rtlsdr' an
             for (std::size_t i = 0UZ; i < rx_channels->size(); ++i) {
                 output[i] = std::span<T>(outputs[i]).subspan(0, maxSamples);
             }
-            ret = _rxStream.readStreamIntoBufferList(flags, time_ns, max_time_out_us, output);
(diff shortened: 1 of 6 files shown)


RalphSteinhagen (approved):
@chrisjohgorman thanks for following this up! :+1:

RalphSteinhagen:
@chrisjohgorman the clang-formatting is missing (see Restyled CI [action](https://github.com/fair-acc/gnuradio4/actions/runs/16121511184/job/45488213856?pr=616)). Other than that, this could be merged.

chrisjohgorman:
Okay @RalphSteinhagen .  Took a look at what is missing and it appears to be the desire to format with headers in alphabetical order.  What is there is right, `windows.h` must be included before `psapi.h`, or things fail.  To accomplish this without causing errors, do I just put the includes between a `// clang-format off` and `// clang-format on` set of comments?

PR #617, opened 2025-07-08
Add string() member to filesystem paths

chrisjohgorman:
- win32 requires the explicit use of the string() member on filesystem paths whereas linux allows the use of filesystem paths without string().

- windows uses wchar_t for the std::filesystem::path format where linux uses char.  Therefore we would need to be using std::wstring rather than std::string for our string representations to use the implicit conversion.  Since we are using std::string, under windows we must add a .string() member suffix to build our string properly.

- tested on windows and linux.

### DIFF
--- a/blocklib_generator/tools/parse_registrations.cpp
+++ b/blocklib_generator/tools/parse_registrations.cpp
@@ -298,7 +298,7 @@ int main(int argc, char** argv) try {
 
     const auto integratorSourceFile = (options.outDir / "integrator.cpp");
     if (!std::filesystem::exists(integratorSourceFile)) {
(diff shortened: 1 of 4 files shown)


RalphSteinhagen (approved):
Thanks! :+1:

chrisjohgorman:
Sorry for the confusion @RalphSteinhagen.  I resubmitted this PR here https://github.com/fair-acc/gnuradio4/pull/621.  It's the same code.  I mistakenly thought the code had been submitted to main, instead of just being reviewed.  Then I deleted it, both my local branch and the remote one.  I guess the lesson is not to do work in the middle of the night.

PR #618, opened 2025-07-08
Refactor Graph/Scheduler ownership & traversal: shared_ptr<BlockModel>, freestanding forEachBlock/Edge, better diagnostics

RalphSteinhagen:
This PR introduces a series of internal refactorings and ownership improvements for the `Graph`, `Edge`, and `SchedulerBase` subsystems, centred on robustness, reusability, and safe inspection of nested sub-graphs:

---

### Key changes:

* **Ownership modernisation**

  * Changed `Graph::_blocks` and `Edge::_sourceBlock/_destinationBlock` to use `std::shared_ptr<BlockModel>` instead of `unique_ptr` or raw pointers.
  * This enables:

    * **Safe flattening of nested sub-graphs** (copy-by-reference).
    * **Thread-safe traversal and introspection**, especially across scheduler boundaries.
    * **Consistent lifetime guarantees** during runtime reconfiguration (e.g., adoption, zombie cleanup).

* **New freestanding traversal functions** (replacing `forAllUnmanagedBlocks(...)`, `Graph::forEachBlock(...)`, etc.)
  Introduced under `namespace gr::graph`, with support for nested traversal:

  ```cpp
  template<block::Category traverseCategory, typename Fn>
  void forEachBlock(GraphLike auto const& root, Fn&& function, block::Category filter = block::Category::All);

  template<block::Category traverseCategory, typename Fn>
  void forEachEdge(GraphLike auto const& root, Fn&& function, Edge::EdgeState filter = Edge::EdgeState::Unknown);
  ```

* These support recursive traversal of arbitrary sub-graph structures (including transparent and scheduled block groups).

* new `GraphLike` new concept to generalise over `Graph`, `Block` (managed/unmanaged sub-graphs) and `GraphWrapper`.

* **Changed return types of `findBlock(...)`** to `std::expected`, to allow better diagnostics in failure cases
* **Renamed and removed redundant in-Graph traversal methods**
* **Simplified `Edge` structure**
  * Replaced move/delete logic with simple `shared_ptr`-based value semantics.
  * Added `operator==` for edge comparisons.
  * Updated `graph::flatten(...)` to support reusable sub-graph collapsing logic.

---

### Rationale

* **Shared ownership** enables dynamic composition and graph transformation without invalidating references.
* **Freestanding algorithms** decouple traversal logic from `Graph`, enabling composability across wrappers and future scheduler backends.
* **`expected` over `optional`** improves observability in graph diagnostics and integration testing.

-- 

Please have a look at the new API, notably free-standing helper functions. I kept the individual commits for those who prefer comit-by-comit reading. However, they can be squash-merged if deemed necessary or not useful for future readers. 

Thanks in advance.

### DIFF
--- a/core/include/gnuradio-4.0/BlockModel.hpp
+++ b/core/include/gnuradio-4.0/BlockModel.hpp
@@ -38,15 +38,15 @@ struct Edge {
     enum class EdgeState { WaitingToBeConnected, Connected, Overridden, ErrorConnecting, PortNotFound, IncompatiblePorts };
 
     // Member variables that are controlled by the graph and scheduler
-    BlockModel*    _sourceBlock      = nullptr; /// non-owning reference
-    BlockModel*    _destinationBlock = nullptr; /// non-owning reference
-    PortDefinition _sourcePortDefinition;
-    PortDefinition _destinationPortDefinition;
-    EdgeState      _state            = EdgeState::WaitingToBeConnected;
-    std::size_t    _actualBufferSize = -1UZ;
-    PortType       _edgeType         = PortType::ANY;
-    DynamicPort*   _sourcePort       = nullptr; /// non-owning reference
-    DynamicPort*   _destinationPort  = nullptr; /// non-owning reference
+    std::shared_ptr<BlockModel> _sourceBlock;
+    std::shared_ptr<BlockModel> _destinationBlock;
+    PortDefinition              _sourcePortDefinition;
+    PortDefinition              _destinationPortDefinition;
+    EdgeState                   _state            = EdgeState::WaitingToBeConnected;
+    std::size_t                 _actualBufferSize = -1UZ;
+    PortType                    _edgeType         = PortType::ANY;
+    DynamicPort*                _sourcePort       = nullptr; /// non-owning reference
+    DynamicPort*                _destinationPort  = nullptr; /// non-owning reference
 
     // User-controlled member variables
     std::size_t  _minBufferSize;
@@ -81,13 +81,13 @@ public:
         return *this;
     }
 
-    Edge(BlockModel* sourceBlock, PortDefinition sourcePortDefinition, BlockModel* destinationBlock, PortDefinition destinationPortDefinition, std::size_t minBufferSize, std::int32_t weight, std::string name) : _sourceBlock(sourceBlock), _destinationBlock(destinationBlock), _sourcePortDefinition(sourcePortDefinition), _destinationPortDefinition(destinationPortDefinition), _minBufferSize(minBufferSize), _weight(weight), _name(std::move(name)) {}
+    Edge(std::shared_ptr<BlockModel> sourceBlock, PortDefinition sourcePortDefinition, std::shared_ptr<BlockModel> destinationBlock, PortDefinition destinationPortDefinition, std::size_t minBufferSize, std::int32_t weight, std::string name) : _sourceBlock(sourceBlock), _destinationBlock(destinationBlock), _sourcePortDefinition(sourcePortDefinition), _destinationPortDefinition(destinationPortDefinition), _minBufferSize(minBufferSize), _weight(weight), _name(std::move(name)) {}
 
-    [[nodiscard]] constexpr const BlockModel& sourceBlock() const noexcept { return *_sourceBlock; }
-    [[nodiscard]] constexpr const BlockModel& destinationBlock() const noexcept { return *_destinationBlock; }
-    [[nodiscard]] PortDefinition              sourcePortDefinition() const noexcept { return _sourcePortDefinition; }
-    [[nodiscard]] PortDefinition              destinationPortDefinition() const noexcept { return _destinationPortDefinition; }
-    [[nodiscard]] constexpr EdgeState         state() const noexcept { return _state; }
+    [[nodiscard]] constexpr const std::shared_ptr<BlockModel>& sourceBlock() const noexcept { return _sourceBlock; }
+    [[nodiscard]] constexpr const std::shared_ptr<BlockModel>& destinationBlock() const noexcept { return _destinationBlock; }
+    [[nodiscard]] PortDefinition                               sourcePortDefinition() const noexcept { return _sourcePortDefinition; }
+    [[nodiscard]] PortDefinition                               destinationPortDefinition() const noexcept { return _destinationPortDefinition; }
+    [[nodiscard]] constexpr EdgeState                          state() const noexcept { return _state; }
 
     [[nodiscard]] constexpr std::size_t      minBufferSize() const noexcept { return _minBufferSize; }
     [[nodiscard]] constexpr std::int32_t     weight() const noexcept { return _weight; }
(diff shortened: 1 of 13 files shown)


core/include/gnuradio-4.0/BlockModel.hpp
```diff
@@ -103,14 +82,21 @@ struct Edge {
     constexpr PortType    edgeType() const { return _edgeType; }
 
     constexpr bool hasSameSourcePort(const Edge& other) const noexcept {
```
ivan-cukic: This can be just `return sourceBlock() == other.sourceBlock() && sourcePortDefinition().definition == other.sourcePortDefinition().definition` (to use getters as `operator==` does)

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -48,16 +56,55 @@ namespace graph {
...
+    for (const auto& block : graph.blocks()) {
+        os << std::format("   -block: {} ({})\n", block->name(), block->uniqueName());
+    }
+    for (const auto& edge : graph.edges()) {
+        os << std::format("   -edge: {}\n", edge);
+    }
+    return os.str();
+}
+
+std::expected<std::shared_ptr<BlockModel>, Error> findBlock(GraphLike auto const& graph, BlockLike auto& what, std::source_location location = std::source_location::current()) {
```
ivan-cukic: Maybe these should be marked as noexcept to communicate errors are propagated only with `expected`

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -264,7 +284,9 @@ struct Graph : public gr::Block<Graph> {
 
             const bool        isArithmeticLike       = sourcePort->kIsArithmeticLikeValueType;
             const std::size_t sanitizedMinBufferSize = minBufferSize == undefined_size ? graph::defaultMinBufferSize(isArithmeticLike) : minBufferSize;
-            self._edges.emplace_back(sourceBlock, PortDefinition{sourcePortIndex, sourcePortSubIndex}, destinationBlock, PortDefinition{destinationPortIndex, destinationPortSubIndex}, sanitizedMinBufferSize, weight, std::move(edgeName));
+            self._edges.emplace_back(sourceBlock.value(), PortDefinition{sourcePortIndex, sourcePortSubIndex}, //
```
ivan-cukic, line 286: `.value()` performs another check while we've already checked sourceBlock for validity. It is ok to remain, a bit slower, but this is not hot path anyhow.

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -356,31 +378,49 @@ struct Graph : public gr::Block<Graph> {
      */
     [[nodiscard]] const Sequence& progress() const noexcept { return *_progress.get(); }
 
-    BlockModel& addBlock(std::unique_ptr<BlockModel> block) {
-        auto& newBlock = _blocks.emplace_back(std::move(block));
-        newBlock->init(_progress, this->compute_domain);
-        // TODO: Should we connectChildMessagePorts for these blocks as well?
-        return *newBlock.get();
+    std::shared_ptr<BlockModel>& addBlock(std::shared_ptr<BlockModel> block, bool initBlock = true) {
```
ivan-cukic: Should be const reference to shared pointer - we don't want someone changing to what the pointer points to.

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -356,31 +378,49 @@ struct Graph : public gr::Block<Graph> {
...
+    TBlock& emplaceBlock(gr::property_map initialSettings = gr::property_map()) {
         static_assert(std::is_same_v<TBlock, std::remove_reference_t<TBlock>>);
-        auto& newBlock    = _blocks.emplace_back(std::make_unique<BlockWrapper<TBlock>>(std::move(initialSettings)));
+        auto& newBlock    = _blocks.emplace_back(std::make_shared<BlockWrapper<TBlock>>(std::move(initialSettings)));
         auto* rawBlockRef = static_cast<TBlock*>(newBlock->raw());
         rawBlockRef->init(_progress);
         return *rawBlockRef;
     }
 
     [[maybe_unused]] auto& emplaceBlock(std::string_view type, property_map initialSettings) {
```
ivan-cukic: Let's make this return const as well, and maybe spell out the return type. I know I overuse auto and I'm trying to fix that :)

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -780,29 +812,87 @@ struct Graph : public gr::Block<Graph> {
...
-        std::ranges::for_each(_edges, f);
-    }
-};
+namespace detail {
+template<block::Category traverseCategory, typename Fn>
+void traverseSubgraphs(GraphLike auto const& root, Fn&& visitGraph) {
+    using enum block::Category;
 
-static_assert(BlockLike<Graph>);
+    auto recurse = [&visitGraph](GraphLike auto const& graph, auto& self) -> void {
```
ivan-cukic, line 829: Accustomed to seeing the recursive `self` as the first arg, but this is fine as well

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -780,29 +812,87 @@ struct Graph : public gr::Block<Graph> {
...
+            }
+        }
+    });
+}
+
+template<block::Category traverseCategory, typename Fn>
+requires(std::invocable<Fn, const Edge&> || std::invocable<Fn, Edge&>)
+void forEachEdge(GraphLike auto const& root, Fn&& function, Edge::EdgeState filter) {
+    using enum Edge::EdgeState;
+    using EdgeType = std::conditional_t<std::invocable<Fn, const Edge&>, const Edge&, Edge&>;
```
ivan-cukic: This will not detect properly const vs non-const for generic lambdas
https://godbolt.org/z/v75d5cxv7

core/include/gnuradio-4.0/Graph_yaml_importer.hpp
```diff
@@ -24,11 +24,11 @@ inline void loadGraphFromMap(PluginLoader& loader, gr::Graph& resultGraph, gr::p
         const auto blockType = std::get<std::string>(grcBlock["id"]);
 
         if (blockType == "SUBGRAPH") {
-            auto& subGraph           = resultGraph.addBlock(std::make_unique<GraphWrapper<gr::Graph>>());
-            createdBlocks[blockName] = &subGraph;
-            subGraph.setName(blockName);
+            std::shared_ptr<BlockModel>& subGraph = resultGraph.addBlock(std::make_shared<GraphWrapper<gr::Graph>>());
```
ivan-cukic: `const s_p`

core/include/gnuradio-4.0/Graph.hpp
RalphSteinhagen: done.

core/include/gnuradio-4.0/Graph.hpp
RalphSteinhagen: done

core/include/gnuradio-4.0/Graph.hpp
RalphSteinhagen: done

core/include/gnuradio-4.0/Graph_yaml_importer.hpp
RalphSteinhagen: done

core/include/gnuradio-4.0/BlockModel.hpp
RalphSteinhagen: done

core/include/gnuradio-4.0/Graph.hpp
RalphSteinhagen: Thanks for the feedback. I’ve dropped the `requires(...)` clause and type conditional, as it doesn’t reliably improve diagnostics, especially with generic lambdas where `std::invocable<Fn, const T&>` may yield false positives. The compiler provides a meaningful error at the point of invocation anyway.

ivan-cukic (approved):

PR #620, opened 2025-07-09
Fixes for CmakeLists.txt files for building under win32

chrisjohgorman:
- CMakeLists.txt
  - added conditional to make testing available for win32
  - run fftw build with bash.exe.  win32 doesn't have a facility to run configure scripts, use bash as a command to run the configure script.
  - also specify build compiler, without specifying the build compiler, build under clang64 attempted to use msys2's gcc, rather than the required clang.exe
  - under gcc use -Wa,-mbig-obj to build big object files, without this the build fails.
  - add -lstdc++exp to the build this facilitates the use of <print> and <format> for gcc.
  - define BUILDING_MYPLUGIN for library loading

- core/src/CMakeLists.txt
  - missed include file with semi colon separating $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR};${CMAKE_CURRENT_BINARY_DIR}>

- rest of CMakeLists.txt files added dl library to link against for dlopen(), dlclose(), etc.


CMakeLists.txt
```diff
@@ -396,7 +420,9 @@ if(EMSCRIPTEN)
 else()
   set(FFTW_CFLAGS "${CFLAGS} -fPIC -w -O2 -g0 -march=native -mtune=native -ffunction-sections -fdata-sections")
   set(FFTW_CONFIG
+      ${BASH}
       ${FFTW_PREFIX}/src/configure
+      ${BUILD_COMPILER}
```
wirew0rm:
```suggestion
      CC=${CMAKE_C_COMPILER}
```

I currently have no way of testing on windows, but this should give directly the full path to the compiler passed to cmake. This seems preferable to me to hardcoding compiler binaries.

blocks/soapy/src/CMakeLists.txt
```diff
@@ -8,3 +8,6 @@ target_link_libraries(
           gr-testing
           gr-soapy
           ut)
+if(WIN32)
+  target_link_libraries(soapy_example PRIVATE dl)
```
wirew0rm: Instead of adding dl to every target manually, we should identify which target actually depends on dl and a a `PUBLIC` or `INTERFACE` (transitive) dependency there.
Should be `gnuradio-core` or `gnuradio-blocklib-core` depending on which of them contain the actual calls.

Same for all the other instances of adding dl directly to the individual binary targets.

CMakeLists.txt
```diff
@@ -360,6 +363,27 @@ install(DIRECTORY ${pmt_SOURCE_DIR}/include/ DESTINATION ${CMAKE_INSTALL_INCLUDE
...
+  # clang64 build requires setting CC or configure finds /usr/bin/gcc from msys2
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    set(BUILD_COMPILER "CC=clang.exe")
+  endif()
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    set(BUILD_COMPILER "CC=gcc.exe") # ensure we're using the right compiler for fftw build
+    # we need to set -Wa,-mbig-obj or build fails with gcc
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj")
+    # on windows gcc requires stdc++exp to link for <format> and <print> functions
+    set(CMAKE_CXX_STANDARD_LIBRARIES "${CMAKE_CXX_STANDARD_LIBRARIES} -lstdc++exp")
```
wirew0rm: Shouldn't these be moved to somewhere around L52, since they seem to set global build options not related or at least not limited to the FFTW build?

CMakeLists.txt
```diff
@@ -360,6 +363,27 @@ install(DIRECTORY ${pmt_SOURCE_DIR}/include/ DESTINATION ${CMAKE_INSTALL_INCLUDE
 add_library(vir INTERFACE)
 target_include_directories(vir INTERFACE ${vir-simd_SOURCE_DIR}/)
 
+# the script configure won't run on win32 without asking bash to do it
+if(WIN32)
+  set(BASH "bash.exe")
+  # define GNURADIO_EXPORT for library loading
+  add_definitions(-DBUILDING_MYPLUGIN)
```
wirew0rm: We try to avoid these global definitions in favor of the target specific ones, in this case `add_target_definitons(...)`.
Could you try to find out which parts of the code use this macro and add it to the specific target (probably `gnuradio-blocklib-core`)?

Also looks like a global gcc specific option which should go somewhere around L52 if it stays a global definiton or along the corresponding target?

chrisjohgorman: Thank you for the feedback @wirew0rm.  There are quite a few places in the code that don't work unless `BUILDING_MYPLUGIN` is defined.  It conditionally defines GNURADIO_EXPORT, which I found in a lot of places in the code.  My original approach was to set a `target_compile_definition(...)` wherever the build broke.  As a result, I have 13 snippets like the following in `CMakeLists.txt` files in my dev branch.
```
if (WIN32)
  target_compile_definitions(gnuradio-plugin PRIVATE BUILDING_MYPLUGIN)
endif()
```
I can add these back in and remove the `add_definitions(-DBUILDING_MYPLIUGIN)` call.  Is that the correct approach, or did you have something else in mind?

wirew0rm: Similar as to the case with the `dl` library, this indicates to me that these targets would probably have a shared target that they all depend on to which the library or definition can be added. In your example `gnuradio-plugin` depends on `gnuradio-core` and `gnuradio-options`, which are both targets that many other targets depend on and which could be the logically correct place for this.

I think for the definitions also being applied to the targets depending on it the scope has to be changed from PRIVATE to INTERFACE/PUBLIC depending on the target type.

I didn't look deeply into what `GNURADIO_EXPORT` does, but gnuradio-core would be the place for shared functionality that any library using the library will need, while gnuradio-options is a special target we use to add things like diagnostic compiler warnings etc.

chrisjohgorman: Okay, I think I get what you're after.  Not a global definition using `add_definitions(-DBUILDING_MYPLUGIN)`, but one `target_compile_definitions(root-library INTERFACE BUILDING_MYPLUGIN)` in the appropriate file.  I suspect based on what you've told me I'll be using `gnuradio-core` as my root-library.  Thanks for the tips and recommended design fixes.

CMakeLists.txt
wirew0rm: This was not neccessary in the end? Even better if it just works out of the box, just checking not that something got lost along the way.

wirew0rm (approved):
Thanks for following up on the comments and tidying up the changes, looks good now.

I noticed you dropped explicitly specifying the compiler, please let me know whether this was dropped by accident or if turned out not to be necessary, otherwise I think this is ready to merge.

PR #622, opened 2025-07-09
CircularBuffer: enforce power-of-two size after page alignment

drslebedev:
The wraparound logic relies on `index = cursor & (size - 1)`, which requires `size` to be a power of two.
While this invariant held previously, it could silently break when the element size isn’t a power of two and page alignment produces a non-power-of-two buffer size.

We encountered this issue in `TimingSource` from the `gr-digitizer` repository, where storing 56-byte `Timing::Event` elements caused premature cursor wrapping, leading to event duplication and data loss.


RalphSteinhagen (changes requested):
Please add a reproducable test case that mitigates the failure that prompted this qfix.

RalphSteinhagen:
> The issue was that CircularBuffer size was calculated incorrectly when the object size wasn’t a power of two. I’ve already added two new tests to cover this scenario.

@drslebedev I appreciate the added tests, but they currently only confirm that the new constraint (`has_single_bit(sizeof(T))`) is enforced — they don’t actually exercise or unit-tests the original issue.

The bug appeared to be a mismatch between buffer size and mmap alignment when `sizeof(T)` wasn’t a power of two (notably for compound/non-fundamental types `T`). Without a test that fails under the old implementation — e.g. by verifying allocator fallback or catching an exception, the proposed tests only test what we *expect* because of the implementation, not what was *broken*.

This is serious stuff, possibly involving UB (according to @mattkretz and @ivan-cukic), and it deserves at least one test that would have failed before the fix, so we’re not just codifying the workaround but actually guarding against regressions and the original issue/bug.

wirew0rm (commented):
I agree that it would be good to add a unit-test that tests actual observable behavior which breaks without the changes and gets fixed by the changes instead of just checking for some invariant that should be more of an implementation detail. So writing and reading data from a buffer with a non-power-of-two struct (which is at the same time not trivially copyable) as this is what was missing from the existing tests.

Regarding the actual fix, this is orthogonal to the UB topic, as it is an in comparison simple index calculation bug. The mmap approach requires a buffer size that is a power of two while the calculation of the index using bit-shifts instead of a full modulo operation requires the number of stored elements to be a power of two.
The existing code by enforcing the first property inherently broke the second (with types with power-of-two sizes accidentally working), breaking the core assumption that makes the index calculation work.

Since it is such a critical assumption I think we should assert that `CircularBuffer::_size` is actually a power of two in the Constructor instead of just having the [comment here](1f269ca829d3252d3bba9f761921686817079d456be5fd9f64d1df624f88bc24R237).

drslebedev:
I've updated the unit tests and added `WraparoundTests` for various non-fundamental types.

Additionally, I included a separate commit that introduces an assertion in the `CircularBuffer` constructor (specifically in `BufferImpl`) to enforce that the buffer size is a power of two.

RalphSteinhagen:
> Since it is such a critical assumption I think we should assert that CircularBuffer::_size is actually a power of two in the Constructor instead of just having the [comment here](https://github.com/fair-acc/gnuradio4/pull/1f269ca829d3252d3bba9f761921686817079d456be5fd9f64d1df624f88bc24R237).

@wirew0rm the assumption of power-of-two must not be enforced/enshrined as discussed in the matrix channel unless its absolutely needed for functional reasons (double-mmap-ed buffer, for example). Otherwise, the code and documentation should be updated to reflect that requirement. 

Just because having power-of-two is common doesn't make it the right choice. There are other edge-cases (notably overall memory footprint) created by this.

Since this problem originally occurs in your gr-digitiser timing integration, I'd suggest that this fixed there first by aligning the struct to a 64-byte size — which is also highly beneficial for cache performance. 

Then we'll tackle the proper fix here once the other, more pressing things are dealt with.

wirew0rm:
This seems to need further disucssion. The original issue is now fixed in gr-digitizer, so we can decide on the proper action to take here without any hurry.

RalphSteinhagen:
Will close this draft PR. The bug and required extension have been documented here:
[[3SP] Fixing & Extending circular_buffer<T> #669](https://github.com/fair-acc/gnuradio4/issues/669)

The power-of-2 size may be used as an optimisation but must not be enshrined as a requirement. This needs a new PR.

PR #623, opened 2025-07-10
refactor scheduler adjacency list computation and  traversal logic

RalphSteinhagen:
Cleaned-up and extended infrastructure for graph traversal in schedulers. It replaces hand-rolled adjacency tracking logic with a standardised `AdjacencyList` representation, and simplifies the BFS and DFS `init()` implementations:

 * new `AdjacencyList` type and associated helpers (`computeAdjacencyList(..)`, `findSourceBlocks(..)`, `outgoingEdges(..)`)
 * `std::formatter` specialisations for `PortDefinition` and `AdjacencyList` for readable diagnostics
 * simplified job batching using `detail::batchBlocks(...)`
 * refactored DFS/BFS scheduler traversal logic and added academic reference
 * added missing unit-test for freestanding Graph helper

This should improve correctness, maintainability, and prepare for future extensions (e.g. moving from directed-acyclic- to directed-cyclic graph detection and handling).


core/benchmarks/bm_Scheduler.cpp
```diff
@@ -166,13 +166,19 @@ auto& createSink(gr::Graph& graph, bool instrumentalise = true) {
 enum class GraphTopology { DEFAULT, LINEAR, FORKED, SPLIT };
 
 void printGraphTopology(const gr::Graph& graph, GraphTopology topology) {
-    std::println("Graph Topology: {}:", magic_enum::enum_name(topology));
-    for (const auto& block : graph.blocks()) {
+    gr::Graph  flatGraph        = gr::graph::flatten(graph);
+    const bool neededFlattening = graph.blocks().size() < flatGraph.blocks().size();
```
wirew0rm, line 170: Note: I think in this diagnostic context it's not critical, but in the case of a subgraph containing just a single block or a graph containing just a single subgraph, this would not be covered.

If this is something that is relevant to users of `flatten` in a non-test context, it should probably explicitly return if it has done any flattening.

core/include/gnuradio-4.0/BlockModel.hpp
```diff
@@ -33,6 +32,26 @@ struct PortDefinition {
     constexpr PortDefinition(std::string name) : definition(StringBased(std::move(name))) {}
     bool operator==(const PortDefinition& other) const { return (definition == other.definition); }
 };
+} // namespace gr
+
+namespace std { // needs to be defined in std namespace to be used e.g. in std::unordered_map
+template<>
+struct hash<gr::PortDefinition> {
+    std::size_t operator()(const gr::PortDefinition& p) const noexcept {
+        if (p.definition.index() == 0UZ) { // index-based
```
wirew0rm:
```suggestion
        if (std::holds_alternative<gr::PortDefinition::IndexBased>(p.definition)) {
```

more readable and consistent with line 578

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -887,7 +886,84 @@ gr::Graph flatten(GraphLike auto const& root, std::source_location location = st
...
+        sources.push_back(src);
+        for (const auto& [_, edges] : ports) {
+            for (const auto* edge : edges) {
+                destinations.insert(edge->destinationBlock());
+            }
+        }
+    }
+
+    sources.erase(std::remove_if(sources.begin(), sources.end(), [&](const auto& b) { return destinations.contains(b); }), sources.end());
+    std::sort(sources.begin(), sources.end(), [](const auto& a, const auto& b) { return a->name() < b->name(); });
```
wirew0rm, line 924: Note: Sorting by name seems a bit arbitrary, especially since we only seem to do it for sources here (e.g. `outgoingEdges()` returns in definition order). Is this required somewhere later or is this just to get stable and reproducible output?

I don't think it's critical, just didn't expect it from the function name.

wirew0rm (approved):
Looks good, as always everything prefixed with "Note" does not expect any change to the PR to be made and is just to better understand the code.

core/benchmarks/bm_Scheduler.cpp
RalphSteinhagen: This is primarily for diagnostics reasons, as the default print-out does not indicate whether it's a flat or nested graph.

core/include/gnuradio-4.0/Graph.hpp
RalphSteinhagen: Yes, this was needed to have reproducible unit-tests. Because of the flattening and other post-processing steps the execution order could (notably for nested graphs) end up with sources in either ('S1', 'S2') or ('S2', 'S1') ordering which would have broken the unit-test which I largely did not want to touch (i.e. first reproduce results before breaking behaviour).

This could -- in theory -- be dropped but I found it convenient that the user is presented with an ordering that corresponds to an alphabetically sorted list according to the names that the user chose to provide these blocks. Makes it the printout more human-friendly/readable but is (strictly speaking) not required from a technical point.

core/include/gnuradio-4.0/BlockModel.hpp
RalphSteinhagen: Merged this locally and pushed again.

PR #624, opened 2025-07-10
Add UI constraints and unify block serialization

ivan-cukic:
This patch introduces `ui_constraints`  field to Block for the client applications to store their internal per-block data. A common use-case for an UI client application would be to store the information about the block position in the flowgraph view.

Further, this unifies graph/block introspection messages and YAML serialization and parsing to use the same structure/format.


core/include/gnuradio-4.0/BlockModel.hpp
```diff
@@ -546,6 +552,186 @@ class BlockWrapper : public BlockModel {
...
+    result["buffer_size"s] = edge.bufferSize();
+    result["edge_state"s]  = std::string(magic_enum::enum_name(edge.state()));
+    result["n_readers"s]   = edge.nReaders();
+    result["n_writers"s]   = edge.nWriters();
+    result["type"s]        = std::string(magic_enum::enum_name(edge.edgeType()));
+
+    return result;
+};
+
+namespace BlockSerizalizationFlags {
```
drslebedev: typo in the name
```suggestion
namespace BlockSerializationFlags {
```

core/include/gnuradio-4.0/Settings.hpp
```diff
@@ -239,6 +239,10 @@ template<typename T>
 } // namespace settings
 
 struct SettingsBase {
+    struct ContextSettings {
+        SettingsCtx  context;
+        property_map settings;
+    };
```
drslebedev: Just a small naming suggestion: perhaps something like `CtxSettingsPair` or `CtxSettingsData` could work. The only concern I see is that the main settings class is named `CtxSettings`, so having a similarly named struct might cause some confusion.

ivan-cukic: The confusion would be increased by CtxSettingsData being defined in a superclass of CtxSettings.

Though, the fact we have CtxSettings and SettingsCtx is not that pretty right now.

core/include/gnuradio-4.0/BlockModel.hpp
```diff
@@ -495,6 +498,9 @@ class BlockWrapper : public BlockModel {
 
     UICategory uiCategory() const override { return T::DrawableControl::kCategory; }
 
+    property_map uiConstraints() const override { return blockRef().ui_constraints; }
```
RalphSteinhagen: I don't think that this needs/warrants an extra accessor since the data is already available via the existing settings interface.

RalphSteinhagen (changes requested):
The new public API 'uiConstraints()' and 'setUiConstraints()' must be removed before merging since this already accessible via the settings interface.

We need to keep the public API minimal.

core/include/gnuradio-4.0/BlockModel.hpp
```diff
@@ -546,6 +546,186 @@ class BlockWrapper : public BlockModel {
...
+    if (flags & BlockSerializationFlags::StaticData) {
+        result.emplace("type_name"s, std::string(block->typeName()));
+        result.emplace("is_blocking"s, block->isBlocking());
+        result.emplace("block_category"s, std::string(magic_enum::enum_name(block->blockCategory())));
+        result.emplace("ui_category"s, std::string(magic_enum::enum_name(block->uiCategory())));
+        result.emplace("meta_information"s, block->metaInformation());
+    }
+
+    if (flags & BlockSerializationFlags::Settings) {
+        result.emplace("settings"s, block->settings().getStored().value_or(property_map()));
```
RalphSteinhagen: This looks like a lot like hardcoded magic strings. We have a compile-time type introspection system. Why not use that?

This is very brittle in case names are changed in the `Block<T>` definition.

core/include/gnuradio-4.0/BlockModel.hpp
```diff
@@ -546,6 +546,186 @@ class BlockWrapper : public BlockModel {
...
+        for (const auto& [ctx, ctxParameters] : stored) {
+            if (std::holds_alternative<std::string>(ctx) && std::get<std::string>(ctx) == ""s) {
+                continue;
+            }
+
+            for (const auto& [ctxTime, settingsMap] : ctxParameters) {
+                pmtv::map_t ctxParam;
+
+                // Convert ctxTime.context to a string, regardless of its actual type
+                std::string contextStr = std::visit(
```
RalphSteinhagen: A simple if-else definition would easier and faster to compile than the `std::visit` pattern.

core/include/gnuradio-4.0/BlockModel.hpp
RalphSteinhagen: Feels also a bit like a PMO: what is this selectivity trying to achieve?

core/include/gnuradio-4.0/BlockModel.hpp
```diff
@@ -546,6 +546,186 @@ class BlockWrapper : public BlockModel {
...
+    Ports      = 4,  //
+    Settings   = 8,  //
+    Children   = 16, //
+    All        = 0xFF
+};
+}
+
+template<typename PluginLoader>
+requires(not std::is_pointer_v<std::remove_cvref_t<PluginLoader>>)
+inline property_map serializeBlock(PluginLoader& pluginLoader, const std::shared_ptr<BlockModel>& block, int flags) {
```
RalphSteinhagen: Not clear why we need this function. This is already present in the Settings implementatino.

core/include/gnuradio-4.0/Graph_yaml_importer.hpp
```diff
@@ -147,17 +168,18 @@ inline void loadGraphFromMap(PluginLoader& loader, gr::Graph& resultGraph, gr::p
...
-            pmtv::map_t map;
-            map["name"] = std::string(block->name());
-
-            const auto& fullTypeName = loader.registry().typeName(*block);
+        gr::graph::forEachBlock<gr::block::Category::NormalBlock>(rootGraph, [&](const std::shared_ptr<BlockModel>& block) {
+            property_map map;
+            const auto&  fullTypeName = loader.registry().typeName(block);
             if (fullTypeName == "gr::Graph") {
                 map.emplace("id", "SUBGRAPH");
+                map["uniqueName"] = std::string(block->uniqueName());
```
RalphSteinhagen: hardcoded magic field names?

RalphSteinhagen: We have a type/field_name reflection that could remedy this.

core/test/qa_SchedulerMessages.cpp
```diff
@@ -445,6 +447,72 @@ const boost::ut::suite TopologyGraphTests = [] {
...
+        for (std::size_t replyIndex = 0UZ; replyIndex < replyCount; replyIndex++) {
+            const Message reply = getAndConsumeFirstReplyMessage(scheduler.fromScheduler);
+            std::println("Got a reply {}:\n{}", replyIndex, reply);
+        }
+
+        expect(copy1->settings().applyStagedParameters().forwardParameters.empty());
+        expect(copy2->settings().applyStagedParameters().forwardParameters.empty());
+
+        auto uiConstraintsFor = [](const auto& block) {
+            return std::visit(meta::overloaded{
```
RalphSteinhagen, line 375: It's just a unit-test but: having only two options a simple if-else would be more efficient w.r.t. compile-time (and probably also runtime) performance.

core/include/gnuradio-4.0/BlockModel.hpp
```diff
@@ -546,6 +546,186 @@ class BlockWrapper : public BlockModel {
     [[nodiscard]] void*                      raw() override { return std::addressof(blockRef()); }
 };
 
+property_map serializeEdge(const auto& edge) {
```
RalphSteinhagen: Is this needed as part of the BlockModel.hpp?

Looks a bit like a convenience function. What is it use-case? Could it be put elsewhere (in the UI)?

core/include/gnuradio-4.0/BlockModel.hpp
RalphSteinhagen: Does this have to be part of the BlockModel.hpp?

core/include/gnuradio-4.0/BlockModel.hpp
RalphSteinhagen: Does this need to be part of the BlockModel?

ivan-cukic: On the contrary,  if the internal names change, the format will not and pre-rename GRC files will remain valid. Which would not be the case if this was based on the reflection data. Also, changing a name will not break messaging API.

core/include/gnuradio-4.0/BlockModel.hpp
ivan-cukic: It isn't. This can serialize everything block-related including ports.

Also, it isn't new, we already had that function in Graph.hpp and a similar one in the yaml serializer.

core/include/gnuradio-4.0/Graph_yaml_importer.hpp
ivan-cukic: Same comment as above. Though, the names should be snake case as in the rest of GRC.

core/include/gnuradio-4.0/BlockModel.hpp
ivan-cukic: It is a non-member function, isn't a part of BlockModel.

It is a part of the messaging mechanism, not something that can be UI only.

(also, not a new function, just moved)

core/include/gnuradio-4.0/BlockModel.hpp
ivan-cukic: Selectivity -- we already have issues parsing the meta_information as it is slow, so Alex disabled it in OD. This way, we can choose to return only the information that is needed.

The optional ports serialization, along with 'static data' is also not needed in GRC as it is not something we can read and change as it is fixed.

core/include/gnuradio-4.0/BlockModel.hpp
ivan-cukic: Should this be changed in the whole code-base as we use std::visit to implement port definition handling everywhere?

core/include/gnuradio-4.0/BlockModel.hpp
RalphSteinhagen: I am OK with (optionally) suppressing the meta_information (since this is needed only once per block). The other information (block settings, a few limit UI constraints) is nearly always needed in full. Maybe some caching would help there.

We follow the principle of 'the code is the single source of truth'.... thus also the reliance on compile-time reflection since this avoids a lot (!!) of development errors. 

Please do not hard-code string literals in the main code. For unit-tests it's less critical but the library should use only proper names and if possible those derived from the class design... thus also the naming conventions.

core/include/gnuradio-4.0/BlockModel.hpp
RalphSteinhagen: I'd propose to apply these changes only to the code that has been touched here (MVP) and for cases where we have only 2-3 cases. 

*N.B. We are discussion (within the GR Architecture group) whether we drop using the `rva::variant` for the pmt types.*

core/include/gnuradio-4.0/BlockModel.hpp
RalphSteinhagen: Is it urgnetly needed in the GR4 library? Does it need to reside in `BlockModel[.hpp]`. This appears to be PMO and if needed for functional GR-lib<->UI communication this should perhaps be shifted to the yaml importer/exporter.

core/include/gnuradio-4.0/Graph_yaml_importer.hpp
RalphSteinhagen: Please use the compile-time reflection and given type names. Please fix.

core/include/gnuradio-4.0/BlockModel.hpp
RalphSteinhagen: @ivan-cukic it's part of the `BlockModel[.hh]`. That it's not a member function doesn't change that fact.

We should always ask ourselves: are these functions needed for everyone who needs runtime polymorphism and access to the `BlockModel` wrapped block implementation? If not this should be rationalised, at least moved (within the GR4 lib or to the UI), and/or even eliminated.

core/include/gnuradio-4.0/BlockModel.hpp
ivan-cukic: > Is it urgnetly needed in the GR4 library?

It is already a part of GR4, it is moved from another place. And yes, it is needed, it is used for GRC saving and for communicating information about the flowgraph via the messaging system.

> Does it need to reside in BlockModel[.hpp]. 

It doesn't need to, can be in a separate header. But it is a function that serializes info about a BlockModel into a map. It can not go to yaml importer exporter as it is also used for messaging.

RalphSteinhagen (approved):
@ivan-cukic structurally OK. Would it be OK to rebase this to PR #639. 

This would possibly simplify some of the code due to the new BlockModel API, notably:

```cpp
    /**
     * @brief used to store non-graph-processing information like UI block position etc.
     */
    [[nodiscard]] virtual property_map& uiConstraints() noexcept = 0;
    [[nodiscard]] virtual const property_map& uiConstraints() const = 0;
    
    /**
     * @brief used to store static non-graph-processing information like Annotated<> info etc.
     */
    [[nodiscard]] virtual property_map& metaInformation() noexcept = 0;
    [[nodiscard]] virtual const property_map& metaInformation() const = 0;

    [[nodiscard]] virtual std::vector<gr::PortMetaInfo> inputMetaInfos(bool reset = true) noexcept  = 0;
    [[nodiscard]] virtual std::vector<gr::PortMetaInfo> outputMetaInfos(bool reset = true) noexcept = 0;
```

We should perhaps move also the other API (notably property messages) that are generic and runtime-polymorphic to BlockModel. This would help making the `Block<T>` CRTP a bit leaner and perhaps also faster to compile. :smile:

PR #626, opened 2025-07-12
WASM thread limit enforcement

RalphSteinhagen:
Part I / II:
Adds unit tests to validate global thread count tracking and limit enforcement.
This is notably relevant for Emscripten/WASM where PTHREAD_POOL_SIZE imposes a strict cap.
Ensures exceptions are thrown on overallocation, and confirms correct cleanup and counter behaviour.

Part II / II:
Refactors default thread pool initialisation in Manager() using a new computeDefaultThreadSplit() helper.
Ensures conservative CPU/IO thread split under WASM, respects global thread cap with system-wide reserve, and improves testability.

*N.B. the split is to ease reviewing, and both parts should be squash-merged once reviewed.*


core/include/gnuradio-4.0/thread/thread_affinity.hpp
```diff
@@ -370,6 +380,59 @@ inline void setThreadSchedulingParameter(Policy scheduler, int priority, thread_
...
+#if defined(__linux__) && !defined(__EMSCRIPTEN__)
+        // native Linux: use kernel/user process limits
+        return std::min({detail::getUserProcessLimit(), detail::getKernelThreadLimit(), 50000UZ});
+#elif defined(__EMSCRIPTEN__)
+#ifdef GR_MAX_WASM_THREAD_COUNT
+        // WASM: use compile-time override if defined
+        return GR_MAX_WASM_THREAD_COUNT;
+#else
+        // WASM: query browser thread count dynamically
+        return static_cast<std::size_t>(EM_JS(int, (), { return navigator.hardwareConcurrency || 2; })());
```
wirew0rm:
```suggestion
        return static_cast<std::size_t>(EM_ASM_INT({ return navigator.hardwareConcurrency || 2; }));
```

`EM_JS` defines a function which is not valid in this context and produces a compilation error.

This passes the CI because in the CI `GR_MAX_WASM_THREAD_COUNT` is set. I think this points out that we need to be very careful when expanding the set of available configuration options, otherwise there will always be some combination that cannot be properly tested.

core/include/gnuradio-4.0/thread/thread_pool.hpp
```diff
@@ -649,10 +658,38 @@ class BasicThreadPool {
...
+inline std::atomic<uint64_t> BasicThreadPool::_globalPoolId      = 0U;
+inline std::atomic<uint64_t> BasicThreadPool::_taskID            = 0U;
 static_assert(ThreadPool<BasicThreadPool>);
 
+inline std::size_t getTotalThreadCount() {
+#if defined(__EMSCRIPTEN__) || defined(EMBEDDED)
+#if defined(__EMSCRIPTEN__) && defined(GR_MAX_WASM_THREAD_COUNT)
+    std::size_t maxThreads        = static_cast<std::size_t>(GR_MAX_WASM_THREAD_COUNT);
+    std::size_t nThreadsRemaining = static_cast<std::size_t>(EM_ASM_INT({ return PThread ? PThread.unusedWorkers.length : 0; }));
+    std::size_t nTreadsWASM       = maxThreads >= nThreadsRemaining ? maxThreads - nThreadsRemaining : maxThreads;
```
wirew0rm: typo:
```suggestion
    std::size_t nThreadsWASM       = maxThreads >= nThreadsRemaining ? maxThreads - nThreadsRemaining : maxThreads;
```

core/include/gnuradio-4.0/thread/thread_pool.hpp
```diff
@@ -649,10 +658,38 @@ class BasicThreadPool {
...
+        if (line.starts_with("Threads:")) {
+            if (int val = std::stoi(line.substr(8)); val >= 0) {
+                return static_cast<std::size_t>(val);
+            }
+            throw std::runtime_error("could not get total thread count");
+            return 0UZ;
+        }
+    }
+    throw std::runtime_error("could not get total thread count");
+    return 0UZ;
```
wirew0rm, line 700: Return after throw?

core/include/gnuradio-4.0/thread/thread_pool.hpp
```diff
@@ -649,10 +658,38 @@ class BasicThreadPool {
...
+#if defined(__EMSCRIPTEN__) && defined(GR_MAX_WASM_THREAD_COUNT)
+    std::size_t maxThreads        = static_cast<std::size_t>(GR_MAX_WASM_THREAD_COUNT);
+    std::size_t nThreadsRemaining = static_cast<std::size_t>(EM_ASM_INT({ return PThread ? PThread.unusedWorkers.length : 0; }));
+    std::size_t nTreadsWASM       = maxThreads >= nThreadsRemaining ? maxThreads - nThreadsRemaining : maxThreads;
+    return std::max(nTreadsWASM, BasicThreadPool::_globalThreadCount.load(std::memory_order_relaxed));
+#else
+    return BasicThreadPool::_globalThreadCount.load(std::memory_order_relaxed);
+#endif
+#else
+    std::ifstream status{"/proc/self/status"};
```
wirew0rm, line 688: This should probably be guarded behind a feature macro for systems which actually provide procfs.

core/include/gnuradio-4.0/thread/thread_pool.hpp
```diff
@@ -649,10 +658,38 @@ class BasicThreadPool {
...
 };
 
-inline std::atomic<uint64_t> BasicThreadPool::_globalPoolId = 0U;
-inline std::atomic<uint64_t> BasicThreadPool::_taskID       = 0U;
+inline std::atomic_size_t    BasicThreadPool::_globalThreadCount = 0UZ;
+inline std::atomic<uint64_t> BasicThreadPool::_globalPoolId      = 0U;
+inline std::atomic<uint64_t> BasicThreadPool::_taskID            = 0U;
 static_assert(ThreadPool<BasicThreadPool>);
 
+inline std::size_t getTotalThreadCount() {
```
wirew0rm, line 666: To me it looks like this function returns slightly different things based on the target system:
- wasm with max threads defined: maximum - remaining number of threads or number used threads in the thread pool, whichever is bigger. Basically the first is the fallback and the second is also accounting for threads not allocated via threadpool but e.g. via std::thread directly?
- embedded + wasm without thread maxiumum defined: number threads used by the thread pool
- native with procfs: number of threads used by the executable
- native without procfs (e.g. windows): std::runtime_error

Some ideas to make this clearer:
- split into a function that returns the counter that is always available and a separate function that on a best effort basis tries to get the total thread usage via procfs/javascript/etc
- add comments for the platform specific logic
- merge the two levels of `#ifdef`s onto a single level with multiple elseifs (alternatively add comments to else/endif statements, but there is always the danger of them becoming outdated)

core/include/gnuradio-4.0/thread/thread_pool.hpp
```diff
@@ -771,14 +838,15 @@ class Manager {
...
-        const std::size_t maxConcurrency = std::max(4U, std::thread::hardware_concurrency());
-#else
-        const std::size_t maxConcurrency = std::thread::hardware_concurrency();
-#endif
-        const std::size_t maxThread = maxConcurrency <= 2UZ ? 2UZ : maxConcurrency - 2UZ;
-        auto              cpu       = std::make_shared<ThreadPoolWrapper>(std::make_unique<BasicThreadPool>(std::string(kDefaultCpuPoolId), TaskType::CPU_BOUND, maxThread, maxThread), "CPU");
-        auto              io        = std::make_shared<ThreadPoolWrapper>(std::make_unique<BasicThreadPool>(std::string(kDefaultIoPoolId), TaskType::IO_BOUND, 2U, std::numeric_limits<uint32_t>::max()), "CPU");
+        const auto [cpuMin, cpuMax, ioMin, ioMax, _] = detail::computeDefaultThreadSplit();
+        const std::size_t nThreadsTotal              = cpuMax + ioMax;
+        if (nThreadsTotal >= gr::thread_pool::thread::getThreadLimit()) {
```
wirew0rm, line 854: Isn't this false by construction of `computeDefaultThreadSplit`? Since the call to that logic is hardcoded, i think the check could also be moved inside of that function. At least to me the error message claims it's configuration, when there is no way to actually directly configure this.

core/include/gnuradio-4.0/thread/thread_pool.hpp
```diff
@@ -771,14 +838,15 @@ class Manager {
...
+            throw std::out_of_range(std::format("gr::thread_pool::Manager - config violation #CPU {} + #IO {} >= getThreadLimit(): {}", cpuMax, ioMax, gr::thread_pool::thread::getThreadLimit()));
+        }
+
+        auto cpu = std::make_shared<ThreadPoolWrapper>(std::make_unique<BasicThreadPool>(kDefaultCpuPoolId, TaskType::CPU_BOUND, cpuMin, cpuMax), "CPU");
+        auto io  = std::make_shared<ThreadPoolWrapper>(std::make_unique<BasicThreadPool>(kDefaultIoPoolId, TaskType::IO_BOUND, ioMin, ioMax), "CPU");
+
         registerPool(std::string(kDefaultCpuPoolId), std::move(cpu));
         registerPool(std::string(kDefaultIoPoolId), std::move(io));
```
wirew0rm, line 862: I still think that it would be preferable to initialize these default thread pools only on the first time they are requested, to allow a user to not create them at all by replacing the default pool with a custom pool at the beginning of their code. Imo this would solve the root problem and not just minimize the impact.

I don't really see any arguments against moving this initialization from the Manager's constructor to the get call between the find and throwing the exception.

However this is more or less orthogonal and can be discussed and implemented separately, The only connection is that it impacts the same usecases, that's why I'm adding it here.

CMakeLists.txt
```diff
@@ -102,6 +102,10 @@ option(THREAD_SANITIZER "Enable thread sanitizer" OFF)
 option(ENABLE_TBB "Enable the TBB dependency for std::execution::par in gcc" OFF)
 option(ENABLE_EXAMPLES "Enable Example Builds" ${GR_TOPLEVEL_PROJECT})
 option(ENABLE_TESTING "Enable Test Builds" ${GR_TOPLEVEL_PROJECT})
+set(GR_MAX_WASM_THREAD_COUNT
```
wirew0rm, line 111: I'm not super happy with adding ever more configuration settings to gnuradio, since this increases the testing surface significantly and all downstream users need to know what they are expected to add to these settings and what will be influenced by them.

Specifically in this case, I think that instead of trying to guess what downstream users expect from thread management, we should default to a level that works for the 99% cases, but allows downstream users to customize the behavior if they either need to tune for performance reasons or work around target specific limitations.

Also 60 threads is in my opinion quite excessive as a default. Just for completeness, there is an absolute limit which in my version of firefox (and also on firefox-esr 128) is configured as `dom.workers.maxPerDomain: 512`, so as long as we don't open 10 dashboards on the same domain, this should work.

core/include/gnuradio-4.0/thread/thread_pool.hpp
RalphSteinhagen: Yes, to suppress the IDE's static code analysis complaining... also works in case exceptions are disabled. ;-)

core/include/gnuradio-4.0/thread/thread_pool.hpp
RalphSteinhagen: This does the same for WASM, EMBEDDED, and OTHER and always returns (best effort) the total number of active threads.

In 
  * EMBEDDED this falls back to the atomic counter -- which fails when threads are created directly via `std::thread`
 * WASM fall back as in EMBEDDED and/or count the runtime available threads -- indirectly via 'nUnusedWorker - maxAvailable' (WASM doesn't have an active worker thread count, only how many unused are available)
 * OTHER using the proc system which is the most reliable way to get the threads.
 
 I looked into portable other option (Windows, Apple, iOS, sys-based, ...) but none seemed to be as widely supported (most UNIX have proc not all sys). I think we should add these once we have/target actual support for these non-standard systems.

core/include/gnuradio-4.0/thread/thread_pool.hpp
RalphSteinhagen: I left it ouside so that 'computeDefaultThreadSplit(..)' is unit testable. With the actual thread limit it won't anymore and depends on where and how the CI executes the test (e.g. also w/ or w/o rlimits).

core/include/gnuradio-4.0/thread/thread_pool.hpp
RalphSteinhagen: If you look carefully `cpuMin` and `ioMin` are '0UZ' for WASM. So there isn't anything created until the first worker is being invoked.

This should/must not be evaluated lazily but should be allowed to fail early. Otherwise one gets funny results later on.

CMakeLists.txt
RalphSteinhagen: > Also 60 threads is in my opinion quite excessive as a default.

Hey, you left 60 threads **hardcoded**  before. :smirk: 

The reason why I added the to be configurable is because this is the **only** way to communicate and limit the max allowed threads reliably to the program. There is no WASM, Emscription, or JS Browser API for that and the few fragements that exist are unreliable and differ between nodejs and browsers (Google/Firefox implement this differently).

core/include/gnuradio-4.0/thread/thread_pool.hpp
RalphSteinhagen: Linux does. The others are not yet supported and I would take that bridge once we are there...

core/include/gnuradio-4.0/thread/thread_pool.hpp
wirew0rm: The thing is, if you do not customize the thread pool, it will also only get initialized once the first user requests the instance of the thread pool manager, so if it gets initialized in the constructor or in the getter will not actually make a difference.

However I agree that changing the minimum values to 0 should have the same effect :+1: can be resolved

core/include/gnuradio-4.0/thread/thread_affinity.hpp
```diff
@@ -370,6 +380,59 @@ inline void setThreadSchedulingParameter(Policy scheduler, int priority, thread_
...
+        // WASM: use compile-time override if defined
+        return GR_MAX_WASM_THREAD_COUNT;
+#else
+        // WASM: query browser thread count dynamically
+        return static_cast<std::size_t>(EM_ASM_INT({ return navigator.hardwareConcurrency || 2; }));
+#endif
+#elif defined(_WIN32) || defined(__APPLE__)
+        // Windows or macOS: use default 50k fallback
+        return 50000UZ;
+#else
```
ivan-cukic, line 423: `elif defined`?

core/include/gnuradio-4.0/thread/thread_pool.hpp
ivan-cukic: I'm fine with the extra check as it is not in a hot path -- and am reading this as a contract instead of an `if` (no matter if C++ contracts are hated even before they become a part of the standard :) )

core/include/gnuradio-4.0/thread/thread_pool.hpp
ivan-cukic: I'd put an ifdef windows or apple to return a fixed value similar to `getThreadLimit` plus a warning/error message so that developers using Windows or Apple can see the error and implement this properly. (otherwise, they will get a surprising error 'file not found' or similar)

Issue #637, opened 2025-07-21
Scheduler Block API Update


RalphSteinhagen:
### New Methods:

1. **`std::span<std::size_t> BlockModel::availableInputSamples()`**
2. **`std::span<std::size_t> BlockModel::minInputSamples()`**
3. **`std::span<std::size_t> BlockModel::maxInputSamples()`**
4. **`"gr::ratio" BlockModel::resamplingRatio()`** -- resampling ratio between input/output chunks (if needed & fixed)
5. **`bool BlockModel::hasAsyncInputPorts()`** and **`bool BlockModel::hasAsyncOutputPorts()`**

### Additional Updates:

* **`scheduler::exchange(Graph&& graph) -> Graph`**: Swaps the current graph with a new one.
* **Prime Input Ports with N Samples**: Needed for cyclic graphs (API TBD).
* **`max_work_items` change**: Transition from `std::size_t` to `gr::size_t` for consistency.

### Action Items:

* Finalise and update the scheduler API (rstein + Ivan).
* Ensure the new API accommodates cyclic graph constraints and other scheduler types.

This update streamlines the interface while supporting new scheduling strategies based on explicit port constraints for better control over the next `Block<T>::work()` execution.

Issue #148, opened 2023-08-11
full port, block, graph, scheduler API (-> see signatures on GR 4.0 architecture room’s Etherpad)


wirew0rm:

RalphSteinhagen:
completed.

PR #639, opened 2025-07-25
Blockmodel API addresses #148 & #637

RalphSteinhagen:
##Summary

Removes the monolithic `getPortLimits(...)` in favor of `PortCache<>`, introduce runtime `gr::Ratio`, extend the `BlockModel` surface (ports, stats, meta/ui props, work/state…), add `port::BitMask/BitPattern` helpers, and expose `meta_information` / `ui_constraints`. Includes broad UT/bench updates (new qa\_BlockModel), allocator helpers, and transitional asserts to keep legacy paths honest.

tackles long-pending issues: #148 and notably #637
---

### Why / What

* **Scheduling & perf**: Fine-grained per-port caches instead of one giant limits call.
* **Runtime math**: `gr::Ratio` enables dynamic resampling ratios.
* **API completeness**: `BlockModel` now mirrors what `Block<T>` can do, polymorphically.
* **Port semantics**: `BitMask/BitPattern`, async rules, stronger type checks.
* **Introspection/UX**: `meta_information` vs `ui_constraints` separated and exposed.
* **Tests/benches**: Coverage for all the above and parity checks during migration.

bm_portLimits benchmark results:

image

#### COMMIT c0f62b0

replaces getPortLimits() with PortCache<> + gr::Ratio, tests/benches - SQUASH-ME

This patch removes the monolithic `Block<T>::getPortLimits(...)` usage in favour of a reusable `PortCache<...>` that stores per‑port masks, limits and availability, and adds a runtime `gr::Ratio` type. Legacy code is kept only for cross‑checks and will be dropped next.

* add `PortCache<Derived, PortDirection, PortType>` with lazy `invalidateConfig()` / `invalidateStatistic()` and span/aggregate accessors
* move helper algos to `detail::` namespace: `min/max_element(_masked)`, `compareRangesMasked()`, `compareSpans()`
* wire caches into `Block` (input/output stream caches, ctor/move‑ctor init, settings-change invalidation) and replace local vectors in work loop (`on_scope_exit` guard updates)
* extend `BlockModel` API: `resamplingRatio() -> gr::Ratio` and `stride()`; implement in wrapper
* introduce `gr::Ratio` (runtime fraction): tag `normalise_t`, constexpr parse from `std::string_view`, arithmetic, `std::formatter`
* unit tests: add suites for `gr::Ratio` and `on_scope_exit`; integrate into `qa_fixed_string.cpp`
* benchmark update: rewrite `bm_portLimits.cpp` to compare legacy vs cache paths and separate “avail-only” vs “full recompute” cases
* add/use `gr::allocator::{Aligned, Logger, Default}` (already introduced) where appropriate
* drop custom `this_source_location()` helper, use `std::source_location::current()`
* sundry tidy-ups: `0UZ` literals, `[[nodiscard]]`, assert casts, NOSONAR annotations, minor formatting
* extend `BlockModel` API further:
  `blockInputTypes()/blockOutputTypes()`, `available{Input,Output}Samples(reset)`,
  `min/max{Input,Output}Requirements()`, `hasAsync{Input,Output}Ports()`,
  `{input,output}MetaInfos(reset)`, and `primeInputPort(...)`
* added qa\_BlockModel behavioural unit-test
* remove `PortDirection::ANY` (query-only); adjust encoders/decoders and asserts accordingly
* optional **input** ports are now implicitly async (`kIsSynch` logic updated)
* `PortMetaInfo` gains a `data_type` field (auto-filled via `meta::type_name<T>()`); unify single definition and use everywhere
* `PortWrapper` initialises `metaInfo.data_type` on construction/move
* `DynamicPort`: add `non_owned_reference_tag` ctor, `portName()`, `portMetaInfo()`, `portMaskInfo()`; connect now sanity-checks port/data types (DEBUG throws)
* more Boost.UT coverage: PortIn/Out edge cases, `publishPendingTags` merge, DynamicPort connect/disconnect/error paths, message ports, tag-distance helpers; bigger BitMask/BitPattern suite

The original `getPortLimits(...)` remains temporarily with `assert(...)` guards to ensure behavioural parity until it is removed in the later commit.

---

#### COMMIT b91d2b4

refactor/split getPortLimits(...) & new port::BitMask helpers - SQUASH-ME

This commit splits the legacy `Block<T>::getPortLimits(...)` function into modular, per-port functions (`getMinInputItems(...)`, `getMaxOutputItems(...)`, etc.) to retain individual port constraints before aggregation.
These functions enable more flexible and fine-grained scheduling decisions in the GR4 scheduler.

* introduce strongly typed `enum class PortSync`
* add `port::BitMask` + `port::BitPattern` and \`port::pattern\<Enums...>()\`\` for compact per‑port state & masked matching
* new helpers: `getPortTypes()`, `getPortConstraints()`, `[min|max]_element(_masked)()`, `compareRangesMasked()`, `compareSpans()`
* extend qa\_Port suite by BitMask/BitPattern tests
* added `gr::allocator::Aligned<T,N>`, `gr::allocator::Logger<T>`, and `gr::allocator::Default<T>` definitions and unit-test -> long-term goal: refactor recurring memory-pool (arena, stack, ..., mmap-aligned, no-alloc/static into common abstraction <-> HPC/embedded tuning)

The original `getPortLimits(...)` is retained temporarily with `assert(...)` checks to ensure correctness and test coverage continuity during the transition before being removed in the following commit.

#### COMMIT 575cfe8

extended Block\[Model]: new kMetaInformation/kUiConstraints props + API

* introduce block::property::{kMetaInformation, kUiConstraints}
* add ui\_constraints property\_map and keep meta\_information static
* expose both via message callbacks and BlockModel(Wrapper) accessors
* include ui\_constraints/meta\_information in GR\_MAKE\_REFLECTABLE
* extend/rename annotation tests; add ui\_constraints test
* bump expected parameter counts in qa\_Settings
* throw on unsupported message commands for new properties

### Follow‑ups / Next Steps

* Remove legacy `getPortLimits(...)` path and related asserts once parity is proven in CI.
* Consolidate other allocator abstractions into a common pool manager.
* Broaden scheduler heuristics to fully exploit per-port caches and async semantics.
* Remove and fix some loitering 'TODO's


core/test/qa_Block.cpp
```diff
@@ -886,29 +886,48 @@ const boost::ut::suite<"Stride Tests"> _stride_tests = [] {
     };
 };
 
-const boost::ut::suite<"Drawable Annotations"> _drawableAnnotations = [] {
+const boost::ut::suite<"Annotations"> _drawableAnnotations = [] {
```
wirew0rm, line 888: nit: also change variable name?

core/include/gnuradio-4.0/Port.hpp
```diff
@@ -555,8 +557,13 @@ struct Port {
     using TagWriterSpanType = decltype(std::declval<TagWriterType>().reserve(0UZ));
 
     // public properties
-    constexpr static bool kIsSynch    = !std::disjunction_v<std::is_same<Async, Attributes>...>;
-    constexpr static bool kIsOptional = std::disjunction_v<std::is_same<Optional, Attributes>...>;
+    // kIsSynch:
+    //   true  -> port participates in synchronous scheduling with other sync ports
+    //   false -> port is asynchronous (does not gate scheduling)
+    // Rule: an input marked Optional is implicitly also async (i.e. otherwise it would block the sync block data processing);
+    //       outputs stay synchronous unless explicitly annotated with Async.
```
wirew0rm, line 564: Why should this be handled differently for input ports vs output ports?

I'd agree that it is easier to find examples for optional sync outputs (e.g. complex -> Im + Re) compared to optional sync inputs (the only thing I could come up with was something like an optional gating/scaling input, that would just default to true/1.0 if unconnected).

Even if we decide that this will be unsupported, I still think this special case makes the port logic more difficult to document and understand.

If I'm not mistaken, the argument given in the comment would apply to output ports in the same way. If this behaves differently in the implementation I'd prefer to make the implementation more consistent if there are no fundamental restrictions.

core/test/qa_BlockModel.cpp
```diff
@@ -0,0 +1,293 @@
...
+        };
+
+        auto sinkAvailBeforeWork = sinkBlockPtr->availableInputSamples(true);
+        expect(eq(sinkAvailBeforeWork[0UZ], 0UZ)) << "downstream block (sink) should see zero samples (no processing done)";
+    };
+
+    "input/output meta infos"_test = [] {
+        BlockWrapper<SimpleBlock> simpleWrapper;
+        // meta infos are cached by port caches; reset=true forces reading
+        auto inputInfos  = simpleWrapper.inputMetaInfos(true);
```
wirew0rm, line 237: Wouldn't it be better to test this without forcing the cache reset? (and then maybe change something and check that forcing the reset actually gets the updated value?)
I would assume that most users would want the cached variant, so it should also be used in the unittests so users looking for the correct way to use it there get the correct API and are not led to always passing true.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -87,6 +88,345 @@ template<PortType portType, PortReflectable Self>
...
+        _maxSyncAvailable  = detail::min_element_masked<PortSync::SYNCHRONOUS>(_available, _types).value_or(gr::undefined_size);
+        _hasASyncAvailable = detail::compareRangesMasked<PortSync::ASYNCHRONOUS>(_available, _minSamples, _types);
+        _dirtyAvailable    = false;
+    }
+
+public:
+    PortCache(Derived& self) : _self(self) {}
+
+    void invalidateConfig() noexcept { _dirtyConfig = true; }
+    void invalidateStatistic() noexcept { _dirtyAvailable = true; }
```
wirew0rm, line 261:
```suggestion
    void invalidateAvailable() noexcept { _dirtyAvailable = true; }
```
Or rename the flag, but personally If this is not meant to be extended later on with other metrics to me statistics could mean a lot of different things, so I'd prefer the function to really say what's going on.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -87,6 +88,345 @@ template<PortType portType, PortReflectable Self>
...
+    return std::nullopt;
+}
+
+template<auto... MatchPortEnums, std::ranges::range Range, typename T = std::ranges::range_value_t<Range>>
+[[nodiscard]] constexpr std::optional<T> min_element_masked(Range&& range, const std::span<const port::BitMask>& portMaskVec) {
+    auto zipped   = std::ranges::views::zip(range, portMaskVec);
+    auto filtered = zipped | std::views::filter([](const auto& pair) {
+        const auto& mask = std::get<1>(pair);
+        return port::pattern<MatchPortEnums...>().matches(mask); // actual & pattern.mask == pattern.value
+    }) | std::views::transform([](const auto& pair) { return std::get<0>(pair); });
```
wirew0rm:
```suggestion
    auto filtered = std::ranges::views::zip(range, portMaskVec)
        | std::views::filter([](const auto& pair) {
        const auto& mask = std::get<1>(pair);
        return port::pattern<MatchPortEnums...>().matches(mask); // actual & pattern.mask == pattern.value
    })
        | std::views::transform([](const auto& pair) { return std::get<0>(pair); });
```

Might be more a question of taste, but since it comes up in a few places, I don't see why the intermediate steps in a ranges pipeline should get names, imo the beauty of the pipe operator is to be able to write the logical flow without being forced to assign it to a name.
Can also be discussed outside of this PR. In general I quite like the use of ranges here, reads a lot nicer than iterator based custom search loops :+1:

core/include/gnuradio-4.0/Port.hpp
RalphSteinhagen: > [..] that would just default to true/1.0 if unconnected

That would be possible but would require an additional branching logic and matched padding of data in the unconnected port. ZOH aside there are many choices here.

How the present sync mechanism is implemented, if you have several input ports one is unconnected the whole logic blocks which is not what people would expect. Thus my low-hanging-fruit approach to make it implicitly async.

core/test/qa_Block.cpp
RalphSteinhagen: ideally, the field suite name has no meaning ... just needs to be unique. Could be also `_0`, or `_ 1`. Left it like this because the `_` alone is not yet but will be eventually supported where this would be an ideal choice.

core/include/gnuradio-4.0/Block.hpp
RalphSteinhagen: It updates not only the available but also the min/max port constraints. More could be added later.

PR #648, opened 2025-08-20
Check when loadGraphFromMap if the keys exists and can be retrieved

dantti:
Currently Opendigitizer crashes on an empty property_map


drslebedev (approved):
Thanks a lot for the PR!

Changes look good to me.

core/include/gnuradio-4.0/Graph_yaml_importer.hpp
```diff
@@ -90,7 +96,13 @@ inline void loadGraphFromMap(PluginLoader& loader, gr::Graph& resultGraph, gr::p
         }
     } // for blocks
 
-    auto connections = std::get<std::vector<pmtv::pmt>>(yaml.at("connections"));
+    std::vector<pmtv::pmt> connections;
+    if (auto it = yaml.find("connections"); it != yaml.end()) {
+        if (auto connRef = std::get_if<std::vector<pmtv::pmt>>(&it->second)) {
```
drslebedev: You could use const auto* here to make it clear this is a pointer.

core/include/gnuradio-4.0/Graph_yaml_importer.hpp
```diff
@@ -16,7 +16,13 @@ inline void loadGraphFromMap(PluginLoader& loader, gr::Graph& resultGraph, gr::p
 
     std::map<std::string, std::shared_ptr<BlockModel>> createdBlocks;
 
-    auto blks = std::get<std::vector<pmtv::pmt>>(yaml.at("blocks"));
+    std::vector<pmtv::pmt> blks;
+    if (auto it = yaml.find("blocks"); it != yaml.end()) {
+        if (auto blkRef = std::get_if<std::vector<pmtv::pmt>>(&it->second)) {
```
drslebedev: You could use `const auto*` here to make it clear this is a pointer.

drslebedev (approved):

PR #650, opened 2025-08-31
Introduce managed sub-graphs

iamsergio:
There's now a SchedulerRegistry and custom schedulers can be loaded by a plugin.

Done-with: Ivan


core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -118,13 +121,15 @@ void forEachEdge(const GraphLike auto& root, Fn&& function, Edge::EdgeState filt
...
-template<typename TSubGraph>
-class GraphWrapper : public BlockWrapper<TSubGraph> {
+template<typename TSelf, typename TSubGraph = TSelf>
+class GraphWrapper : public BlockWrapper<TSelf> {
+protected:
     std::unordered_multimap<std::string, std::string> _exportedInputPortsForBlock;
     std::unordered_multimap<std::string, std::string> _exportedOutputPortsForBlock;
 
 public:
-    GraphWrapper(gr::property_map init = {}) : gr::BlockWrapper<TSubGraph>(std::move(init)) {
```
RalphSteinhagen, line 127: Please revert:
  * derived/user-defined blocks should not define their own constructor (disables the CRTP-ones).
  * if they defined one, it should be always `XXX(gr::property_map init = {})` to support non-C++ bindings and runtime polymorphism.

core/include/gnuradio-4.0/BlockModel.hpp
```diff
@@ -789,13 +780,41 @@ class BlockWrapper : public BlockModel {
         _dynamicPortsLoader.instance = this;
     }
 
+    template<typename Arg, typename... Args>
+    requires(not std::is_same_v<std::remove_cvref_t<Arg>, gr::property_map> or sizeof...(Args) > 0)
+    explicit BlockWrapper(Arg&& arg, Args&&... args) : _block(std::forward<Arg>(arg), std::forward<Args>(args)...) {
```
RalphSteinhagen: Should avoid defining a custom constructor. If needed then only `XXX(gr::property_map init ) {}` ... see above.

core/src/Graph.cpp
```diff
@@ -0,0 +1,102 @@
+#include <gnuradio-4.0/Graph.hpp>
```
RalphSteinhagen, line 1: The GR4 core is and should remain (largely) header-only. Please inline this to the class definition. Also this level of indirection makes this harder to read/refactor and to maintain. Only use `.cpp` impl-files when absolutely needed.

core/include/gnuradio-4.0/Graph.hpp
ivan-cukic: Will do.

But mind that GraphWrapper is Graph's type-eraseure like BlockModel is for Block - it is *not* a gr::Block, no CRTP here.

Now that blocks only support property-map-based constructors, this should be possible to do instead of wrappers having to support an arbitrary number of construction arguments and perfect-forward them.

PR #651, opened 2025-09-03
SchedulerBase: eliminate custom constructors & add Graph exchange(Graph&&)

RalphSteinhagen:
*Design Goals*:
Remove custom constructors and add `Graph` exchange API so that the scheduler behaves more consistently like a standard `Block<T>` and be used for managed sub-graphs.

*Specific changes*:
- remove graph-taking constructors from user-defined schedulers (Simple, BreadthFirst, DepthFirst)
- add `std::expected<Graph, Error> exchange(Graph&& ...)` method for runtime graph replacement
- Schedulers now use customInit()/customStart()/etc. hooks instead of overriding lifecycle methods (needed due to double CRTP-indirection)
- minimises CPU thread resource allocations by reusing scheduler instances
- update all tests to use default constructor + exchange() pattern
- fix profiler to use pointer semantics for thread-local handlers

Thanks to @drslebedev and @wirew0rm for the additional changes, fixes and pre-review.

### DIFF
--- a/core/include/gnuradio-4.0/Scheduler.hpp
+++ b/core/include/gnuradio-4.0/Scheduler.hpp
@@ -179,19 +179,6 @@ public:
         std::ignore = this->settings().applyStagedParameters();
     }
 
-    // explicit SchedulerBase(gr::Graph&& graph, std::string_view defaultPoolName, const profiling::Options& profiling_options = {}) //
-    //     : _graph(std::move(graph)), _profiler{profiling_options}, _profilerHandler{_profiler.forThisThread()}, _pool(gr::thread_pool::Manager::instance().get(defaultPoolName)) {
-    //     registerPropertyCallbacks();
-    // }
-    //
-    // explicit SchedulerBase(gr::Graph&& graph, property_map initialSettings, std::string_view defaultPoolName = gr::thread_pool::kDefaultCpuPoolId) //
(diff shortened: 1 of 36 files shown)


core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -354,23 +440,35 @@ struct SchedulerBase : Block<Derived> {
...
         base_t::processScheduledMessages(); // make sure initial subscriptions are processed
         connectBlockMessagePorts();
+
+        if constexpr (requires(Derived& d) { d.customInit(); }) {
+            static_cast<Derived*>(this)->customInit();
+        }
     }
 
     void reset() {
-        _executionOrder->clear();
```
drslebedev, line 363: Before, Simple / BreadthFirst / DepthFirst::reset() explicitly called init() and job lists were rebuilt as part of the reset sequence. Now reset() methods are removed.
In the new version, the base reset() no longer clears _executionOrder, which means job lists may remain stale.

do we actually need a reset()  at all?

RalphSteinhagen: The `_executionOrder->clear();` is done as part of the `customInit()` function. Not strictly related to the 'reset()' lifecycle logic ... and as such thus removed.

PR #653, opened 2025-09-05
Update PortMetaInfo from input tags.

drslebedev:


core/include/gnuradio-4.0/Port.hpp
```diff
@@ -226,7 +226,9 @@ Follows the ISO 80000-1:2022 Quantities and Units conventions:
             if (std::holds_alternative<T>(value)) {
                 member = std::get<T>(value);
             } else {
-                throw gr::exception("invalid-argument: incorrect type for " + key);
+                if constexpr (!isNoexcept) {
```
RalphSteinhagen: Should perhaps either drop the exception in favour of `std::expected` (preferred for internal GR4-lib API) or at least provide another path and not just swallow the error. 

Also, please use `std::source_location` in the method parameter and `gr::exception`/`std::expected<T, Error>` to aid localising errors.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1092,6 +1092,7 @@ class Block : public lifecycle::StateMachine<Derived> {
 
         if (inputTagsPresent()) {
             settings().autoUpdate(_mergedInputTag); // apply tags as new settings if matching
+            for_each_port([&]<PortLike Port>(Port& inputPort) { inputPort.metaInfo.template update<true>(_mergedInputTag.map); }, inputPorts<PortType::STREAM>(&self()));
```
RalphSteinhagen: This would be a good place to apply a filter function that deduplicates each input tag streams, apply the tag to the port fields, and returns a stream view of the deduplicated canonical fields that have been detected. This way, the `gr::tag::kDefaultTags` have to be filtered once, and the result being transferred to the `Block<T>` settings.

Similarly, this probably also (nearly) eliminates large parts of the usages/need for the `_mergedInputTag`.

The follow-up PR should address keeping a reference on this view, and forward these filtered tags to each output, before eliminating the `_mergedInputTag` entirely.

drslebedev: Most deduplication is needed across different ports. And we probably need that each port applies only the tags it actually received (probably `sample_rate` can be an exception). But the current `mergedTag` approach effectively applies all tags to all ports.

I experimented with a free-standing filter function, `getNonDuplicatedInputTags()`, to deduplicate input tags on the fly, but ran into a key issue: obtaining tags requires calling `InputSpan::tags`, which in turn requires `InputSpan`s via`CircularBuffer::Reader::get()` . Doing this inside `getNonDuplicatedInputTags()` conflicts with the buffer lifecycle and triggers:

```
The method CircularBuffer::Reader::get() was invoked after consume() was explicitly invoked.
```

Basically here the problem is similar if I have a list of ports, I can't easily get list of tags for each port using `for_each_port` because  obtaining tags requires calling `InputSpan::tags`.

Given the current GR4 architecture, this seems hard to resolve cleanly. A pragmatic path forward would be to maintain a lightweight cache:

```cpp
std::vector<std::pair<std::ptrdiff_t, std::reference_wrapper<const property_map>>> _inputTags;
```

This cache can be refreshed once per `workInternal` call, stores only references,.

core/include/gnuradio-4.0/Port.hpp
drslebedev: I’ve made `PortMetaInfo::update` always `noexcept` and it now returns `std::expected<void, Error>`.
The open question was how the constructor should handle failures. 

Possible strategies were: 1) throwing constructor, 2) static factory returning `expected`, 3) record error, 4) ignore, 5) terminate.

I implemented  a throwing constructor to match `Block` constructor behavior when parameters are invalid.
```cpp
    explicit PortMetaInfo(std::initializer_list<std::pair<const std::string, pmtv::pmt>> initMetaInfo) noexcept(false) //
        : PortMetaInfo(property_map{initMetaInfo.begin(), initMetaInfo.end()}) {}
    explicit PortMetaInfo(const property_map& metaInfo) noexcept(false) {
        if (auto res = update(metaInfo); !res) {
            throw gr::exception(res.error().message, res.error().sourceLocation);
        }
    }
```

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1076,20 +1077,34 @@ class Block : public lifecycle::StateMachine<Derived> {
      * Merge tags from all sync ports into one merged tag, apply auto-update parameters
      */
     void updateMergedInputTagAndApplySettings(auto& inputSpans, std::size_t untilLocalIndex = 1UZ) noexcept {
+        std::size_t untilLocalIndexAdjusted = untilLocalIndex;
+        if constexpr (!backwardTagForwarding) {
+            untilLocalIndexAdjusted = 1UZ;
+        }
+        const auto isSame1 = [](const auto& lhs, const auto& rhs) { return lhs.first == rhs.first; };
+        const auto isSame2 = [](const auto& lhs, const auto& rhs) { return lhs.first == rhs.first && lhs.second.get() == rhs.second.get(); };
```
wirew0rm:
```suggestion
        const auto isIndexEqual = [](const auto& lhs, const auto& rhs) { return lhs.first == rhs.first; };
        const auto isTagMapEqual = [](const auto& lhs, const auto& rhs) { return lhs.first == rhs.first && lhs.second.get() == rhs.second.get(); };
```

Right now that processing function seems more or less obvious, but after some time i think it might become hard to know what what the code is doing without descriptive names.

meta/include/gnuradio-4.0/meta/RangesHelper.hpp
```diff
@@ -0,0 +1,96 @@
...
+    explicit constexpr AdjacentDeduplicateView(T&& t) : eq(std::forward<T>(t)) {}
+
+    template<ViewableForwardRange Range>
+    constexpr auto operator()(Range&& range) const {
+        return std::forward<Range>(range) | std::views::chunk_by(eq) | std::views::transform([](auto&& chunk) -> decltype(auto) { return *std::ranges::begin(chunk); });
+    }
+};
+
+template<class Eq>
+AdjacentDeduplicateView(Eq) -> AdjacentDeduplicateView<std::decay_t<Eq>>;
```
wirew0rm: This seems not to be used anymore in the current iteration of the code. Do you expect this to be needed for the next steps or is this superseded by the Pairwise deduplication below and should maybe be removed?

core/include/gnuradio-4.0/Block.hpp
wirew0rm: Just to update this with the current state of the implementation:
- The port properties are updated on a per port basis with the deduplicated tag stream from each port. The problem of obtaining the ports and the corresponding tag stream in one place was solved by iterating over the ports and input spans in parallel (`Block::for_each_port_and_reader_span`), which needs no storage of any sort.
- The Ports now provide a range of deduplicated tags, which are used in work internal to update the `_mergedTag` which is stil used to update the settings and can be accessed from the user processing function. This will be replaced in a later PR by a non-merged tag API.

wirew0rm (approved):
Looks good.

If i didn't overlook something there is a rather big part of the ranges logic that is not used anymore and could possibly be removed to keep the codebase a bit smaller. If this is correct and it's also not needed for the upcoming PRs, I would suggest to remove that before merging.

meta/include/gnuradio-4.0/meta/RangesHelper.hpp
drslebedev: Currently it is not used, also I'm not 100% sure if we need it later. So I will remove it.

PR #654, opened 2025-09-10
Feature: Feedback Loop Support for Cyclic Directed Graphs

RalphSteinhagen:
This PR adds support for feedback loops (cyclic directed graphs) in GNU Radio 4, enabling closed-loop signal processing topologies commonly used in control systems, IIR filters, and adaptive algorithms.

image

**Key Features:**
- **Automatic Loop Detection**: Identifies all feedback loops in a graph using depth-first search
- **Loop Priming**: Calculates minimum samples required based on block constraints (resampling ratios, stride, minimum input requirements)
- **Rate Consistency Validation**: Ensures feedback loops maintain stable sample rates (net transformation must be 1:1, e.g., 3:1 decimation must be paired with 1:3 interpolation before closing the loop)
- **Scheduler Integration**: Automatically primes detected loops during graph initialisation

**Technical Details:**
- implemented `detectFeedbackLoops()` using DFS-based cycle detection
- created `calculateLoopPrimingSize()` to determine initialisation requirements
- added `primeLoop()` to inject initial samples at loop closure points
- extended scheduler to handle both flat and hierarchical graphs with feedback
- added `std::hash<Edge>` support
- added comprehensive test coverage including simple loops, resampling paths, and nested topologies

**Supported Scenarios:**
- simple feedback loops (e.g., IIR filters, control loops)
- resampling feedback paths with interpolation/decimation
- multiple nested/interconnected feedback loops
- hierarchical sub-graphs with feedback across subgraph boundaries

Some additional visual examples (N.B. UTF-8 plotting will be a follow-up PR):
image
image

*N.B. There are likely still some remaining edge cases, notably rare striding/resampling combinations, that are not (yet) covered. Further PRs/improvements are welcome.*


core/include/gnuradio-4.0/BlockModel.hpp
```diff
@@ -72,6 +72,8 @@ struct Edge {
     std::int32_t _weight = 0;
     std::string  _name   = "unnamed edge"; // custom edge name
 
+    property_map _uiConstraints{};
```
ivan-cukic, line 75: Could use docs what uiconstraints are for edges

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -67,14 +67,14 @@ inline std::string format(GraphLike auto const& graph) {
     return os.str();
 }
 
-std::expected<std::shared_ptr<BlockModel>, Error> findBlock(GraphLike auto const& graph, BlockLike auto& what, std::source_location location = std::source_location::current()) noexcept {
+std::expected<std::shared_ptr<BlockModel>, Error> findBlock(GraphLike auto const& graph, BlockLike auto const& what, std::source_location location = std::source_location::current()) noexcept {
```
ivan-cukic, line 70: We should not mix east-const and west-const in a single project

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -90,6 +90,19 @@ std::expected<std::shared_ptr<BlockModel>, Error> findBlock(GraphLike auto const
     return std::unexpected(Error(std::format("Block {} not found in:\n{}", uniqueBlockName, format(graph)), location));
 }
 
+std::expected<std::size_t, Error> blockIndex(GraphLike auto const& graph, std::string_view uniqueBlockName, std::source_location location = std::source_location::current()) noexcept {
```
ivan-cukic, line 93: It would be nice if this was unified with the `findBlock`. For example, a `namespace detail` find that returns iterator, and then findBlock can be implemented to extract the pointer from the iterator and blockIndex could be std::distance(....)

core/include/gnuradio-4.0/BlockModel.hpp
RalphSteinhagen: This slipped in for a follow-up PR (UTF-8 graph graphing). The usage is similar to that of the one in BlockModel: meta-information that is not necessary for the signal processing but more the display (e.g. colour, width of edge, edge drawing constraints, etc.).

Will add this to the next PR.

core/include/gnuradio-4.0/BlockModel.hpp
```diff
@@ -840,6 +881,111 @@ class BlockWrapper : public BlockModel {
...
+        }
+        const auto tail = s.substr(head.size() + 1UZ);
+
+        std::size_t v{};
+        const char* first = std::to_address(tail.cbegin());
+        const char* last  = std::to_address(tail.cend());
+        if (auto [ptr, ec] = std::from_chars(first, last, v); ec == std::errc{} && ptr == last) {
+            return {head, v};
+        }
+        return {head, std::nullopt};
```
wirew0rm, line 901:
```suggestion
        return {head, gr::meta::Invalid_index};
```
Why not directly return invalid index here? This should make the code around l917-924 a lot simpler.

core/include/gnuradio-4.0/Graph.hpp
wirew0rm: I think this coudld better be handled by adding `QualifierAlignmentStyle=Left` (if that's what we agree on) to `.clang-format` (and probably doing a global formatter pass). As this line was inconsistent even before the change, there are probably more. So I think it's good to point this out and it's easy to fix, but I wouldn't block on this without at the same time automating it for the future.

core/include/gnuradio-4.0/BlockModel.hpp
RalphSteinhagen: `gr::meta::Invalid_index` has, unfortunately, a dual-use/meaning in the connection API. I used optional to distinguish this from the use in this helper. When revising the conenction API, we could tackle this.

PR #660, opened 2025-09-15
wip: Fix crash when calling Scheduler::exchange twice

iamsergio:
Here's the crashing tests while the fix is being written


RalphSteinhagen (approved):
Thanks @iamsergio for identifying and pragmatic fix!

PR #661, opened 2025-09-16
Better handling of moved-from destruction

iamsergio:
- moved-from ~Block was running a lot of code related to active state, so move state to stopped when moved-from

- Remove code duplication between StateMachine move-ctor and move-assign

- moved-from ~Reader was decrementing the _reader_count while move-ctor didn't increment it. Instead, don't do anything.


ivan-cukic:
I like it, I think we used to count how many readers a buffer has wrongly before this.

PR #670, opened 2025-09-24
Remove assignment operators from the core GR classes

ivan-cukic:
- Assignment was often broken (UB or semantically impossible), this patch removes those operations
- Where assignment is needed, we're now using a simple implementation of a std::indirect-like wrapper type
- const fields in classes that need to be move-constructible are replaced with gr::meta::immutable which is a movable const-like wrapper type
- functions that deal only with graph topology no longer instantiate full Graphs with Block members, ports, etc.


RalphSteinhagen (approved):
OK as discussed in the channel. Thanks for the additional iterations/adjustments. 👍 

Can be merged once the CI passes (probably unrelated but seems to be systematically triggered by this PR).

PR #675, opened 2025-11-23
gr::Tensor[View], SIMD GEMM/GEMV, and pmr::vector support

RalphSteinhagen:
This PR introduces a unified **Tensor / TensorView** abstraction (static, semi-static, and fully dynamic) plus **SIMD-accelerised GEMM/GEMV** helpers, backed by a lightweight `gr::pmr::vector` and allocator-migration utilities. Mandatory unit-tests aside, it adds a `bm_tensor` benchmark to aid performance improvements/regression testing for GEMM/GEMV.

---

### User-facing API (TL;DR)

```c++
#include <gnuradio-4.0/Tensor.hpp>
#include <gnuradio-4.0/TensorMath.hpp>

using gr::Tensor;
using gr::TensorView;
using gr::math::gemm;
using gr::math::gemv;
using enum gr::math::TransposeOp;

// fully dynamic shape (rank & extents at runtime)
Tensor<float> A({M, K});
Tensor<float> B({K, N});
Tensor<float> C({M, N});

// static shape (rank & extents fixed at compile time)
Tensor<double, 2, 3> staticMat;

// static square matrix – designed to be trivially copyable & PMR/USM friendly
Tensor<float, N, N> square;

// element access (row-major)
A[0, 0] = 1.0f;
A[0, 1] = 2.0f;
auto v  = A[3];              // linear index

// non-owning view (shares storage, no copy)
TensorView<float> view(A);
view[0, 0] = 42.0f;          // writes into A

// interoperabilkity with std::vector and other STL containers
std::vector<float> vdata = static_cast<std::vector<float>>(A);
Tensor<float> T_from_vec(vdata, {M, K});

// basic math example:
gemm(C, A, B); // general dense matrix–matrix: C = A * B
gemm<NoTrans, NoTrans>(gr::math::cpu_policy{}, C, A, B, /*alpha=*/1.0f,  /*beta=*/0.0f); // explicit CPU policy

gemv(y, A, x); // matrix–vector: y = A * x
gemv(y, A, x, /*alpha=*/0.5f, /*beta=*/2.0f); // y = 0.5*A*x + 2*y

C = A * B; // B being another tensor, scalar, similar for  +-/
TensorOps<T>::sum(TensorOf<T> auto const& A) -> T;        // sum of all elements
TensorOps<T>::product(TensorOf<T> auto const& A) -> T;  // product of all elements
TensorOps<T>::mean(TensorOf<T> auto const& A) -> T;
TensorOps<T>::variance(TensorOf<T> auto const& A) -> T;
TensorOps<T>::std_dev(TensorOf<T> auto const& A) -> T;
TensorOps<T>::min(TensorOf<T> auto const& A) -> T;
TensorOps<T>::max(TensorOf<T> auto const& A) -> T;
TensorOps<T>::arg_min(TensorOf<T> auto const& A) -> std::size_t;
TensorOps<T>::arg_max(TensorOf<T> auto const& A) -> std::size_t;
TensorOps<T>::sum_axis(TensorOf<T> auto const& A, std::size_t axis) -> T; // aka. projection
TensorOps<T>::mean_axis(TensorOf<T> auto const& A, std::size_t axis) -> T; // aka. projection
// [..]
```

All `Tensor<T, Ex...>` and `TensorView<T, Ex...>` variants (dynamic or static) are designed to work with **pmr memory resources** (host side), and be **trivially copyable** wherever possible so the *same* tensor type can be used on host and device when memory is shared via **USM (SYCL)** or **Unified Memory (CUDA)**.

### Tensor design and API

`core/include/gnuradio-4.0/Tensor.hpp` provides:

* `Tensor<T, Ex...>` – **owning** tensor

  * **Fully dynamic**: `Ex...` empty; rank & extents at runtime
  * **Semi-static**: static rank (e.g. `<dynamic_extent, dynamic_extent>`), extents at runtime
  * **Fully static**: all extents known at compile time (e.g. `<2, 3>`, `<N, N>`)
* `TensorView<T, Ex...>` – **non-owning view**, same shape semantics

Key properties / design goals:

* **Canonical tensor type for GR4**
  replaces ad-hoc `std::vector + extents + strides` combinations. always row-major and contiguous for predictable SIMD/BLAS/GPU behaviour.
* **Static vs dynamic storage**

  * fully static: uses an internal `std::array<T, N>`; no heap allocation, trivial layout (usually used for small matrices).
  * dynamic/semi-static: uses `gr::pmr::vector<T, managed>` as the storage backend for PMR-aware allocation and allocator migration.
* **Host/device friendliness** (CPU- & GPU-based algorithms)

  * `Tensor<T, N, N>` and the view types are designed to be **trivially copyable** where feasible.
  * explicitly structured to work with pmr on the host while remaining compatible with **USM / Unified Memory** for device sharing.
  * the aim is that the same Tensor type can be used in CPU-only blocks and GPU-backed kernels without API forks.
* **Shape and indexing**

  * `rank()`, `size()`, `extents()`, `extent(i)`, `strides()`
  * Element access:

    * `operator[](i, j, k, ...)` – fast multi-index, unchecked
    * `at(i, j, ...)` – checked, throws on out-of-range
    * `at(std::span<const std::size_t>)` – generic multi-index
  * Iteration:

    * contiguous random-access iterators (`begin()`, `end()`, etc.)
* **Conversions / interop**

  * To STL containers:

    * `static_cast<std::vector<T>>(tensor)`
    * `static_cast<std::pmr::vector<T>>(tensor)`
    * `static_cast<std::array<T, N>>(tensor)`, `std::deque<T>`, …
  * From containers / ranges:

    * constructors/CTAD from `std::vector`, `std::array`, generic ranges.
  * mdspan alignment:

    * `to_mdspan()` returns a `std::mdspan` over the tensor’s storage, using its extents/strides.
* **Concepts and traits**

  * `TensorLike`, `TensorOf<T>`, `StaticTensorOf<T>`, `StaticRankTensorOf<T>`, `DynamicTensorOf<T>`, `TensorViewOf<T>`.
  * `tensor_traits` exposes rank, static/dynamic-ness, and view/owning information for template code and block APIs.

---

### TensorMath: GEMM/GEMV and basic ops

`core/include/gnuradio-4.0/TensorMath.hpp` introduces:

#### Execution policies

* `cpu_policy` – current implementation target.
* `gpu_policy<Queue>` – placeholder for SYCL/CUDA queues and other devices.
* concepts: `ExecutionPolicy`, `CpuExecutionPolicy`, `GpuExecutionPolicy`. (TODO: needs further generalisation)

All GEMM/GEMV entry points take (or deduce) an `ExecutionPolicy` so GPU backends can be wired in later without changing call sites.

#### GEMM – matrix–matrix

  ```c++
  gemm(C, A, B);  // policy auto-detected as CPU

  gemm<NoTrans, NoTrans>(gr::math::cpu_policy{},
      C, A, B, /*alpha=*/1.0f, /*beta=*/0.0f);
  ```

Implementation details in `math/gemm_simd.hpp`:

#### GEMV – matrix–vector

  ```c++
  gemv(y, A, x);                     // y = A*x
  gemv(y, A, x, alpha, beta);        // y = alpha*A*x + beta*y
  ```

Implementation in `math/gemv_simd.hpp`:

#### Element-wise ops and reductions

TensorMath also provides:

* element-wise operations with shape checks and broadcasting:

  * `operator+`, `-`, `*`, `/` for Tensor–Tensor and Tensor–scalar combinations.
* Reductions:

  * `sum`, `mean`, `min/max`, `argmin/argmax`,
  * `sum_axis(tensor, axis)` → reduces along the given axis (1D result).

---

### pmr::vector and allocator migration

`core/include/gnuradio-4.0/PmtCollections.hpp`:

* `gr::pmr::vector<T, bool managed>`:

  * PMR-aware contiguous container used as the dynamic backend for Tensor and PMT collections.
  * `managed=true` → owns storage (normal pmr vector).
  * `managed=false` → non-owning façade (no deallocation; used for views / foreign memory).
  * STL-like interface: iterators, `push_back`, `emplace_back`, `resize`, `assign`, comparisons, etc.

`core/include/gnuradio-4.0/MemoryAllocators.hpp`:

* `gr::allocator::pmr::migrate<T>(...)`:

  * move a contiguous block of `T` between two `std::pmr::memory_resource`s.
  * fast path for trivially copyable types (`memcpy`),
    otherwise uses `uninitialized_move_n` / element-wise copy with strong exception safety.
  * guarantees:

    * source remains intact on exception,
    * partially constructed targets are destroyed and deallocated.

`qa_PmtCollections.cpp` and `qa_MemoryAllocators.cpp` validate:

* ownership vs non-ownership (`managed` flag),
* non-trivial and throwing-copy types,
* exception safety and correct destruction counts.

---

### Benchmarks and integration
image

**N.B. gemm/gemv implementations are subject to further improvements.**

Planned follow-ups (not in this PR):

* integrate into `gr::pmt::Value` abstraction (smaller binary size footprint, faster compile-times, MCU-compatibility, ...)
* switch from `pmtv::pmt` -> `gr::pmt::Value` in settings and tag handling (goal: near-zero copy tag handling).
* implement GPU-backed GEMM/GEMV via `gpu_policy` and SYCL/CUDA queues using shared USM/Unified Memory buffers.
* extend TensorMath with SVD, convolution, and other ML/RF-relevant primitives.


core/include/gnuradio-4.0/TensorMath.hpp
```diff
@@ -0,0 +1,785 @@
...
+struct cpu_parallel_policy {
+    static constexpr bool is_gpu       = false;
+    static constexpr bool use_parallel = true;
+    std::size_t           num_threads  = 0UZ; // 0 = auto-detect
+};
+
+template<typename QueueType>
+struct gpu_policy { /// GPU policy (for future SYCL/CUDA support)
+    static constexpr bool is_gpu = true;
+    QueueType&            queue;
```
ivan-cukic, line 99: To make sure nobody misuses this, it should probably have `=delete`d copy and move operations. Or, at least, `operator=`

core/include/gnuradio-4.0/TensorMath.hpp
```diff
@@ -0,0 +1,785 @@
...
+};
+
+enum class MatrixSize : std::int8_t {
+    Tiny,   /// typ. 32x32
+    Medium, /// typ 4096x4096
+    Large   /// > Medium
+};
+
+struct cpu_policy {
+    static constexpr bool is_gpu       = false;
```
ivan-cukic, line 86: Maybe an enum instead of bool - so that we can have other types of devices (maybe to follow opencl device types)?

core/include/gnuradio-4.0/TensorMath.hpp
```diff
@@ -0,0 +1,785 @@
...
+template<typename T, std::size_t... Ex>
+struct TensorOps {
+    using TensorType = Tensor<T, Ex...>;
+
+    [[maybe_unused]] static constexpr TensorType& add_inplace(TensorType& self, const TensorType& other) {
+        auto same_shape = [](std::span<const std::size_t> a1, std::span<const std::size_t> b1) { return a1.size() == b1.size() && std::equal(a1.begin(), a1.end(), b1.begin()); };
+        if (!same_shape(self.extents(), other.extents())) {
+            throw std::runtime_error("Tensor dimensions must match for element-wise operations");
+        }
+#if defined(__GLIBCXX__)
```
ivan-cukic, line 143: What is the reason for not using unseq on other lib implementations?

core/include/gnuradio-4.0/TensorMath.hpp
```diff
@@ -0,0 +1,785 @@
...
+#endif
+        return self;
+    }
+
+    [[nodiscard]] static constexpr TensorType divide_elementwise(const TensorType& a, const TensorType& b) {
+        TensorType result(a);
+        return divide_elementwise_inplace(result, b);
+    }
+
+    [[nodiscard]] static constexpr T sum(const TensorType& self) { return std::accumulate(self.begin(), self.end(), T{0}); }
```
ivan-cukic, line 243: Any reason for using accumulate here, and reduce in product? Also, would `unseq` for reduce make a difference here?

core/include/gnuradio-4.0/TensorMath.hpp
ivan-cukic: Also, maybe a definition of a default exec policy could be added to the top so that we can avoid the many `if defined ...`s in this file.

core/include/gnuradio-4.0/PmtCollections.hpp
```diff
@@ -0,0 +1,485 @@
...
+
+    vector(vector&& other) noexcept : _resource{other._resource}, _size{other._size}, _capacity{other._capacity}, _data{other._data} {
+        other._data     = nullptr;
+        other._size     = 0;
+        other._capacity = 0;
+    }
+
+    vector& operator=(const vector& o)
+    requires(std::is_copy_constructible_v<T>)
+    {
```
ivan-cukic, line 137: Not strong exception safe -- if allocate throws an exception, we have lost the old data.

core/include/gnuradio-4.0/PmtCollections.hpp
```diff
@@ -0,0 +1,485 @@
...
+        }
+    }
+
+    template<class It, class End>
+    requires std::input_iterator<It> && std::sentinel_for<End, It>
+    void assign(It first, End last) {
+        const auto new_size = static_cast<std::size_t>(std::ranges::distance(first, last));
+        clear();
+        if (new_size > 0) {
+            _data     = static_cast<T*>(_resource->allocate(new_size * sizeof(T), alignof(T)));
```
ivan-cukic, line 434: allocate and dealocate are called in quite a few places. Might be useful to move into an inlined dedicated function.

core/include/gnuradio-4.0/PmtCollections.hpp
```diff
@@ -0,0 +1,485 @@
...
+    void assign(std::span<const T> s) { assign(s.begin(), s.end()); }
+    void assign(const std::pmr::vector<T>& src) {
+        clear();
+        if (src.size() > 0) {
+            _data     = static_cast<T*>(_resource->allocate(src.size() * sizeof(T), alignof(T)));
+            _capacity = src.size();
+            if constexpr (std::is_trivially_copyable_v<T>) {
+                std::memcpy(_data, src.data(), src.size() * sizeof(T));
+            } else {
+                std::uninitialized_copy_n(src.data(), src.size(), _data);
```
ivan-cukic: No try-catch here as for uninit_copy above?

core/include/gnuradio-4.0/TensorMath.hpp
RalphSteinhagen: The `unseq` support isn't great for libc++ and I needed to guard the versions w/ and w/o. If it's not being used, then mostly for lazyness. ;-)

core/include/gnuradio-4.0/TensorMath.hpp
RalphSteinhagen: This is WIP and MVP only. This needs to be certainly improved and extended when integrating the SYCL/CUDA abstraction.

core/include/gnuradio-4.0/TensorMath.hpp
RalphSteinhagen: `unseq` support is flaky in libc++ (as is `std::uninitialized_move(...)`)  what I just learned. 

The idea is to remove the if-else macros once the libc++ bugs are fixed.

core/include/gnuradio-4.0/PmtCollections.hpp
RalphSteinhagen: Opted for explicit use of `pmr->allocate(...)` since this is official C++ API and didn't want to hide that. I feel that it would be harder to reason...

core/include/gnuradio-4.0/PmtCollections.hpp
RalphSteinhagen: added.

RalphSteinhagen:
Needed to add a workaround for some semi-known libc++ (clang, release) bugs related to `std::uninitialized_move(...)`.
Also added some fixes addressing the review comments, as well as a best-effort `bm_tensor.py` Python benchmark:

image

PR #677, opened 2025-11-27
 Owning `gr::pmt::Value` Polymorphic Type with Monadic Access

RalphSteinhagen:
This PR introduces `gr::pmt::Value`, a compact owning polymorphic type container that replaces the `rva::variant`-based `pmtv::pmt`. The primary goals are reducing compile-time overhead, eliminating binary bloat, and providing exception-free operation for embedded targets.

A future `ValueView` will provide non-owning access over YaS wire-format buffers for zero-copy/zero-allocation tag processing. Both `Value` and `ValueView` will share the same access API.

---
## TL;DR – Typical Usage

```cpp
using gr::pmt::Value;

// Construction
Value v1{std::int64_t{42}};
Value v2{std::string_view{"hello"}};
Value v3{std::complex<float>{1.0f, 2.0f}};
```

#### 1. Classic get_if<T>() C++ pattern

```cpp
if (auto* p = v1.ref_if<std::int64_t>();) {
    *p += 1;
} else {
    // nullptr on mismatch -> handle error
}

if (auto opt = v2.get_if<std::pmr::string>();  // std::optional<T>) {
    std::println("string = {}", *opt);
}
```

#### 2. `value_or<T>` – copy / borrow / take, encoded in `T`

```cpp
Value v{std::int64_t{42}};
std::int64_t fallback = 0;

// copy
auto x  = v.value_or<std::int64_t>(-1);            // x = 42, v still holds 42

// nutable borrow
auto& rx = v.value_or<std::int64_t&>(fallback);    // rx refers to stored 42

// const borrow
const auto& cx = v.value_or<const std::int64_t&>(fallback);

// move / take (Value becomes monostate on match)
Value vs{std::pmr::string{"world"}};
std::pmr::string s = vs.value_or<std::pmr::string&&>({});      // s == "world", vs -> monostate, (move)
```

#### 3. Monadic helpers: `or_else`, `transform`, `and_then`

```cpp
// lazy fallback: only constructs default if type does not match
auto val = v.or_else<std::int64_t>([] { return std::int64_t{123}; });

// transform to something else, default on mismatch
auto len = v2.transform<std::pmr::string>([](auto& s) { return s.size(); });

// transform with explicit fallback
auto len2 = v2.transform_or<std::pmr::string>([](auto& s) { return s.size(); }, std::size_t{0});

// monadic chaining: only runs on matching type
Value doubled = v.and_then<std::int64_t>([](auto& x) { return Value{x * 2}; });
```

The same API will be available on `ValueView` for zero-copy/zero-allocation access to serialized PMT data; switching between owning (`Value`) and non-owning (`ValueView`) will not require changes to user-level code.

---

## For Reviewers -- Motivation

The previous `pmtv::pmt` implementation based on `rva::variant` had several issues:

  1. **Compile-Time Overhead**: Recursive variant instantiation pulls in substantial template machinery, significantly impacting build times in a large library context.

  2. **Binary Bloat**: The variant type machinery inflates symbol tables and binary size, problematic for embedded deployments.

  3. **Exception Dependency**: `std::bad_variant_access` requires exception support, incompatible with `-fno-exceptions` builds common on MCU targets.

  4. **RTTI Dependency**: Runtime type information requirements conflict with `-fno-rtti` builds.

This `gr::pmt::Value` implementation addresses these issues with a simpler, hand-rolled type-erased container. While `Value` can cross CPU/MCU/GPU boundaries when using a USM-compatible PMR resource, the upcoming `ValueView` over YaS wire-format buffers will be the preferred approach for zero-copy heterogeneous access.

---

### Storage Layout

`Value` uses a 24-byte footprint:
- 1 byte: type info (4 bits `ValueType` + 4 bits `ContainerType`)
- 8 bytes: inline union storage for scalars
- 8 bytes: PMR memory resource pointer
- 7 bytes: padding

```cpp
static_assert(sizeof(gr::pmt::Value) <= 24);
```

Scalars (bool, int8–64, uint8–64, float, double) are stored inline with zero heap allocation. Complex numbers, strings, tensors, and maps allocate through the PMR resource.

### Type System

```cpp
enum class ValueType : uint8_t {
    Monostate, Bool,
    Int8, Int16, Int32, Int64,
    UInt8, UInt16, UInt32, UInt64,
    Float32, Float64,
    ComplexFloat32, ComplexFloat64,
    String, Value  // Value used for Tensor<Value> and Map
};

enum class ContainerType : uint8_t {
    Scalar, Complex, String, Tensor, Map
};
```

---

## Ownership Semantics

The `value_or<T>()` template parameter encodes the **lifetime relationship**, not just access mechanics:

| Category | Meaning | Value's Role After Call |
|----------|---------|-------------------------|
| `T` | "Give me a copy" | Retains ownership |
| `T&` | "Let me modify yours" | Retains ownership |
| `const T&` | "Let me borrow for read" | Retains ownership |
| `T&&` | "Give it to me" | Relinquishes → **monostate** |

This design eliminates the need for separate `take()` methods — `value_or<T&&>()` is the ownership transfer operation.

### Type Strictness

The fallback parameter must match the raw type exactly. This prevents subtle bugs from implicit conversions:

```cpp
Value v{3.14f};  // float
float  ok  = v.value_or<float>(0.0f);       // ✓ compiles
// double bad = v.value_or<double>(0.0);    // ✗ compile error: float ≠ double
```

---

## Complete Monad Set

| Method | Purpose | On Match | On Mismatch |
|--------|---------|----------|-------------|
| `value_or<T>` | Copy with fallback | Return copy | Return fallback |
| `value_or<T&>` | Mutable borrow | Return reference | Return fallback ref |
| `value_or<const T&>` | Const borrow | Return const ref | Return fallback ref |
| `value_or<T&&>` | Ownership transfer | Move + reset | Return fallback |
| `or_else<T>` | Lazy fallback | Return value | Invoke factory |
| `transform<T>` | Apply function | Return result | Return default |
| `transform_or<T>` | Apply with fallback | Return result | Return fallback |
| `and_then<T>` | Monadic bind | Return Value | Return monostate |
| `get_if<T>` | Optional access | Return optional | Return nullopt |

All methods with `T&&` variant reset Value to monostate on successful match.

---

## Exception-Free Design

Type mismatches on unchecked accessors use:

```cpp
[[noreturn, gnu::cold]] inline void value_type_mismatch() noexcept {
    assert(false && "gr::pmt::Value type mismatch");
    std::unreachable();
}
```

- **Debug**: Assert fires with diagnostic
- **Release**: UB enables optimisation
- **`-fno-exceptions`**: Fully compatible

---

## Binary Size Optimisation

Explicit template instantiation in `Value.cpp` for all supported types:

```cpp
#define X(T) \
    template bool Value::holds<T>() const noexcept; \
    template T* Value::ref_if<T>() noexcept; \
    // ...
GR_PMT_VALUE_TENSOR_ELEMENT_TYPES
#undef X
```

Methods that take callables (`or_else`, `transform`, `and_then`) remain header-only because the callable type varies per call site.

---

## Relationship to Future Work

This `Value` is the **owning** polymorphic type. A companion `ValueView` will provide:
- Non-owning access over YaS wire-format buffers
- Zero-copy tag processing in SPMC ring buffers
- The same monadic access API (`value_or`, `or_else`, `transform`, `and_then`)

Together they form the `gr::pmt` type system:
- `Value`: owning, for construction and manipulation
- `ValueView`: non-owning, for zero-copy access over serialised data


core/include/gnuradio-4.0/Value.hpp
```diff
@@ -0,0 +1,643 @@
...
+        ComplexFloat32 = 12U, ComplexFloat64 = 13U,
+        String = 14U,
+        Value = 15U  // for Tensor<Value> and Map<string,Value>
+    };
+    // clang-format on
+    enum class ContainerType : uint8_t { Scalar = 0U, Complex = 1U, String = 2U, Tensor = 3U, Map = 4U };
+
+    using Map = std::pmr::unordered_map<std::pmr::string, Value>; // TODO: replace with std::flat_map or other more simpler key-value map once available (libc++ >=20, stdlibc++ >=15)
+
+    uint8_t _type_info{0}; // lower 4 bits: ValueType, upper 4 bits: ContainerType
```
ivan-cukic: Would this be easier to handle? (the compiler would do the bitmask magic and the getters would be simpler)
```
int _value_type : 4;
int _container_type: 4;
```

core/src/Value.cpp
```diff
@@ -0,0 +1,701 @@
...
+}
+
+Value::Value(Value&& other) noexcept : _type_info(other._type_info), _storage(other._storage), _resource(other._resource) {
+    other.set_types(ValueType::Monostate, ContainerType::Scalar);
+    other._storage.u64 = 0UZ;
+}
+
+Value::~Value() { destroy(); }
+
+Value& Value::operator=(const Value& other) {
```
ivan-cukic, line 375: My usual complaint that `operator=` needs to use copy and swap idiom for strong exception guarantees.

core/src/Value.cpp
```diff
@@ -0,0 +1,701 @@
...
+void Value::copy_from(const Value& other) {
+    // Note: Copy uses THIS object's _resource (target's allocator), not other's.
+    // This allows copying between different memory domains (e.g., host ↔ device).
+    switch (other.container_type()) {
+    case ContainerType::Scalar: _storage = other._storage; break;
+
+    case ContainerType::Complex:
+        if (other.value_type() == ValueType::ComplexFloat32) {
+            void* mem    = _resource->allocate(sizeof(std::complex<float>), alignof(std::complex<float>));
+            _storage.ptr = new (mem) std::complex<float>(*static_cast<const std::complex<float>*>(other._storage.ptr));
```
ivan-cukic: Could std::construct_at be used instead of placement new? (since std::destroy_at is used, would be more symmetric)

core/src/Value.cpp
```diff
@@ -0,0 +1,701 @@
...
+void Value::set_types(ValueType vt, ContainerType ct) noexcept { _type_info = static_cast<uint8_t>(static_cast<uint8_t>(ct) << 4) | static_cast<uint8_t>(vt); }
+
+void Value::destroy() noexcept {
+    if (is_monostate() || container_type() == ContainerType::Scalar || _storage.ptr == nullptr) {
+        return;
+    }
+    switch (container_type()) {
+    case ContainerType::Complex:
+        if (value_type() == ValueType::ComplexFloat32) {
+            _resource->deallocate(_storage.ptr, sizeof(std::complex<float>), alignof(std::complex<float>));
```
ivan-cukic: `std::destroy_at` should be called even for types that have an empty destructor.

core/include/gnuradio-4.0/Value.hpp
```diff
@@ -0,0 +1,643 @@
...
+#include <gnuradio-4.0/Tensor.hpp>
+
+namespace gr::pmt {
+
+class Value;
+
+namespace detail {
+[[noreturn, gnu::cold]] inline void value_type_mismatch() noexcept {
+    assert(false && "gr::pmt::Value type mismatch");
+    std::unreachable();
```
ivan-cukic, line 27: `std::terminate`

Getting to `std::unreachable` is UB.

core/include/gnuradio-4.0/Value.hpp
RalphSteinhagen: :+1:  done/incorporated.

core/src/Value.cpp
RalphSteinhagen: :+1: done and incorporated.

core/src/Value.cpp
RalphSteinhagen: :+1: done and incorporated.

core/src/Value.cpp
RalphSteinhagen: :+1: done and incorporated.

core/include/gnuradio-4.0/Value.hpp
RalphSteinhagen: used `std::abort();` in favour to avoid having to include `<exception>`.

PR #679, opened 2025-12-01
File I/O.

drslebedev:
## File I/O for gnuradio

This module provides unified `Reader`/`Writer` abstractions for various kinds of I/O in gnuradio:

- Local files via `file:/` URIs (native + MEMFS under Emscripten)  
- HTTP(S) `GET` / long-polling via `http://` and `https://`  
- In-memory sources (`readAsync(std::span<const std::uint8_t>)`)  
- `dialog:/` URIs (open file dialog → memory or local file)  
- `download:/` URIs (WASM "download to disk")

## Reader

The `Reader` API is designed for asynchronous, chunked consumption of data.

### Configuration

`ReaderConfig` controls:

- `chunkBytes` - maximum chunk size when pushing data into the buffer  
- `offset` - optional starting offset into the file/source  
- `httpTimeoutMs` - HTTP timeout
- `longPolling` - enable repeated HTTP requests for long-polling streams  

### Core API

- `readAsync(uri, config)`  returns a `Reader` backed by a `CircularBuffer`. The actual I/O happens on a worker thread.
- `poll(callback, maxSize, doWait)`  delivers chunks or errors via `Reader::PollResult`. The callback is invoked with either: `data` span (on success), or  `Error` (on failure),  and a final message (`isFinal == true`) at the end of the stream.
- `get()`: convenience helper: blocks on a worker thread, consumes all chunks via `poll()`, and concatenates everything into a single `std::vector<std::uint8_t>`.
- `cancel()`  requests a `soft cancel`.  Under Emscripten, cancellation is best-effort only (we cannot synchronously abort `emscripten_fetch` from another thread).

### Example: local file

```cpp
auto rExp = readAsync("file:/tmp/test.bin", ReaderConfig{.chunkBytes = 4096});
if (!rExp) {
    throw std::runtime_error(rExp.error().message);
}

auto reader = std::move(*rExp);

// collect whole file
auto dataExp = reader.get();
if (!dataExp) {
    throw std::runtime_error(dataExp.error().message);
}
auto& bytes = dataExp.value();
```

### Example: streaming with poll

```cpp
auto rExp = readAsync("file:/tmp/test.bin", {.chunkBytes = 4096});
if (!rExp) {
    throw std::runtime_error(rExp.error().message);
}
auto reader = std::move(*rExp);

// poll individual chunks in while loop
bool       finished = false;
ReadResult readResult;
  while (!finished) {
  reader.poll(
      [&](const Reader::PollResult& res) {
          finished = res.isFinal;
          if (res.data) {
              // process res.data.value()
          } else {
              std::println("Error: {}", res.data.error().message);
          }
      }, /*maxSize=*/8192, /*doWait=*/true
  );
}
```

## Writer

The Writer API covers local file writes and HTTP POST.

### Configuration
`WriterConfig` supports:
 - `WriteMode::overwrite`
 - `WriteMode::append` (only valid for local files, not HTTP)

### Core API
 - `writeAsync(uri, bytes, config)` runs on a worker pool and returns a `Writer` that can be waited on via `wait()` and queried via `result()`.
 - `write(uri, bytes, config)`: Synchronous convenience wrapper that internally uses `writeAsync()` + `wait()` + `result()`.
 - For local files, `WriteResult` always has: `httpStatus == 0` and empty `httpResponseBody`

## Native HTTP implementation

On native builds, HTTP support is implemented via CPR when `GR_FILEIO_CPR_AVAILABLE = 1`:
 - Uses `cpr::Session` for HTTP GET and POST.
 - Long-polling is implemented by repeatedly calling `runHttpGetNativeOnce()`.
 - HTTP errors (status >= 400) or CPR errors are reported as `gr::Error`.

## CMake controls for native HTTP (CPR + libcurl):
- `GR_ENABLE_NATIVE_HTTP=ON`: Require libcurl. If libcurl is found, CPR is enabled and GR_NATIVE_HTTP_ENABLED=1; otherwise CMake fails with an error.
- `GR_ENABLE_NATIVE_HTTP=OPTIONAL`: Enable CPR only if system libcurl is found. If libcurl is missing, GR_NATIVE_HTTP_ENABLED=0 and HTTP(S) operations are treated as "disabled at build time".
- `GR_ENABLE_NATIVE_HTTP = OFF`: CPR is never fetched; GR_NATIVE_HTTP_ENABLED=0 and all HTTP(S) operations return "disabled at build time".

## Emscripten implementation
On Emscripten targets:
 - Uses `emscripten_fetch` + callbacks for GET / POST, with optional `longPolling` (recursive re-fetch).
 - Blocking wait() / write() is explicitly forbidden on the main WASM runtime thread: Attempts are detected and result in an error/warning.
 - download:/ URIs trigger a JS `download` in the browser.

# Follow-up issues:
* https://github.com/fair-acc/gnuradio4/issues/685
* https://github.com/fair-acc/gnuradio4/issues/684
* https://github.com/fair-acc/gnuradio4/issues/683

### DIFF
--- a/algorithm/include/gnuradio-4.0/algorithm/fileio/README.md
+++ b/algorithm/include/gnuradio-4.0/algorithm/fileio/README.md
@@ -245,9 +245,7 @@ On the web, browsers don’t let you write arbitrary paths on the user’s disk.
 
 - HTTP uses `emscripten_fetch`; `ReaderConfig::longPolling` re-issues requests in a loop.
 
-- Blocking calls on **main runtime thread**:
-
-  - `Reader::wait()`, `Writer::wait()`, `write()` -> refuse to block, return early / error.
+- Blocking calls on **main runtime thread**: `Reader::wait()`, `Writer::wait()`, `write()` -> refuse to block, return early / error.
 
 - `download:/` is implemented via JS in the browser.
 


CMakeLists.txt
```diff
@@ -112,6 +112,16 @@ set(GR_MAX_WASM_THREAD_COUNT
     60
     CACHE STRING "Max number of threads for WASM pthread pool")
 add_compile_definitions(GR_MAX_WASM_THREAD_COUNT=${GR_MAX_WASM_THREAD_COUNT}) # propagate to C++
+set(GR_INSTALL_CPR
```
wirew0rm: This seems missleading, "INSTALL" imho implies that this would install the cpr header to the system? Maybe something like `ENABLE_FILEIO_CPR`.

algorithm/CMakeLists.txt
```diff
@@ -9,6 +10,47 @@ target_link_libraries(
...
+
+  if(NOT
+     GR_INSTALL_CPR
+     STREQUAL
+     "OFF")
+
+    find_package(CURL QUIET)
+
+    # ON -> always install CPR; OPTIONAL -> only install CPR if system libcurl is available
+    if(GR_INSTALL_CPR STREQUAL "ON" OR (GR_INSTALL_CPR STREQUAL "OPTIONAL" AND CURL_FOUND))
```
wirew0rm: If it is set to `ON`, I would expect to get an error if libcurl is not available.

algorithm/CMakeLists.txt
```diff
@@ -9,6 +10,47 @@ target_link_libraries(
...
+        set(CPR_USE_SYSTEM_CURL
+            ON
+            CACHE BOOL "Use system libcurl in CPR" FORCE)
+      endif()
+
+      FetchContent_Declare(
+        cpr
+        GIT_REPOSITORY https://github.com/libcpr/cpr.git
+        GIT_TAG 1.12.0)
+      FetchContent_MakeAvailable(cpr)
```
wirew0rm: The cmake documentation recommends to only call `FetchContent_MakeAvailable` once per project after all dependencies have been declared. This will work but if one of the dependencies declared in the top-level cmake will start using cpr at some point they might override our depencency.

For me it also would feel nicer to declare dependencies as close as possible to their usage, but probably this should be moved together with all the other dependencies.
https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable (paragraph above the code samples)

algorithm/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 add_library(gnuradio-algorithm INTERFACE)
+target_link_libraries(gnuradio-algorithm INTERFACE httplib::httplib)
```
wirew0rm: Shouldn't httplib only be a dependency for the tests?

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
```diff
@@ -0,0 +1,1117 @@
...
+namespace gr::algorithm::fileio {
+
+// Read from file and http GET
+
+namespace detail {
+inline constexpr std::string_view kMessageDataKey = "data";
+}
+
+struct ReaderConfig {
+    std::size_t                chunkBytes    = 65536;
```
RalphSteinhagen: 65536 -> magic-number -> maybe use `std::numeric_limits<std::size_t>::max()` (safer on 32-bit/roll-over)

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
```diff
@@ -0,0 +1,1117 @@
...
+    std::optional<std::size_t> offset        = std::nullopt;
+    std::int32_t               httpTimeoutMs = 30000; // only for http(s)
```
RalphSteinhagen: please use `std::uint64_t` and ns as for other timing related consts (e.g. trigger).

e.g.
```cpp
std::uint64_t               httpTimeoutNanos = 30'000'000'000;  // 30 s time-out for http(s)
```

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
```diff
@@ -0,0 +1,1117 @@
...
+
+    void wait() {
+        if (_state == nullptr) {
+            return;
+        }
+
+#if __EMSCRIPTEN__
+        // Do NOT block main WASM thread.
+        if (emscripten_is_main_runtime_thread()) {
+            std::println("[fileio] wait() on main WASM thread is not allowed; returning immediately");
```
RalphSteinhagen: avoid using `std::println("...")` and either make it a hard error (e.g. std::abort())  or use `std::puts(..)`.

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
```diff
@@ -0,0 +1,1117 @@
...
+    if (in.bad()) {
+        in.close();
+        pushErrorFinal(state.get(), std::format("I/O error while reading file: {}", state->uri));
+    } else {
+        in.close();
+        pushFinal(state.get());
+    }
+}
+
+#if !defined(__EMSCRIPTEN__) && GR_FILEIO_CPR_AVAILABLE
```
RalphSteinhagen: Please rename `GR_FILEIO_CPR_AVAILABLE` to `GR_HTTP_ENABLED` since this is not strictly related to the CPR wrapper but the use of libcurl and httpLib. For Empscriten, this new `GR_HTTP_ENABLED` could be defaulted to 'true' or similar since the fetch-API etc. is always available.

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
```diff
@@ -0,0 +1,1117 @@
...
+
+    std::atomic<bool>                     done{false};
+    std::atomic<bool>                     cancelRequested{false};
+    std::atomic<std::size_t>              updateCounter{0};
+    std::expected<WriteResult, gr::Error> result; // Result is available when done == true
+
+    WriterState(std::string uri_, WriterConfig config_) : uri(std::move(uri_)), config(config_) {}
+
+    ~WriterState() {
+#ifndef NDEBUG
```
RalphSteinhagen, line 725: Good guard. Perhaps, put the '#ifndef ...' macro-guard around the destructor and allow C++ to define it's own i.e. `~WriterState() = default;`

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
```diff
@@ -0,0 +1,1117 @@
...
+        }
+        auto state = std::make_shared<WriterState>(std::string(uri), config);
+        state->data.assign(data.begin(), data.end());
+        gr::thread_pool::Manager::defaultIoPool()->execute([state]() mutable {
+            auto r = runHttpPostNative(state);
+            state->setResult(std::move(r));
+        });
+        return Writer{state};
+#else
+        return std::unexpected(gr::Error{"HTTP(S) disabled at build time"});
```
RalphSteinhagen: Perhaps change this to a `static_assert(gr::meta::always<T /* T to be replaced */, "HTTP(S) disabled at build time. See GR_HTTP_ENABLED for details.");`

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
```diff
@@ -0,0 +1,1117 @@
...
+    }
+#endif
+}
+
+[[nodiscard]] inline std::expected<WriteResult, gr::Error> write(std::string_view uri, std::span<const std::uint8_t> data, const WriterConfig& config = {}) {
+
+#if __EMSCRIPTEN__
+    // Do NOT block main WASM thread.
+    if (emscripten_is_main_runtime_thread()) {
+        return std::unexpected(gr::Error{"fileio::write() can not be called from main thread, it is not allowed to block main WASM thread"});
```
RalphSteinhagen, line 1174: To evaluate: stronger error since this is an API usage error.

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIoHelpers.hpp
```diff
@@ -0,0 +1,100 @@
...
+    return cb;
+}
+
+} // namespace detail
+
+inline void setDialogOpenCallback(detail::DialogOpenCallback cb) { detail::dialogOpenCallback() = std::move(cb); }
+
+namespace detail {
+[[nodiscard]] inline bool startsWithScheme(std::string_view uri) {
+    const auto p = uri.find(":");
```
RalphSteinhagen: should be `://` because `www.helloworld.org:20/` is also a valid URI and `www.helloworld.org` not a scheme but the authority.

URI scheme check should also be case insensitive, i.e. `https` and `HTTPS` should be treated as identical.

algorithm/test/FileioTestApp/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -s USE_PTHREADS=1 -s PTHREAD_POOL_SIZE=8")
+
+target_link_options(
+  FileioTestApp
+  PRIVATE
+  "SHELL:-s WASM=1"
+  "SHELL:-s EXPORTED_RUNTIME_METHODS=['ccall','cwrap','HEAPU8']"
+  "SHELL:-s EXPORTED_FUNCTIONS=['_malloc','_free','_fileio_dialog_on_bytes','_fileio_dialog_on_cancel','_fileio_dialog_on_error','_test_dialog_read','_test_save_file','_test_http_get','_test_http_get_404','_test_http_post','_main']"
```
RalphSteinhagen: Exported functions should be inlined in the header files since this is alot of "secret sauce" for the users that most should not/do not need to know.

e.g. in the **CMakeLists.txt** you need something like

```cmake

add_executable(FileioTestApp
    main.cpp
    fileio_api.cpp
)

if (EMSCRIPTEN)
    # Optional, but handy for conditional code in C++
    target_compile_definitions(FileioTestApp PUBLIC FILEIO_BUILD_WASM)

    target_link_options(
      FileioTestApp
      PRIVATE
        "SHELL:-s WASM=1"
        "SHELL:-s EXPORTED_RUNTIME_METHODS=['ccall','cwrap','HEAPU8']"        
        "SHELL:-s EXPORTED_FUNCTIONS=['_main','_malloc','_free']"  # only truly generic WASM exports
        "SHELL:-s ASSERTIONS=2"
        "SHELL:-s EXIT_RUNTIME=0"
        "SHELL:-s ALLOW_MEMORY_GROWTH=1"
    )
endif()

```

and in the API declaring headers, something like

```cpp
// header guards omitted

#include <cstddef>
#include <cstdint>

#if defined(__EMSCRIPTEN__)
    #include <emscripten/emscripten.h>
    #define FILEIO_EXPORT EMSCRIPTEN_KEEPALIVE
#else
    #define FILEIO_EXPORT
#endif

// C ABI for JS interoperability, but still be usable from C++
#ifdef __cplusplus
    #define FILEIO_EXTERN_C extern "C"
#else
    #define FILEIO_EXTERN_C
#endif

FILEIO_EXTERN_C {

FILEIO_EXPORT
void fileio_dialog_on_bytes(const std::uint8_t* data, std::size_t size);

FILEIO_EXPORT
void fileio_dialog_on_cancel();

FILEIO_EXPORT
void fileio_dialog_on_error(int error_code);

FILEIO_EXPORT
void test_dialog_read();

FILEIO_EXPORT
void test_save_file();

FILEIO_EXPORT
void test_http_get();

FILEIO_EXPORT
void test_http_get_404();

FILEIO_EXPORT
void test_http_post();

} // FILEIO_EXTERN_C
```

Needs some minor modifications...

CMakeLists.txt
```diff
@@ -112,6 +112,16 @@ set(GR_MAX_WASM_THREAD_COUNT
...
+    "ON"
+    CACHE STRING
+          "CPR HTTP client (libcurl wrapper): ON=always install, OFF=never, OPTIONAL=only if system libcurl is found")
```
RalphSteinhagen: `GR_INSTALL_CPR` -> maybe do not tightly couple this to CPR but more a generic `HTTP` or similar support.

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
```diff
@@ -0,0 +1,1117 @@
...
+            if (_state->cancelRequested.load(std::memory_order_acquire)) {
+                return;
+            }
+            auto expected = _state->updateCounter.load(std::memory_order_acquire);
+            _state->updateCounter.wait(expected, std::memory_order_acquire);
+        }
+    }
+
+    template<typename TCallback>
+    requires std::invocable<TCallback, const PollResult&> && std::same_as<std::invoke_result_t<TCallback, const PollResult&>, void>
```
RalphSteinhagen: No `const` for the PollResult. Please use: `std::invocable<TCallback, PollResult>`

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
```diff
@@ -0,0 +1,1117 @@
...
+        }
+        const std::size_t nReadSize = static_cast<std::size_t>(nRead);
+        if (nReadSize < values.size()) {
+            values.resize(nReadSize);
+        }
+        pushData(state.get(), std::move(values));
+        if (in.eof()) {
+            break;
+        }
+        remaining -= nReadSize;
```
RalphSteinhagen: *N.B. signed/unsigned mismatch possible.*

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
```diff
@@ -0,0 +1,1117 @@
...
+
+inline void pushFinal(ReaderState* state) {
+    if (state == nullptr) {
+        return;
+    }
+
+    bool expected = false;
+    if (!state->finalPublished.compare_exchange_strong(expected, true, std::memory_order_acq_rel)) {
+#ifndef NDEBUG
+        assert(false && "pushFinal called more than once on the same ReadState");
```
RalphSteinhagen: In release this silently fails ... -> indicate harder error since this is an API/design/usage error: either exception or std::abort

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
```diff
@@ -0,0 +1,1117 @@
...
+            res.data = std::unexpected(std::move(msg.data.error()));
+            std::invoke(std::forward<TCallback>(callback), std::move(res));
+            std::ignore = span.consume(1);
+            return;
+        }
+
+        auto it = msg.data->find(detail::kMessageDataKey);
+        if (it != msg.data->end()) {
+            if (auto bytes = std::get_if<std::vector<std::uint8_t>>(&it->second); bytes != nullptr) {
+                if (bytes->size() <= maxSize) {
```
RalphSteinhagen, line 226: The user may specify (by accident) a maxSize that is systematically too small. This will indefinitely spin this function on the same message.

algorithm/CMakeLists.txt
drslebedev: Updated:
CMake controls for native HTTP (CPR + libcurl):
- GR_ENABLE_NATIVE_HTTP=ON: Require libcurl. If libcurl is found, CPR is enabled and GR_NATIVE_HTTP_ENABLED=1; otherwise CMake fails with an error.
- GR_ENABLE_NATIVE_HTTP=OPTIONAL: Enable CPR only if system libcurl is found. If libcurl is missing, GR_NATIVE_HTTP_ENABLED=0 and HTTP(S) operations are treated as "disabled at build time".
- GR_ENABLE_NATIVE_HTTP = OFF: CPR is never fetched; GR_NATIVE_HTTP_ENABLED=0 and all HTTP(S) operations return "disabled at build time".

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
drslebedev: Renamed to `GR_HTTP_ENABLED`.

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
drslebedev: It is not hard error, so I changed it to `std::puts`.

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
drslebedev: We can’t really use `static_assert` here because it depends on the runtime user-provided `uri` value. A `static_assert` would either always fire when HTTP is disabled or never fire.

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
drslebedev: `maxSize` was only introduced to make `processBulk` easier to implement: the caller should pass the number of free slots in the output buffer, so after `poll` they can write all data directly without intermediate storage. However, the user must also ensure `chunkSize <= output buffer size`, otherwise chunks might not fit, which is a bit error-prone. But `maxSize` itself does not bring to the infinite poll. An infinite loop can only occur in the user code (`while(){pool()}`) if the caller fails to handle errors returned by poll, which is a logical bug in the user code. I’ve now set all default sizes to std::numeric_limits<std::size_t>::max() and added documentation for both `chunkSize` and `maxSize`.

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
drslebedev: I’d keep this as a returned `std::expected` error. Calling `write()` on the main WASM thread is an API misuse, and we signal that clearly by returning an `Error` instead of data.

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIoHelpers.hpp
drslebedev: I think using just ":" is still safe because we also check std::isalpha, but I went ahead and switched it to ":/" as suggested (only one "/" because of "file:/"). I also added case-insensitive handling for the scheme. Right now, addresses without a scheme aren’t supported – do we actually want to accept bare www.example.com as input? It ca be easily added if required.

PR #692, opened 2025-12-31
Introduce Value, Value::Map and Tensor everywhere

ivan-cukic:
- Replace vectors with Tensor
- Port everything to Value and Value::Map from pmtv and old property_map
- Fixes iterators in Tensors
- Fixes Values on Emscripten not handling std::size_t

### DIFF
--- a/blocks/basic/test/qa_TriggerBlocks.cpp
+++ b/blocks/basic/test/qa_TriggerBlocks.cpp
@@ -25,7 +25,7 @@ const suite<"SchmittTrigger Block"> triggerTests = [] {
     }
 
     using enum gr::trigger::InterpolationMethod;
-    "SchmittTrigger"_test =
+    skip / "SchmittTrigger"_test =
         [&enableVisualTests]<class Method> {
             Graph graph;
 


blocks/math/include/gnuradio-4.0/math/Rotator.hpp
```diff
@@ -14,7 +14,7 @@ namespace gr::blocks::math {
 
 GR_REGISTER_BLOCK(gr::blocks::math::Rotator, [T], [ std::complex<float>, std::complex<double> ])
 
-template<pmtv::Complex T>
+template<typename T>
```
RalphSteinhagen: minor: `gr::meta::complex_like`

core/include/gnuradio-4.0/PmtCollections.hpp
```diff
@@ -201,7 +206,7 @@ struct vector<T, true> { // managed vector
                         std::uninitialized_move(first, last, _data);
 #else
                         if constexpr (detail::iter_yields_nonconst_T_ref<T, decltype(first)> && !std::is_integral_v<T>) {
-                            std::uninitialized_move(first, last, _data); // only use move for non-integral types where it matters
```
RalphSteinhagen, line 204: Is this still an issue?

core/include/gnuradio-4.0/ValueHelper.hpp
```diff
@@ -1059,6 +1059,98 @@ extern template std::expected<std::map<std::string, Value>, ConversionError>
     convertTo<std::map<std::string, Value>>(const Value&, std::pmr::memory_resource*);
 // clang-format on
 
+struct ValueVisitor {
```
RalphSteinhagen, line 1062: Am still wondering whether we really need this visitor pattern (expensive w.r.t. compile-time overhead, rarely used, constexpr-if often suffices),

core/include/gnuradio-4.0/PmtCollections.hpp
ivan-cukic: Yes, still a problem with libc++

PR #761, opened 2026-04-06
math: populate UncertainValue for MathOpMultiPortImpl blocks

mormj:
These were removed from the previous plugin instantiations for MathOpMultiPortImpl, so they are put back and the require statement had to be updated.


RalphSteinhagen (approved):
q-fix. OK.

PR #762, opened 2026-04-06
[For Discussion] Deploy plugins for Simple Schedulers

mormj:
The idea is to create the .so file that has the default Simple schedulers so that higher layer tooling can discover. 

Just switching between SimpleSingle and SimpleMulti gives a good idea for the MVP UI what is possible.

I'm not sure if this is the right place to put the code.


RalphSteinhagen:
I feel a bit hesitant to put the 'Simple', 'BreadthFirst', 'DepthFirst', and other schedulers into the official plugin registry. 

These are primarily intended to test the required scheduler API surface and what's possible to implement with it. 

While they work and demonstrate certain GR4 capabilities, they are **not** vetted high-performance, production-grade implementations, aside from testing the API surface. There is (or should be) a lot more graph theory involved for general-purpose use. 

The existing API-testing scheduler should be supplanted by the community's best general-purpose candidate (maybe through a competition) and- by design- can also be added/addressed in OOT specialisations where users can make their own choices w.r.t. scheduling priorities (e.g. throughput vs. latency, graph traversal, topology choices, ... *<long list>*).

RalphSteinhagen:
@mormj, you raised an important point and should IMO be discussed in the architecture channel.

PR #764, opened 2026-04-08
Added sub-graph assets support

ivan-cukic:
- Introspection messages now return YAML definitions if the "serialization_format" message property is "yaml"
- Block instantiation message accepts YAML-definitions of blocks
- The plugin loader mechanism now recognizes yaml-based sub-graph definitions
- The definitions are cached to the local storage


RalphSteinhagen (approved):
Thanks for finishing the missing sub-graph registry feature [#764 ](https://github.com/fair-acc/opendigitizer/issues/428)

Added some minor additions/fixes.

PR #768, opened 2026-04-15
Add namespaced alias to exported targets

dantti:
Namespaced targets allow for better CMake checking of target dependencies, allowing it to fail at
configure time rather than silently or at link time.

Use namespaced aliases to link in tree.

### DIFF
--- a/.github/workflows/cpack.yml
+++ b/.github/workflows/cpack.yml
@@ -15,12 +15,12 @@ jobs:
       CCACHE_DIR: ${{ github.workspace }}/.ccache
       CCACHE_BASEDIR: ${{ github.workspace }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           fetch-depth: 0
 
       - name: Cache ccache
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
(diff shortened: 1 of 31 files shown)


RalphSteinhagen (approved):
@dantti thanks for improving the pluging/packaging! Much appreciated. 👍 

- Adds `gnuradio4::` ALIAS targets for every in-tree library — mistyped deps now fail at configure time, not link time.
- Exports shared/static block-library wrappers via `$<TARGET_OBJECTS:>`, installs previously uninstalled libs/headers (`gr-basic`, `gr-testing`, `exprtk.hpp`, `bench/benchmark.hpp`, …), adds `gnuradio4PluginTargets` to the package config, and fixes static libs to use `ARCHIVE DESTINATION`.
- Drops an inverted `gr-basic → gr-testing` self-link and relocates generated `GrXxxBlocks.hpp` to `build/include/gnuradio-4.0/` (canonical include prefix).

Heads-up: existing build trees need a clean reconfigure — `GEN_DIR` moved from `build/generated_plugins/` to `build/plugins/`.

Follow-ups: finish the in-tree namespacing sweep (a few `core/`, test, and block CMakeLists still use bare names); add a CI smoke job covering `find_package(gnuradio4)` and `FetchContent` consumers.

PR #769, opened 2026-04-16
refactor(core): prepare Block/Graph/Settings for device integration

### LINKED ISSUE #367: Handling of input data before the end-of-stream tag that cannot be processed due to port/block constraints
The current handling in this situation is in an incomplete and inconsistent state:
  - processes last chunk before EoS without considering constraints
  - but only port constraints and not resampling constraints
  - no way to customize that behavior

After some discussion with me, @RalphSteinhagen and @drslebedev the following implementation is proposed:
  - by default,  drop last samples(and tags except EoS of course) that violate the constraints (resampling, stride & port min/max)
  - provide an optional customisation point:  `Derrived::processBulkEpilogue(args same as process bulk modulo std::span vs ConsumableSpan)`
    - that way, the regular process function can be kept free of the branch for handling incomplete data
    - it will provide all available data on all ports before the end of stream tag as arguments and allows to publish or not to publish samples at the implementers discretion
    - need to add additional unit tests for the extension point

For implementation, the default behaviour is changed in #366 to what is described above, implementing the customisation point will be a separate task to be tackled after the ASYNC handling.

### LINKED ISSUE #439: Redesign of InputSpan and OutputSpan
### Description:

`InputSpan` and `OutputSpan` classes maintain internal states (e.g., `tagsPublished`) that are unique to each instance and not shared among different instances. Copying these objects without proper state management can lead to incorrect behavior when methods like `publishTag()` are called.

For example, in the `processBulk` function, when `InputSpan` or `OutputSpan` objects are passed by value, their internal states are copied. This can result in inconsistent or unintended behavior due to the duplication of state information. While `processBulk` illustrates the problem, the underlying issue is broader: we need to avoid scenarios where internal states are copied for each instance.

### Questions:
- **Should we move all internal state members to `CircularBuffer::Reader` and `CircularBuffer::Writer`?**
  - Centralizing the internal states within the `Reader` and `Writer` makes all states shareable between instances.  

- **Should we prevent copying of `InputSpan` and `OutputSpan` to enforce proper state management?** AND **Should we disallow obtaining spans several times without ensuring that `consume()` or `publish()` methods are called?**
  - By deleting or disabling copy constructors and copy assignment operators, we can ensure that instances cannot be copied. +  there are no 2 instances which point to the same span of data, could ensure that state changes occur in a controlled manner.

**Note:** This issue highlights the need for careful handling of internal states within `InputSpan` and `OutputSpan` to prevent incorrect behavior due to improper copying or state sharing. Feedback and suggestions on the proposed questions and action items are welcome.

### LINKED ISSUE #625: Tag API Refactor: Support Multiple Tags per Sample & Deprecate mergedTag
### Current Issue

The current implementation merges multiple tags into a single `mergedTag`, which can then be retrieved via `Block<T>::mergedInputTag()`. Tags are commonly merged when they share the same index, but the Port API also supports merging tags up to a specified index using `in.getMergedTag(untilLocalIndex)`.

Both tag forwarding and settings auto update rely on this merged tag API.

If multiple tags contain identical keys, only one tag’s data is preserved during merging, and the API does not specify clearly which tag “wins.” Although this behavior generally works, there are scenarios where access to the complete set of tags is required without implicit data loss.

### Proposed Change

Remove the merged-tag API in favor of providing explicit access to the full, unmerged set of tags. The API could include methods similar to those available in `Port<T>`:

```cpp
bool Block<T>::inputTagsPresent();
std::span<std::pair<std::ptrdiff_t, property_map>> Block<T>::tags();
```

### Impact and Tasks

The current mergedTag implementation is utilized in several block implementations and associated unit tests. These will need to be revised as part of this proposed change.

RalphSteinhagen:
**Internal refactoring only — no user-facing API changes.** 
Existing blocks, graphs, and schedulers continue to work without modification.

This PR prepares `Block.hpp`, `Graph.hpp`, and `Settings.hpp` for non-CPU device (e.g. GPUs) integration, (firm) real-time , and embedded fixed-memory systems by:

- replacing the monolithic 'merged-tag' API with per-tag access and composable forwarding policies (#625)
- extracting composable methods from `workInternal()` for custom dispatch
- wiring `ComputeDomain` resolution into edge buffer allocation (PMR)
- adding `processEpilogue()` for trailing EOS samples (#367)
- making `InputSpan`/`OutputSpan` move-only (#439)
- supporting `std::pmr::string` and `std::pmr::vector<T>` in user block settings
- fixing `AtomicRef.hpp` ACPP guard and Meyer's singletons

### Tag refactoring (#625) -- *see new MD documentation page for details*

The mandatory merged-tag API is augmented with per-tag access. `processBulk` blocks read tags via `input.tags()`; `processOne` blocks use `inputTagsPresent()` + `mergedInputTag()` (as before). 
No implicit data loss — all tag key-values are forwarded individually.

*Example (from the doc)*:

```
  input:    ▽           ▼              ▽
            │           │              │
  samples: ─┬──┬──┬──┬──┬──┬──┬──┬──┬──┬──┬─
            0  1  2  3  4  5  6  7  8  9  10

  chunk 1:  ├───────────┤                       (▽ at 0, next tag at 4)
  chunk 2:              ├──────────────┤        (▼ at 4, next tag at 9)
  chunk 3:                             ├─────┤  (▽ at 9, end of data)

  output:   ▽           ▼              ▽        tags forwarded at chunk start
```

Five composable tag propagation policies explicit:
- **Default**: chunks break at tag boundaries, tag at position 0 forwarded
- **ForwardTagPropagation**: fixed-chunk blocks, unconsumed tags carry forward
- **BackwardTagPropagation**: all chunk tags mapped to output position 0
- **MergeTagPropagation**: legacy — merge all tags into one output tag
- **NoTagPropagation**: GR4 framework does not forward

User-overridable `forwardTags()` detected via `requires` for full custom control.

**Performance** (Release, GCC 15, best-of-3 vs baseline):

| Scenario | Before | After | Delta |
|---|---|---|---|
| No tags | 469 MS/s | 500 MS/s | +7% |
| Moderate (1 tag / 10k samples) | 376 MS/s | 385 MS/s | +2% |
| Tag every sample | 475 kS/s | 544 kS/s | +15% |

### processEpilogue (#367)

Opt-in customisation point for blocks that accumulate state (FFT, file writers). When EOS arrives, and the remaining samples cannot form a complete chunk, blocks with `processEpilogue()` receive the trailing samples. 
Detected via `if constexpr` and thus zero cost when unused.

### PMR / device integration

`ComputeDomain::parse("gpu:sycl:0")` parses domain strings (backend strings passed through for SYCL-reported device names, N.B. similar when using CUDA or ROCm). `applyEdgeConnection()` resolves edge domain to PMR resources automatically:

```
explicit EdgeParameters.domain  >  block compute_domain  >  host (default)
```

`std::pmr::string` and `std::pmr::vector<T>` work as block settings fields
(opt-in). On embedded systems, `std::pmr::set_default_resource(&pool)` redirects
all PMR containers to a fixed memory pool.

### Pre-existing Settings bugs fixed

Exposed by removing persistent tag state from `Block<T>`:
- `autoUpdate()` wiped staged params on non-matching tag
- `activateContext()` overwrote auto-update params with stored defaults
- early return at `resampledIn==0` lost pending forward params

### Other fixes
- `InputSpan` / `OutputSpan` made move-only — copying silently broke instanceCount-based refcounting (#439)
- `AtomicRef.hpp`: `sycl.hpp` gated behind `__ACPP__`, not `__has_include`
- Meyer's singleton cleanup for CI stability

## Deferred

- Internal string migration (`Block::name`, `unique_name`, `compute_domain`)
  to `std::pmr::string` — blocked by `std::string` / `std::pmr::string` type
  incompatibility in C++ (no cross-type comparison or map lookup). Needs
  `gr::to_pmr()` / `gr::to_std()` helpers (added) and ~200 site changes (a bit too much for this PR).


core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1069,102 +1087,157 @@ class Block : public lifecycle::StateMachine<Derived>, public BlockBase {
...
+                        return;
+                    }
+                    for (const auto& [relIndex, tagMapRef] : in.tags(tagWindow)) {
+                        publishFiltered(relIndex, tagMapRef.get());
+                    }
+                },
+                inputSpans);
+        } else {
+            using SeenEntry                                      = std::pair<std::ptrdiff_t, const property_map*>;
+            constexpr std::size_t                 kDedupCapacity = 8UZ;
```
drslebedev, line 1163: The multi-port dedup logic only tracks 8 seen tags. After that, duplicate tags may no longer be filtered and can be forwarded again. Should we at least emit a warning when this limit is exceeded?

drslebedev (approved):
Thanks for the PR.

Overall, it looks good to me and I think it can be merged as is. We already discussed some questions in private communication, and those are all resolved. The additional unit tests and documentation also look good and are very helpful.

Just one note for follow-up: some blocks will still need multi-tag support, for example DataSink. But I think this should be handled separately in another PR.

core/include/gnuradio-4.0/Block.hpp
RalphSteinhagen: No. This is in the hot-path, and the dedup capacity is a performance optimisation and could also be implemented by another space-vs-time trade-off. There are no tags dropped in any case.
I chose this solution since the case you are hinting at (i.e. non-deduplication) is *extremely* rare.

Issue #657, opened 2025-09-13
Remove all FetchContent from CMake


marcusmueller:
Can't do FetchContent in any distro builds, nor in openembedded, Conan/artifactory builds etc. Simply a show-stopper for distribution, and thus for a 4.0 release.

We need all externalities to either be packaged or package them ourselves. Thus, external sources without proper release versions get a 🔴 , and us not using an explicit release but either just "get HEAD" or a git commit hash gets a 🔴 .

| name | URL | external: release hygiene? | using release? | available in debian? | conan? |remark |
|---------|----|----|----|-----|---|--|
| pffft | https://github.com/fair-acc/pffft.git | 🔴 (upstream: also 🔴 ) | 🔴  | 🔴 | 🔴 (upstream: 🟡 ) | can we upstream changes and invite upstream to release? |
| boost-external ut | https://github.com/boost-ext/ut.git | 🟢 | 🔴  | 🔴  (but Fedora) | ? | why are we using a random git commit? |
| vir-smd | https://github.com/mattkretz/vir-simd.git | 🟢  | 🟢  | 🔴 | 🔴 | @mattkretz |
| cpp-httplib | https://github.com/yhirose/cpp-httplib.git | 🟢 | 🟢 | 🟢 | 🟢 | should immediately be unvendored |
| pmt | https://github.com/gnuradio/pmt.git | 🔴 | 🔴  | 🔴 | 🔴 | @jsallay we should sit down together and make this a proper lib with proper releases and packages, and a bit of publicity for GR4. @maitbot are you interested in helping us get this into debian? |

mattkretz:
On pffft … we could try. But I suspect our changes are going too far. (Upstream is C and we turned it into C++ and replaced their SIMD "abstraction".) If that's the case then we would need to fork it and call it something else, no?

On vir-simd … is there anything I could/should do?

RalphSteinhagen:
@mattkretz nothing on your side. We will proceed as we discussed earlier at work.

RalphSteinhagen:
* **pffft**: we have a `<simd>` version (WIP, needs clean-up) for it that we will incorporate to the core (see also [here](https://github.com/fair-acc/gnuradio4/issues/656).
* **boost-external ut**: this is only needed for the CI (unit-tests). *N.B. has nothing realy to do with 'boost' but in naming.*
[...]

PR #787, opened 2026-05-05
Latest tag

dantti:
This creates a latests tag that will instead of daily will build once merged to main, and besides gcc and emscripten packages will also create clang packages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,14 +17,14 @@ jobs:
       packages: write
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         if: github.event_name == 'push'
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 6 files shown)


RalphSteinhagen (approved):
@dantti thanks for the update. This should work for our rolling-release caching of package artefacts ... should also check with Maitlin how much this could be used to create and derive Debian packages from it.

dantti:
@RalphSteinhagen the namespaced targets are the most important part, but to be distro friendly we should likely address https://github.com/fair-acc/gnuradio4/issues/657

RalphSteinhagen:
> @RalphSteinhagen the namespaced targets are the most important part, but to be distro friendly we should likely address #657

Most of the remaining dependencies mentioned there are obsolete, already replaced, and/or unit-test- or build-time-related. I'd refer this to later.

PR #788, opened 2026-05-08
refactor(build): centralise dep resolution; harden install/cross-compile

### LINKED ISSUE #657: Remove all FetchContent from CMake

RalphSteinhagen:
Centralise GR4's external-dependency acquisition behind `gr4_declare_dependency` / `gr4_resolve_dependency`, 
fix the out-of-tree install/consume contract end-to-end, and stop the parser-tool sub-build from inheriting cross-compile toolchains.

**Dependency policy (the core refactor)**
- `cmake/Dependencies.cmake`: `gr4_declare_dependency` / `gr4_resolve_dependency` with per-dep `system | fetch | system-or-fetch`  (default), per-dep override `-DGR4_DEP_<NAME>_MODE=…`.
- `DEPENDENCIES.md` is the single BOM. `cmake/lint_dependencies.cmake` runs as an `ALL` target and FATAL_ERRORs on drift (declared deps without a doc row, pinned `FETCH_TAG`s not mentioned in the markdown).
- `cmake/FindLibSoundIo.cmake` covers the libsoundio system gap (no upstream `*Config.cmake`).

**Out-of-tree install/consume**
- `gnuradio4Config.cmake.in` emits `find_dependency()` only for transitive deps actually consumed (vir-simd / TBB / cpr / SoapySDR / libsoundio).
  Build-time fetch policy does not leak to consumers.
- `FETCH_INCLUDE_IN_ALL` per-dep opt-in: when set, the fetched dep stays in `ALL` so its own install rules run. cpr opts in so `cprConfig.cmake` reaches the install prefix; otherwise downstream `find_dependency(cpr CONFIG)` would fail. Default remains `EXCLUDE_FROM_ALL`.
- `GnuRadioBlockLibConfig.cmake.in` drops a broken `include(GnuRadioBlockLibTargets.cmake)` (no such file was ever installed), exposes the parser as `gnuradio4::parse_registrations` IMPORTED EXECUTABLE, and uses  `configure_package_config_file(... PATH_VARS ...)` so absolute `CMAKE_INSTALL_BINDIR` is path-safe.
- `gr_add_block_plugin` gains `NO_INSTALL` (used by `core/test/plugins/GoodBase`) so test artefacts stop polluting `cmake --install`.
- New CTest entry `install_consumer_smoke` (label `install-smoke`) does install + clean `find_package(gnuradio4)` round-trip via a tiny external consumer; gated on `ENABLE_TESTING && GR_TOPLEVEL_PROJECT && NOT EMSCRIPTEN`.

**Cross-compile**
- `blocklib_generator/CMakeLists.txt`: parser-tool sub-build is `CMAKE_CROSSCOMPILING`-aware. Native builds keep using `CMAKE_CXX_COMPILER`; cross builds let the sub-build's CMake pick a host compiler from `PATH`. Inherited `CMAKE_TOOLCHAIN_FILE`, `CMAKE_SYSROOT`, compiler-target args and `CXX/CC/CXXFLAGS/CFLAGS/LDFLAGS/CPPFLAGS` env are stripped before the sub-build invocation. Override the host compiler with `-DGNURADIO_PARSE_REGISTRATIONS_TOOL_CXX_COMPILER=…`. Misspelled legacy `…_COMPLILER` is still honoured with a `DEPRECATION` warning.


drslebedev (approved):
Thanks for the PR.

The changes look good to me. I fixed two minor issues with a couple of additional commits.

I also noticed that cmake-format sometimes produces formatting that is a bit awkward and hard to read. I’ve opened a separate issue to address this: https://github.com/fair-acc/gnuradio4/issues/789

The PR is ready to be merged.

PR #794, opened 2026-05-28
Add static display_name to blocks, and displayName() to registry

the-argus:
Enables blocks to define `using DisplayName = Doc<...>` which can then be read by UIs for a not-necessarily-unique and human readable name.

Will be used to address [opendigitizer/449](https://github.com/fair-acc/opendigitizer/issues/449)

Also includes a small fix for schedulers to use SchedulerBase constructors which is required so they can be registered, that helps [opendigitizer/448](https://github.com/fair-acc/opendigitizer/issues/448)

Potentially breaking change in that it adds a new requirement to the block concept.

### DIFF
--- a/core/include/gnuradio-4.0/Scheduler.hpp
+++ b/core/include/gnuradio-4.0/Scheduler.hpp
@@ -1346,9 +1346,31 @@ protected:
     }
 };
 
+namespace detail {
+template<ExecutionPolicy policy>
+consteval auto executionPolicyName() {
+    if constexpr (policy == ExecutionPolicy::singleThreaded) {
+        return gr::meta::fixed_string("singleThreaded");
+    } else if constexpr (policy == ExecutionPolicy::multiThreaded) {
+        return gr::meta::fixed_string("multiThreaded");
+    } else if constexpr (policy == ExecutionPolicy::singleThreadedBlocking) {
+        return gr::meta::fixed_string("singleThreadedBlocking");
+    } else {
+        static_assert(false, "Unsupported ExecutionPolicy for scheduler display name");
+    }
+}
+
+template<gr::meta::fixed_string SchedulerName, ExecutionPolicy policy>
+consteval auto schedulerDisplayName() {
+    return SchedulerName + "<" + executionPolicyName<policy>() + ">";
+}
+} // namespace detail
+
 template<ExecutionPolicy execution = ExecutionPolicy::singleThreaded, profiling::ProfilerLike TProfiler = profiling::null::Profiler>
 struct Simple : SchedulerBase<Simple<execution, TProfiler>, execution, TProfiler> {
     using Description = Doc<R""(Simple loop based Scheduler, which iterates over all blocks in the order they have beein defined and emplaced definition in the graph.)"">;
+
+    using DisplayName = Doc<detail::schedulerDisplayName<"Simple", execution>()>();
 
     using SchedulerBase<Simple<execution, TProfiler>, execution, TProfiler>::SchedulerBase;
 
@@ -1412,6 +1434,8 @@ template<ExecutionPolicy execution = ExecutionPolicy::singleThreaded, profiling:
 struct BreadthFirst : SchedulerBase<BreadthFirst<execution, TProfiler>, execution, TProfiler> {
     using Description = Doc<R""(Breadth First Scheduler which traverses the graph starting from the source blocks in a breath first fashion
 detecting cycles and blocks which can be reached from several source blocks.)"">;
+
+    using DisplayName = Doc<detail::schedulerDisplayName<"BreadthFirst", execution>()>();
 
     static_assert(execution == ExecutionPolicy::singleThreaded || execution == ExecutionPolicy::multiThreaded, "Unsupported execution policy");
 
@@ -1484,6 +1508,9 @@ detecting cycles and blocks which can be reached from several source blocks.)"">
 template<ExecutionPolicy execution = ExecutionPolicy::singleThreaded, profiling::ProfilerLike TProfiler = profiling::null::Profiler>
 struct DepthFirst : SchedulerBase<DepthFirst<execution, TProfiler>, execution, TProfiler> {
     using Description = Doc<R""(Depth First Scheduler which traverses the graph starting from the source blocks in a depth-first manner.)"">;
+
+    using DisplayName = Doc<detail::schedulerDisplayName<"DepthFirst", execution>()>();
+
     static_assert(execution == ExecutionPolicy::singleThreaded || execution == ExecutionPolicy::multiThreaded, "Unsupported execution policy");
 
     void customInit() {
--- a/core/include/gnuradio-4.0/Scheduler.hpp
+++ b/core/include/gnuradio-4.0/Scheduler.hpp
@@ -1439,6 +1439,8 @@ detecting cycles and blocks which can be reached from several source blocks.)"">
 
     static_assert(execution == ExecutionPolicy::singleThreaded || execution == ExecutionPolicy::multiThreaded, "Unsupported execution policy");
 
+    using SchedulerBase<BreadthFirst<execution, TProfiler>, execution, TProfiler>::SchedulerBase;
+
     void customInit() {
         /* implements Breadth-first search scheduling algorithm (https://en.wikipedia.org/wiki/Breadth-first_search)
          * 1. compute 'adjacencyList'
@@ -1512,6 +1514,8 @@ struct DepthFirst : SchedulerBase<DepthFirst<execution, TProfiler>, execution, T
     using DisplayName = Doc<detail::schedulerDisplayName<"DepthFirst", execution>()>();
 
     static_assert(execution == ExecutionPolicy::singleThreaded || execution == ExecutionPolicy::multiThreaded, "Unsupported execution policy");
+
+    using SchedulerBase<DepthFirst<execution, TProfiler>, execution, TProfiler>::SchedulerBase;
 
     void customInit() {
         /**
--- a/core/include/gnuradio-4.0/BlockRegistry.hpp
+++ b/core/include/gnuradio-4.0/BlockRegistry.hpp
@@ -52,6 +52,7 @@ class GeneralRegistry {
 
     struct TTypeHandler {
         std::string                     alias;
+        std::string                     displayName;
         decltype(this_t::factoryProto)* createFunction = nullptr;
     };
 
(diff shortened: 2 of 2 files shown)


drslebedev (commented):
Thanks for the PR!

  I think the DisplayName / display_name bridge itself is reasonable, since it follows the existing Description / description pattern. One possible alternative would be to avoid adding display_name to Block and keep the extraction logic inside the registry instead, e.g. by checking for TBlock::DisplayName during registration.

  One question: do we also need to extend PluginLoader / the plugin API for this? A plugin can store the display name in its own registry, but from what I see PluginLoader can currently retrieve only block/scheduler names through availableBlocks() / availableSchedulers(), not the associated display metadata.

  Maybe this should eventually become a richer type-info API that returns name, alias, and displayName together, rather than exposing only raw names?

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -1350,6 +1350,17 @@ template<ExecutionPolicy execution = ExecutionPolicy::singleThreaded, profiling:
 struct Simple : SchedulerBase<Simple<execution, TProfiler>, execution, TProfiler> {
     using Description = Doc<R""(Simple loop based Scheduler, which iterates over all blocks in the order they have beein defined and emplaced definition in the graph.)"">;
 
+    constexpr static auto getDisplayName() {
```
drslebedev: Just an optional suggestion to consider. Since this is now used in three scheduler types, one alternative would be to centralize the execution-policy formatting in a shared helper instead of repeating getDisplayName() in each scheduler.

  ```cpp
  template<ExecutionPolicy policy>
  consteval auto executionPolicyName() {
      if constexpr (policy == ExecutionPolicy::singleThreaded) {
          return gr::meta::fixed_string("singleThreaded");
      } else if constexpr (policy == ExecutionPolicy::multiThreaded) {
          return gr::meta::fixed_string("multiThreaded");
      } else if constexpr (policy == ExecutionPolicy::singleThreadedBlocking) {
          return gr::meta::fixed_string("singleThreadedBlocking");
      } else {
          static_assert(false, "Unsupported ExecutionPolicy for scheduler display name");
      }
  }

  template<gr::meta::fixed_string SchedulerName, ExecutionPolicy policy>
  consteval auto schedulerDisplayName() {
      return SchedulerName + "<" + executionPolicyName<policy>() + ">";
  }

  Then each scheduler only needs:

  using DisplayName = Doc<schedulerDisplayName<"Simple", execution>()>;
  using DisplayName = Doc<schedulerDisplayName<"BreadthFirst", execution>()>;
  using DisplayName = Doc<schedulerDisplayName<"DepthFirst", execution>()>;

drslebedev (approved):

core/include/gnuradio-4.0/Scheduler.hpp
the-argus: Oh yeah that fixed string operator + is nifty, I will change that

the-argus:
> One possible alternative would be to avoid adding display_name to Block and keep the extraction logic inside the registry instead, e.g. by checking for TBlock::DisplayName during registration.

Ah okay good point, I just put it there for convenience but it would be nice to not have to modify the concept just for that.

> One question: do we also need to extend PluginLoader / the plugin API for this? A plugin can store the display name in its own registry, but from what I see PluginLoader can currently retrieve only block/scheduler names through availableBlocks() / availableSchedulers(), not the associated display metadata.

> Maybe this should eventually become a richer type-info API that returns name, alias, and displayName together, rather than exposing only raw names?

@drslebedev It should be fine because the loader can just ask the registry for the richer info via `displayName()` and `typeName()`?

I thought it would be nicer to merge those two into one function that returns a type info struct of some kind, but I didn't want to make a potentially breaking change for people. Is that related to your suggestion?

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -1346,10 +1346,32 @@ struct SchedulerBase : Block<Derived> {
...
+consteval auto schedulerDisplayName() {
+    return SchedulerName + "<" + executionPolicyName<policy>() + ">";
+}
+} // namespace detail
+
 template<ExecutionPolicy execution = ExecutionPolicy::singleThreaded, profiling::ProfilerLike TProfiler = profiling::null::Profiler>
 struct Simple : SchedulerBase<Simple<execution, TProfiler>, execution, TProfiler> {
     using Description = Doc<R""(Simple loop based Scheduler, which iterates over all blocks in the order they have beein defined and emplaced definition in the graph.)"">;
 
+    using DisplayName = Doc<detail::schedulerDisplayName<"Simple", execution>()>();
```
drslebedev, line 1373:
```suggestion
    using DisplayName = Doc<detail::schedulerDisplayName<"Simple", execution>()>;
```

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -1413,8 +1435,12 @@ struct BreadthFirst : SchedulerBase<BreadthFirst<execution, TProfiler>, executio
     using Description = Doc<R""(Breadth First Scheduler which traverses the graph starting from the source blocks in a breath first fashion
 detecting cycles and blocks which can be reached from several source blocks.)"">;
 
+    using DisplayName = Doc<detail::schedulerDisplayName<"BreadthFirst", execution>()>();
```
drslebedev, line 1438:
```suggestion
    using DisplayName = Doc<detail::schedulerDisplayName<"BreadthFirst", execution>()>;
```

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -1484,8 +1510,13 @@ detecting cycles and blocks which can be reached from several source blocks.)"">
 template<ExecutionPolicy execution = ExecutionPolicy::singleThreaded, profiling::ProfilerLike TProfiler = profiling::null::Profiler>
 struct DepthFirst : SchedulerBase<DepthFirst<execution, TProfiler>, execution, TProfiler> {
     using Description = Doc<R""(Depth First Scheduler which traverses the graph starting from the source blocks in a depth-first manner.)"">;
+
+    using DisplayName = Doc<detail::schedulerDisplayName<"DepthFirst", execution>()>();
```
drslebedev, line 1514:
```suggestion
    using DisplayName = Doc<detail::schedulerDisplayName<"DepthFirst", execution>()>;
```

drslebedev:
> @drslebedev It should be fine because the loader can just ask the registry for the richer info via `displayName()` and `typeName()`?
> 
> I thought it would be nicer to merge those two into one function that returns a type info struct of some kind, but I didn't want to make a potentially breaking change for people. Is that related to your suggestion?

The only caveat I see is the plugin boundary. The PluginLoader can ask its own local registry for `displayName()` / `typeName()`, but plugin-provided entries are owned by the plugin's internal registry. `PluginLoader` currently sees plugins only through `gr_plugin_base`, which exposes `availableBlocks()` /
  `availableSchedulers()` as names, but not registry metadata.

PluginLoader stores plugin as: gr_plugin_base*

  So it can only call members that exist on gr_plugin_base:
```cpp
  availableBlocks()
  createBlock()
  availableSchedulers()
  createScheduler()
```

drslebedev:
@the-argus I looked through this again and wanted to clarify the motivation for adding `displayName`.

  The registry already has the concept of an alias, which can be set via `registry.insert(...)`. As I understand it, that alias is already a shorter, human-readable name that can be returned from the registry and used for lookup.

  So I am wondering what extra semantics `displayName` is meant to provide beyond the existing `alias`. Is the intention that `displayName` is purely UI metadata, while alias remains a stable registry/instantiation key? 

  Sorry if I missed something. I just want to clarify why a separate display name is needed instead of reusing aliases for the user-facing name.

the-argus:
> PluginLoader stores plugin as: gr_plugin_base*
> So it can only call members that exist on gr_plugin_base

Ah I see, is there a reason why we can't/don't have something like `virtual gr::BlockRegistry& getBlockRegistry()` (and the same for schedulers) in `gr_plugin_base`? It would be trivial for `gr::plugin` to implement that, at least.

Otherwise maybe some new type `gr_block_metadata` and a
`gr_block_metadata gr_plugin_base::getBlockMetadata(std::string_view typeName) =0;`

> So I am wondering what extra semantics displayName is meant to provide beyond the existing alias. Is the intention that displayName is purely UI metadata, while alias remains a stable registry/instantiation key?

@drslebedev Yes pretty much. It seems like a good idea to me to allow arbitrary names and specify them with the type instead of with the `.insert()` call. Especially since alias has some implications like going through `makePortableTypeName` and needing to be unique. And I guess it also seemed useful for me for alias to mean "describes the c++ type, maybe in a shorter way than full typename" and display name to mean something closer to "explanatory title". I think it just happened that I went with this Typename\<Parameters\> style for the names anyways. Which is funny because my initial goal was to get rid of the C++-ness of the names, I had names like "Single-threaded Simple Scheduler" but those were a bit too long. 

Middle ground could be to just have the `DisplayName` declaration determine the default value for the `alias` parameter to `insert()` if none is provided. I think display names would typically be unique anyways.

Though these are all kind of minute requirements and I'm not opposed to just using `alias` for this purpose since that is simpler.

the-argus:
I went to experiment with just using alias, and remembered that we have

```cpp
    result.emplace(serialization_fields::BLOCK_ID, pluginLoader.registry().typeName(block));
```

for serializing blocks but

```cpp
        map.emplace(serialization_fields::BLOCK_ID, "SUBGRAPH");
```

for serializing subgraphs. This made me realize that for schedulers typename isnt accessible just via the graph inspection messages afaict, so in my tests I have been relying on `blockName` being the same as `blockTypeName`, which it is for now (I think that's *also* a bug as blockName should be unique). I think these will need to addressed also though it's not blocking here

drslebedev:
Thanks for the explanation!

I am not sure why there are no methods to access the registry. It seems like this should be possible without major issues. Maybe @ivan-cukic can comment on that. In any case, we can implement it as a follow-up.

My main concern about DisplayName is that it is currently only used by schedulers. If we implement it, I think we should add it to all blocks for consistency.

drslebedev:
Is this the place where we can add an alias to the scheduler? 
https://github.com/fair-acc/opendigitizer/blob/63313788d2cc0c2f7cf502608a1f8c29b1c0934d/src/ui/main.cpp#L257

the-argus:
> My main concern about DisplayName is that it is currently only used by schedulers. If we implement it, I think we should add it to all blocks for consistency.

Ah good point.

> Is this the place where we can add an alias to the scheduler?
https://github.com/fair-acc/opendigitizer/blob/63313788d2cc0c2f7cf502608a1f8c29b1c0934d/src/ui/main.cpp#L257

Yes, I can declare an alias there! Which would mostly work for my needs. Though I guess if GR started to register the schedulers itself that would break.

ivan-cukic:
> I am not sure why there are no methods to access the registry.
> It seems like this should be possible without major issues.
> Maybe @ivan-cukic can comment on that. In any case, we can implement it as a follow-up.

The API was kept minimal for easier ABI backwards compatibility. But, I guess, we could make it a bit more rich if needed.

the-argus:
Okay, I am just going to leave this PR for now and go try to implement the minimum stuff I need for opendigitizer/449. I do think putting the onus on the people writing the blocks to come up with a good display name is a good idea but for schedulers (especially these which are registered by the UI) it will be easy for the UI to handle coming up with display names.

the-argus:
Closing this, at least for now, we discussed in matrix about using uiConstraints for this, I think that's a good idea.

PR #796, opened 2026-06-05
Add ci-latest-wasm

dantti:
- Allows for easier CI - DEV builds with the same cmake flags
 - Builds in Release mode as debug info inflates the files and browsers have limits.

### DIFF
--- a/.github/workflows/latest.yml
+++ b/.github/workflows/latest.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Cache fetchContent
         uses: actions/cache@v5
         with:
-          path: ${{ github.workspace }}/build/_deps
+          path: ${{ github.workspace }}/build-ci-latest-gcc/_deps
           key: fetchContent-cpack-gcc14-relwithdebinfo-${{ hashFiles('CMakeLists.txt') }}
           restore-keys: |
             fetchContent-cpack-gcc14-relwithdebinfo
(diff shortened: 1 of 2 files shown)


.github/workflows/latest.yml
drslebedev: Should this path be updated to `build-ci-latest-wasm/*.deb`?

drslebedev: Same here: this should probably use `build-ci-latest-wasm/*.deb`?

drslebedev: Same here: this should probably use `build-ci-latest-wasm/_deps`

drslebedev: The preset now uses `CMAKE_BUILD_TYPE=Release`, but this artifact name still says `relwithdebinfo`.

CMakePresets.json
```diff
@@ -0,0 +1,45 @@
+{
+  "version": 6,
+  "cmakeMinimumRequired": {
+    "major": 3,
+    "minor": 23,
```
drslebedev: This should probably be at least 3.27, the project itself already requires CMake 3.27.

drslebedev (approved):
Thanks for the PR!

The changes look good to me overall. 

I left a few comments, mostly about paths/names that may need to be updated after moving the WASM build to the new preset. 

Please take a look before merging.

.github/workflows/latest.yml
DanielNicolettiKDAB: fixed

DanielNicolettiKDAB: fixed

DanielNicolettiKDAB: fixed

DanielNicolettiKDAB: fixed

CMakePresets.json
DanielNicolettiKDAB: fixed

DanielNicolettiKDAB:
@drslebedev I've fixed all issues you raised, and since you were not against the preset file I added the gcc/clang variant and simplified the latest.yml file with a preset matrix, deduplicating the steps

PR #799, opened 2026-06-09
Fix deadlock when scheduler is handling `kGraphGRC` message

### LINKED ISSUE #803: [2SP;2SP] Deadlock when setting a scheduler's graph via messages.
## [BUG] Deadlock when setting a scheduler graph via `kGraphGRC`

Sending a `Set` message to `scheduler::property::kGraphGRC` while the scheduler is running can deadlock.

This came up while working on fair-acc/opendigitizer#452: tests needed to load a graph with subgraphs via `Set kGraphGRC`, but the scheduler could hang while trying to exchange the graph. 

The current `propertyCallbackGraphGRC()` implementation loads the new GRC YAML, calls `makeAllZombies()`, and then immediately calls `exchange()`.

If this message is processed from a scheduler worker thread, `exchange()` observes an active scheduler state, requests stop, and calls `waitDone()`. `waitDone()` waits until `_nRunningJobs == 0`, but the worker currently executing the callback is itself still counted in `_nRunningJobs`. Since that worker is now blocked inside `waitDone()`, it never reaches the code path that decrements `_nRunningJobs`, so the scheduler waits forever.

PR #799 adds a regression test and tries deferring the exchange until `runAndWait()` regains control, but that is not a complete fix: `runAndWait()` is intended for unit tests. The real fix needs to work when the graph exchange is triggered from a scheduler worker thread in normal operation.

Part of this task should be to add or expand regression coverage for `Set` on `scheduler::property::kGraphGRC`, based on the test from #799, since this message path was previously untested.

One possible fix is to introduce a scheduler-level concept of zombified worker threads in addition to zombified blocks. A zombified worker should not block on itself when calling `waitDone()`, and it should have clear ownership/accounting rules so it does not incorrectly modify scheduler state after closing, for example by decrementing `_nRunningJobs` after it has handed off responsibility.

Other solutions are welcome if there is a simpler lifecycle-safe way to exchange the graph from a worker thread without self-deadlocking.

the-argus:
This PR adds reproducing tests and a fix for a deadlock in the scheduler, caused by setting a graphs GRC via messages. A worker thread picks it up and then tries to do `exchange()`, which calls `waitDone()`, which waits on itself.

My solution was just to have a worker thread do `waitDone()` for only the other threads, ie. waiting for `_nRunningJobs` to decrease to 1, and then stop the graph and exchange things. This works because only one thread processes scheduler messages.

The thread processing scheduler messages does not quit after the scheduler has been stopped and the graph replaced, it may continue to process messages. AFAICT this is harmless.

I noticed an issue where there was a potential data race on the `_graph` pointer from the watchdog thread. Then I also noticed that it was possible for watchdogs to miss that the graph had been stopped and restarted due to the AB problem with `_nRunningJobs`. (A similar issue still exists with `waitDone()`, where it is theoretically possible for `waitDone()` to be called while threads are still starting up and therefore `_nRunningJobs` goes to zero but other threads start afterwards).
I wanted to have some stronger guarantees about the watchdog threads, partially to fix these two issues and partially because I think it makes things easier to think about when I'm trying to think through all the ways that ordering can happen.  So I added this `WatchdogThreadContext` which enables the calling thread to do `_lastWatchdogThread.stop()` and be certain that the watchdog will not do any work after, but it does not block on a sleeping watchdog. LMK if there are some standard library mechanisms or something that can do this more concisely... it seems like something that should have an existing primitive, but I did not find one.

A problem here is that `propertyCallbackGraphGRC` might fail to do the `exchange()`, in which case it cannot recover from / undo the `makeAllZombies()` call. I think that is fine... I guess if that happens the user has bigger problems.

### DIFF
--- a/core/include/gnuradio-4.0/Scheduler.hpp
+++ b/core/include/gnuradio-4.0/Scheduler.hpp
@@ -134,6 +134,7 @@ protected:
     bool                          _valid{true};
     std::size_t                   _nWatchdogsRunning{0};
     meta::indirect<gr::Graph>     _graph{};
+    std::size_t                   _graphGeneration{0}; // incremented on exchange()
     TProfiler                     _profiler{};
     ProfileHandle                 _profilerHandler{_profiler.forThisThread()};
     std::shared_ptr<TaskExecutor> _pool{gr::thread_pool::Manager::instance().defaultCpuPool()};
@@ -291,6 +292,8 @@ public:
         if (this->state() == ERROR || this->state() == STOPPED) {
             reset(); // reset internal states
         }
+
+        gr::atomic_ref(_graphGeneration).fetch_add(1);
 
         auto oldGraph = std::exchange(_graph, std::move(newGraph));
 
@@ -498,9 +501,9 @@ public:
         return {};
     }
 
-    void waitDone() {
+    void waitDone(bool isCalledFromWorker = false) {
         [[maybe_unused]] const auto pe = _profilerHandler->startCompleteEvent("scheduler_base.waitDone");
-        while (_nRunningJobs->value() > 0UZ) {
+        while (_nRunningJobs->value() > (isCalledFromWorker ? 1UZ : 0UZ)) {
             std::this_thread::sleep_for(std::chrono::milliseconds(timeout_ms));
         }
         // _nRunningJobs == 0 is the only quiescence point shared by the runAndWait /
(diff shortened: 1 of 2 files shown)


the-argus:
~~marking as draft until #804 is merged~~ (it has been merged)

RalphSteinhagen (approved):
👍

RalphSteinhagen (approved):
Thanks for testing and fixing this.

PR #801, opened 2026-06-11
Serialize scheduler block parameters when inspecting graph

the-argus:
Adds block settings into the \"scheduler\" field. Opendigitizer can use this to display the `compute_domain` of a scheduler.

Also has a commit fixing missing constructors of `DepthFirst` and `BreadthFirst` which prevents them from being able to be registered.

Also adds `BLOCK_NAME` for regular blocks, for some reason previously that was only being serialized for subgraphs.

### DIFF
--- a/core/include/gnuradio-4.0/Scheduler.hpp
+++ b/core/include/gnuradio-4.0/Scheduler.hpp
@@ -1415,6 +1415,8 @@ detecting cycles and blocks which can be reached from several source blocks.)"">
 
     static_assert(execution == ExecutionPolicy::singleThreaded || execution == ExecutionPolicy::multiThreaded, "Unsupported execution policy");
 
+    using SchedulerBase<BreadthFirst<execution, TProfiler>, execution, TProfiler>::SchedulerBase;
+
     void customInit() {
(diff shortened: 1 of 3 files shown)


drslebedev (approved):
Thanks for the PR!

  The changes look good to me. 

I have only one point to consider: now that parameter serialization goes through the shared `serializeBlockSettings()` helper, scheduler serialization also includes ctx_parameters.

  However, when loading a managed SUBGRAPH (https://github.com/fair-acc/gnuradio4/blob/c272257c75d7d71b015cad3f0b54ca210816659e/core/include/gnuradio-4.0/Graph_yaml_importer.hpp#L144), the importer currently only reads scheduler.parameters, while the existing ctx_parameters import path is only used for normal blocks. That means scheduler context settings could be emitted in saved YAML and then silently dropped on reload. Should we also handle
  scheduler.ctx_parameters on import?

the-argus:
@drslebedev We don't restore `ctx_parameters` for regular blocks, and it seems like adding that would require expanding the plugin boundary since right now we just send parameters in the create function `std::unique_ptr<TModel> factoryProto(property_map params)`.

That seems like maybe a problem which should be solved in a separate PR?

PR #804, opened 2026-06-23
FEAT: enabling MCU, CPU, GPU, Real-Time, & Network cross-targets — ValueMap[View], PMRs, USM, zero-alloc, ...

### LINKED ISSUE #628: [10SP,25SP] SYCL PoC: Multi-FFT Kernel for Batched Spectral Processing
## SYCL PoC:<br>   Batched FFT Kernel Prototype for Multi-Channel Spectral Processing

This task tracks the development of an initial SYCL-based Proof of Concept (PoC) for GPU offloading in GNU Radio 4.0, starting with a batched FFT kernel suitable for channelisation and spectrogram-style applications.

### Scope
 * Prototype a SYCL kernel for multiple FFTs (batch mode), targeting mid-sized transforms (e.g. N >> 8192).
 * Integrate into a minimal GR4 `Block<T>` wrapper for standalone benchmarking and pipeline integration.
 * Use `std::allocator`-compatible buffers for seamless host<->device transfers (leveraging existing `CircularBuffer`/GR4 infra).
 * Explore post-processing fusion (e.g. log power, magnitude, abs²) within the same kernel or event chain (via `CircularBuffer<T, std::allocator>` paradigm).

### Design Notes
 * Batched FFTs are compute-dense and better amortise memory transfer cost compared to single FFT transforms.
 * Good real-world use cases: 
    * multi-channeliser, 
    * spectral correlation, 
    * GNSS acquisition, 
    * spectrograms (perhaps)
 * Reuse the internal GR4 FFT API shape and port structure where possible.
 * extension: defer support for non-power-of-two (via Bluestein) or SVD-style decompositions to future milestones.

### Rationale: What We Are *Not* Exploring First (and Why)

The following alternatives were evaluated but deliberately excluded from this first PoC due to their architectural or practical limitations. This documentation is intended to inform future contributors about paths that have already been considered, could be re-explored, or to simple prove me wrong. 😄

  * ❌ Single FFT
     * **Memory-bound**: a single transform offers too little compute to justify host↔device transfer latency.
   * **Poor GPU utilisation**: typical GR workloads (e.g. 256–4096 pt) don't benefit unless N > 1M.
  *  ❌ dtandalone FIR Filtering
    * **Low arithmetic intensity**: Especially for short FIRs (<128 taps), GPU offload costs more than it saves.
   * **Poor reuse**: No kernel fusion unless part of a larger windowed processing pipeline (e.g. overlap-save).
  * ❌ IIR Filtering
    * **Recursive structure**: Can't parallelise across data; inherently serial.
    * **Branch divergence**: Results in poor warp/subgroup execution efficiency on GPU.
  * ❌ Singular Value Decomposition (SVD)
    * **High cost of entry**: Effective only for large matrix sizes (e.g. N > 2000×200).
    * **Lack of tensor/matrix infrastructure**: GR4 does not yet support `Tensor<T>` or similar multi-dimensional views.

RalphSteinhagen:
## GR4 Performance: 🏗️ Can we fix it? Yes we can. 👷<br>*(Can we keep the diff small? …no we can't.)*

... but the benefits justify it: **one block source** — the same `processOne` / `processBulk` — now runs unchanged on **CPU, MCU, GPU, and over the wire**, because the value/tag types became plain-old-data (memcpy-able to USM or a socket) and every allocation became an injected policy, not a runtime fact.

The same foundation that fits an MCU also wins on the host — **+51 % tag-path throughput, −77 % resident memory**,  and a freestanding `-fno-rtti -fno-exceptions` build that allocates nothing after `start()`.

Two long-standing edge-memory pathologies have been cured along the way.

### **Motivation**

This PR addresses two long-standing issues that required a fundamental change. Both are harmless at small scale, and both degrade into near-unbounded growth as sample objects get large or the edge count rises.

1. The first relates to live-memory pathologies in GR4's `Graph` edges:

   - **Sample buffers of non-primitive elements grew without practically-limited bounds.**
     A `CircularBuffer` whose element owns heap storage (`std::vector<T>`, a map, or `Value`) kept each slot's
     largest-ever allocation alive for the buffer's lifetime — there is no immediate reclamation behind the reader
     until the writer re-uses the same slot. Resident memory tracks *worst-case payload × slots × edges*, so a single burst of large objects permanently inflates every edge it passed through.
     Previously this was contained only by limiting the number of slots per buffer; it now reclaims each slot's
     storage once the slowest reader has passed and it is known to be no longer needed.
   - **Tag-buffer capacity is coupled to the sample buffer.**
     The tag-slot count is tied to the paired sample count, so a tag-sparse stream over-provisions (idle tag storage on every edge) while a tag-dense or bursty stream under-provisions (dropped tags or stalls). With owning tag payloads the same unbounded-growth dynamic applies — resident memory is Σ(per-edge worst
     case), never the peak-concurrent set.

   Both share one root cause: **owning, individually heap-allocated payloads, sized for the worst case, living in
   the edge and never reclaimed.**

2. A second, orthogonal barrier is that allocation happens **at runtime, from the global heap** — which excludes the targets GR4 needs to reach:
   - **microcontrollers (MCUs) without a memory-management unit (MMU)** — no virtual memory and no usable heap; allocation must be static or drawn from a pre-provisioned arena;
   - **real-time (RT) hosts or MCUs with an MMU** — a heap exists, but any allocation after `start()` is a latency or abort risk; the RT contract is *no runtime heap allocation*;
   - **GPU / accelerator offload** — device kernels cannot call the host allocator; data must live in unified shared memory (USM) the device can map, and the value/tag types must be trivially-copyable into it.

Removing these barriers cannot be bolted on per block — it dictates the *foundations*: the value and tag types must be plain-old-data (no owning pointers, memcpy-able to USM or a wire), the memory boundary must be pluggable (every buffer and string drawn from an injected resource), and the hot path must be exception- and RTTI-free **before** any MCU or GPU backend can sit on top.

The diff is therefore large because it replaces foundations, not because it adds surface — which simultaneously shrinks cache footprint and code size and makes the steady state deterministic (detailed under *What this enables*; stated as design rationale, not measured headline figures).

These properties already pay off on the regular CPU and MCU targets today; just as importantly, the same foundation — contiguous, trivially-copyable value/tag types over a USM-compatible, zero-runtime-allocation memory model — is a prerequisite for the upcoming **GPU-acceleration** and **ML-integration** PRs, which need exactly this device-mappable layout and bounded memory model to share one block source across host, device, and model boundaries.

The change is *additive*: existing hosted graphs, plugins, YAML, and tests are unchanged, because an empty resource profile falls back to `get_default_resource()` (i.e. the default memory allocator).

## TL;DR -- The What?

- **Data model** — `Value` / `ValueMap` / `ValueMapView`, where the in-memory packed-blob layout *is* the on-wire and on-device byte image (zero serialise, zero-copy views).
- **Memory model** — `ResourceProfile` (data / tag / mechanics axes) + `StaticArenaResource<N>` + `NoHeapResource`, with PMR-clean Graph / Block / Port construction.
- **Freestanding proof** — `qa_Embedded` builds `-fno-rtti -fno-exceptions`; `Simple<externalStep>` drives the real graph from an external superloop; `qa_NoHeapScheduler` proves 0-alloc steady state.
- **Wire format V1** — CRC32C-framed `WireFormat` / `WireStream` over the same packed blob.

## Core changes, ranked by impact

Ordered essential → nice-to-have, each tied to the problem it addresses.

**Essential — the basics everything else builds on**

1. **Packed-blob `Value` / `ValueMap` / `ValueMapView`.**
   The in-memory layout (`Header + PackedEntry[] + payload`) *is* the on-wire and on-device byte image: one contiguous, offset-addressed, pointer-free block. It is more compact, lowers alloc/dealloc pressure, and a single `memcpy` moves it MCU↔CPU↔GPU↔socket with no serialise step — the prerequisite for all cross-target (MCU, CPU, GPU, network) transport.
2. **PMR memory model — allocation becomes an injected policy, not a runtime fact.**
   A polymorphic memory resource (PMR) `ResourceProfile` (data / tag / mechanics axes) threaded through Graph→Block→Port→buffers→strings, plus `StaticArenaResource<N>` (no heap after construction), `NoHeapResource` (aborts on stray allocation), PMR-clean construction, and a zero-capacity Port default.
   Result: **zero-allocation (static-PMR) or no-runtime-heap steady state**, with resources that are fully USM / unified-memory compatible, and the data/tag/mechanics split attributes memory to its source (sample buffers vs tag buffers vs mechanics). An empty profile falls back to `get_default_resource()`, so hosted behaviour is unchanged.
3. **RT-safe reclamation behind the reader** (`Clearable<T>` + `reclaimBehindReader`).
   Slot storage is cleared and handed back once the slowest reader passes, so resident memory tracks the **peak-concurrent working set, not Σ(worst case)** — the direct cure for pathology #1.

**High-impact**

4. **Tag storage decoupled from sample count** — `ByteRingBuffer<T>` (a variable-stride byte ring) and the now-default chunked tag buffer (`ChunkPool` / `TagChunkBuffer`) size tag storage to the tags actually in flight and reclaim idle chunks on the scheduler's housekeeping cadence — the cure for pathology #2 and the burst-vs-rare-tag imbalance.
5. **Freestanding subset** — RTTI-free (`-fno-rtti`), exception-free hot path (`#if __cpp_exceptions` gating), and a single `consteval type_name<>()` replacing per-type string machinery. This is what lets the binary fit an MCU flash budget; the same code- and data-size reduction improves cache locality on the hosted path.
6. **Wire format V1** — CRC32C-framed `WireFormat` / `WireStream` over the identical packed blob: the network leg, with no separate serialisation path to keep in sync.

**Enabling / nice-to-have**

7. **Freestanding `Scheduler` + `externalStep`** — drives the real `Graph` from a bare superloop (no thread pool, no `std::thread`), proving construction and wiring under an MCU execution model.
8. **Allocation-free steady-state tag publishing** and **cheapest-route tag forwarding** (settings-free pass-through copies the wire blob verbatim; only value substitution rebuilds).
9. **Executable cross-target invariants** — `qa_Embedded` (`-fno-rtti -fno-exceptions` compile) and `qa_NoHeapScheduler` (0 `operator new` / 0 arena growth in steady state) encode the guarantees as tests.

## What this enables — and why it is subtle

The interesting part is less "faster" (*which it also is*) than *which properties now hold at once* — several goals
usually pursued separately turn out to be the same decision:

- **One layout, three payoffs.** A contiguous, pointer-free blob is simultaneously zero-serialise on the wire, zero-copy mappable into a device address space, and cache-friendly (no per-entry node chasing).
  These are normally three independent engineering efforts; here they fall out of one representation.
- **The MCU constraint and the CPU/GPU performance goal pull the same way.** Shrinking code and data to fit a small flash and a tiny cache is the *same* work that lowers L1/L2/L3 pressure and cache-miss rates on a workstation. 
  A change made "for the MCU" is therefore a throughput win on CPU and GPU, not a trade-off against them.
- **Determinism is a capability, not just a metric.** PMR-everywhere with a zero-allocation steady state makes worst-case execution time analysable and lets a graph run where a single `malloc` would abort — the difference between "runs on a PC" and "runs in a hard-RT control loop, on a no-MMU MCU, or inside a GPU kernel, from the same block source."

Concretely, the class of applications this opens up — none reachable while the value and tag types owned heap memory and the hot path allocated at runtime — includes bare-metal / RTOS edge DSP, bounded-latency RT control loops, GPU offloaded flowgraphs that share the *exact* block code with the host, and wire-compatible distributed graphs over the network.

### `Value` / `ValueMap` / `ValueMapView`

`ValueMap` is a **packed blob** — `Header + PackedEntry[] + payload`. The in-memory layout *is* the wire/USM image, and the view type carries no owning resource pointer, so it is trivially-copyable into a kernel or a static buffer:

```cpp
ValueMap m = tag.map;                          // owning, on a PMR resource
std::span<const std::byte> wire = m.blob();    // the bytes — memcpy to disk / socket / USM, no serialise
ValueMapView v = ValueMap::makeView(wire);     // zero-copy read over external bytes
static_assert(std::is_trivially_copyable_v<ValueMapView>);   // capturable by value into a SYCL kernel
```

- `makeAt(slot, …)` initialises a USM/ring slot without allocating; `assignFrom(view, resource)` **re-homes in
  place** (memcpy into the existing blob when capacity suffices, keeping the resource) instead of destroy+reconstruct.
- Copy/move-assign **preserve the target's PMR resource**.

### Tag handling

- `BasicTag<Owning>` NTTP: `Tag = BasicTag<true>` (owning — the API type for settings / merge / store-past-span); `BasicTag<false>` = `ValueMapView` (the cross-edge / device-side variant). `asView()` bridges owning → view.
- **Documented lifetime contract** on `asView()` / `Port::tags()`: a `BasicTag<false>` is valid only within its originating reader span; outliving or forwarding it requires materialising to owning via `assignFrom`.
- `ByteRingBuffer<T>` — a variable-stride byte ring (SISO / SIMO / MIMO) reusing `CircularBuffer`; the intended zero-copy cross-edge tag carrier and the owning tag buffer, now available opt-in (the chunked buffer below is the default).
- **Tag forwarding takes the cheapest route per tag** — classified once: settings-free pass-through forwards the wire blob verbatim (one buffer copy, no key-by-key rebuild); drop-only filters in place; only value substitution rebuilds. This also removed a hidden per-tag owning-map copy the read-view flip had introduced.

### Chunked tag buffer (now the default)

The default tag buffer is now `TagChunkBuffer<>` — a drop-in `Port` tag buffer that stores a fixed-stride `BasicTag<false>` descriptor ring and serialises each tag's packed blob into pooled byte chunks (`ChunkPool`), the future USM-mapping unit.

Per-edge buffers draw chunks from the injected tag resource (an arena on MCU, not the global heap) and return idle chunks on the scheduler's `houseKeeping(depth)` cadence (Shallow keeps them on a free-list for fast re-acquire, Deep returns them upstream), so resident RAM tracks the peak-concurrent working set rather than Σ(per-edge worst case).
The owning `ByteRingBuffer<Tag>` remains available opt-in via `TagBufferType<ByteRingBuffer<Tag>>`.

Why it is the default: on `qa_PerformanceMonitor` it roughly halves resident tag memory at neutral throughput — the
owning ring's fixed per-edge pre-reservation versus the chunk pool's working-set sizing, a delta present even with zero tags (see *Performance*). The hard cap makes back-pressure exception-free (a full pool returns an empty span, never a caught `bad_alloc`), and an oversized size-class keeps a jumbo forwarded tag from being dropped.
The one caller-visible consequence is the tightened lifetime contract above — consumers now read non-owning `BasicTag<false>` views over the chunk bytes; the core tag-consuming blocks (Selector, Trigger, DataSink, SyncBlock) were audited to forward through the deep-copying `publishTag`, and `qa_Port` / `qa_DynamicPort` exercise both variants under AddressSanitizer.

### Memory model & zero-allocation on MCU

- **`ResourceProfile { data, tag, mechanics }`** — three orthogonal `std::pmr::memory_resource*` axes threaded from the Graph constructor through Block, Port, buffers, and strings. Empty ⇒ `get_default_resource()`, so hosted behaviour is unchanged.
- **`StaticArenaResource<N>`** — span-based bump arena (no heap after construction). **`NoHeapResource`** — aborts on `do_allocate`; install as the default in heap-discipline tests / MCU builds to surface stray allocations at the call site.
- **PMR-clean construction** — the profile is honoured at member-init time, so even setup lands on the arena, not the global heap. **Zero-capacity Port default** + wiring-time materialisation keeps the merge-API / embedded paths heap-free until they actually wire up.
- Buffers (samples and tags) grow their slots on first touch from the buffer PMR and are reclaimed behind the slowest reader; an MCU superloop calls `scheduler.step()`, and steady state touches neither global `new` nor the arena.

```cpp
// MCU / RT: every graph allocation drawn from a static arena — no heap, ever
auto arena = std::make_unique<gr::pmr::OwnedStaticArenaResource<1UZ << 23U>>();      // 8 MiB
gr::Graph graph{gr::ResourceProfile{.data = arena.get(), .tag = arena.get(), .mechanics = arena.get()}};
// ... emplaceBlock / connect ...   (an empty profile ⇒ get_default_resource(); hosted graphs unchanged)

gr::scheduler::Simple<gr::scheduler::ExecutionPolicy::externalStep> sched;
std::ignore = sched.exchange(std::move(graph));
std::ignore = sched.changeStateTo(gr::lifecycle::State::RUNNING);   // primes; spawns no worker thread
while (running) { sched.step(); }                                   // bare superloop; 0 new, 0 arena growth
```

### Wire format

`WireFormat` / `WireStream` + scalar/SIMD `CRC32C`: `StreamHeader` (GR4W magic + V1), `FrameEnvelope`, monotonic sequence, and `BadSequence` / `BadCrc` / `Truncated` status — the networking leg of the same blob.

## Performance

A per-commit sweep of `qa_PerformanceMonitor` over the whole branch (`before` → 21 commits → tip), gcc-15 `-O3`, pinned to one core (`taskset`), **one fixed harness at every commit** (so each point varies only by core code), 5 × 30 s per use case, peak `Rate` and mean resident over the post-10 s window.

Three use cases: **UC1** no tags, **UC2** 1 tag / 10 k samples, **UC3** 1 tag / sample (heaviest).

perf_progress_uc3

**Throughput** (peak, Δ vs `before`; here AMD Ryzen 9 7940HS w/ powersave governor):

| commit                              | UC1 no-tags  | UC2 1/10 k   | UC3 1/sample        |
|-------------------------------------|--------------|--------------|---------------------|
| before                              | 606 MS/s     | 472 MS/s     | 0.544 MS/s          |
| USM-ready `ValueMap`                | 610 (+0.6 %) | 488 (+3 %)   | 0.635 (**+17 %**)   |
| PMR-clean construction              | 578 (−5 %)   | 485 (+3 %)   | 0.681 (**+25 %**)   |
| allocation-free tag publish         | 614 (+1 %)   | 510 (+8 %)   | 0.800 (**+47 %**)   |
| **tip** (ChungTagBuffer default)  | 617 (+2 %)   | 529 (+12 %)   | **0.922 (+69 %)**   |

UC1/UC2 are bandwidth-bound (≈ flat); the tag-bound UC3 carries the gains — **+51 % end-to-end**, in three steps: the packed `ValueMap`, PMR-clean construction, and allocation-free steady-state tag publishing.

**Resident memory** (mean MB, Δ vs `before`):

| commit                              | UC1            | UC2            | UC3            |
|-------------------------------------|----------------|----------------|----------------|
| before                              | 115            | 138            | 138            |
| USM-ready `ValueMap`                | 393 (+242 %)   | 401 (+189 %)   | 401 (+189 %)   |
| PMR-clean construction              | 356 (+210 %)   | 356 (+157 %)   | 356 (+157 %)   |
| allocation-free tag publish         | 56 (−51 %)     | 58 (−58 %)     | 58 (−58 %)     |
| **tip** (chunk tag buffer default)  | **31 (−73 %)** | **31 (−77 %)** | **31 (−77 %)** |

Net **−77 %** resident, but not monotonic: the `ValueMap` V2 commit transiently *tripled* resident (over-reservation, later reclaimed), allocation-free in-place re-home cut it 356 → 56 MB, and the chunk-buffer default took it to 31 MB.
Resident is ≈ use-case-independent here, so this tracks *overall* process memory rather than tag-specific storage.

**Steady-state is alloc-free:** `qa_NoHeapScheduler` drives the real `Graph` + `Simple<externalStep>` for 10,000 cycles with **0 `::operator new` + 0 arena growth** under `-fno-rtti -fno-exceptions` — the MCU "no malloc after `start()`" promise, on the real graph (not a proxy).

> Methodology: absolute rates reflect the powersave governor (consistent, below turbo); run-to-run spread is
> ~2–3 %, so single-digit-% deltas sit at the noise floor. The fixed harness draws from the global heap + mmap ring,
> isolating core-code effects; the optional PMR-pool configuration costs a further ~2 % (an indirection trade for
> determinism / USM-compatibility, not a regression).

## **How to review** (≈16.5 k lines)

After squashing the LLM-based and manual review fixes into their targets, commit by commit, the commit history is clean. 

Suggested order:

1. **Data model (the spine)** — the `Value`/`ValueMap` PMR-assign fix → the USM-ready packed-blob `Value`/`ValueMap` (largest, most important commit) → in-place re-home + `BasicTag<false>` view contract.
   Focus on the blob layout, the POD view types, and the `assignFrom` capacity/resource fast path.
2. **Tag transport** — RT-safe `CircularBuffer` housekeeping → `ByteRingBuffer` + `BasicTag<Owning>` NTTP → the default-tag-buffer flip. Confirm `ByteRingBuffer` is a thin reuse of `CircularBuffer` and the flip is neutral.
3. **Freestanding / MCU subset** — RTTI-free subset + `qa_Embedded` → exception-free subset + `gr::log`
   → `StaticArenaResource` → consteval `type_name` → freestanding `Scheduler` + `externalStep`.
   Focus on the `#if __cpp_exceptions` gating discipline and that `externalStep` adds a policy, not a fork.
4. **PMR-clean construction** — `ResourceProfile` / scope-guard threading + the zero-capacity Port default.
5. **once (1)-(4) are done** — start reading chronologically i.e. commit-by-commit.

The following new unit tests encode the cross-target invariants — start there to see what the branch *guarantees*:

- 'qa_PerformanceMonitor' shows the existing use-case and (optional) PMR specialisation,
- `qa_Embedded` (`-fno-exceptions` AOT proof), and
- `qa_NoHeapScheduler` (0 global-new / 0 arena-growth steady state).


RalphSteinhagen:
*N.B. needed to force-push to do some final CI clean-up/fix-ups (compile-time notes and warning on Emscripten, AdaptiveCpp).

RalphSteinhagen:
Will need to force-push again to address some x86 32-vs-64-bit type strictness issues (WASM vs native) + an additional commit because the Linux kernel apparently changed the `/dev/ptpN` defaults.

the-argus:
Thanks for all this work! I am very happy that ValueMap merges a lot of nested if statements that would find and check type separately, among other things.

My review here is small, I only really got through Value/ValueView/ValueMap/ValueMapView, and some PMR and testing related stuff. I only have a few line specific comments so I will just include those here for conciseness.

- the commit `fix(core): Value copy/move-assign preserve target's PMR resource` commit changes value to have different allocator propagation semantics than standard containers, standard containers use default resource on copy. IMO it is better to follow the semantics of standard containers as I don't think either method is more intuitive than the other so I'd rather just be consistent, but I will not die on this hill, :)

- re: allocators + tests, might be good to have a test that verifies that a nested Value(ValueMap(Tensor(...))) correctly propagates its allocator to all nested containers in the expected way, on copy assign, copy construct, allocator extended copy construct, moves, etc. ATM I don't think I see this functionality, if a Value contains an allocator aware container and does `copy_from` it looks like it only copies the container into the new allocator's memory but the inner container will keep its allocator.

Overall/otherwise this LGTM, but I lack some context about this project. Here are some further questions/concerns, maybe useful, maybe not:

- What is the purpose of the wire format Annotation, exactly? just curious, it seems fine but it unused so far here. Although I am also wondering why it is per-item, at a glance it looks semantically closer to typeinfo, ie. duplicated many times, to me I would define/declare typeinfos first and then things would just have numeric typeids that refer to a preexisting typeinfo. just my intuition though, not sure.

- Values are now indirect types with value semantics, so they can now have variable size, primarily to facilitate the presence of variable sized fields: typeName string, annotations strings, and an optional checksum. as per my question above about integer typeids, maybe values could just be by value if these fields were not present, then ValueView is just Value&.

- the wire format includes binary serialized C++ types. These can include containers like `gr::pmr::vector` which internally contain pointers. I am not sure about the purpose of the serialization here, I am not familiar with the SYCL and USM usecases, so I guess things share address space, and this is fine?

### nitpicks/questions:

- I'm noticing a naming convention here where semantically private/protected member functions have a leading underscore, is that typical? (I have only been doing this for member variables)

- the `CountingResource` in the tests has been duplicated three times, might be worthwhile to make some util header for that.

- This stuff in `Value.hpp` seems paranoid, IMO I would use an assert for something that seems like it would be a programmer error that should not happen, or if this is possible to happen then it should be this way in release mode, too:
```cpp
        // Defence-in-depth bound check on the rank/elementCount bytes. The trusted ingestion
        // path (`from_blob` -> `_validateTensorSubBlob`) already enforces these; the gate
        // exists for `Value::makeView`-injected bytes that bypass validation.
        const std::size_t extentsBytes = paddedTensorExtentsBytes(rank);
        if constexpr (gr::meta::kDebugBuild) {
            if (payloadByteCount() < kTensorBlobHeaderSize + extentsBytes) {
                return std::nullopt;
            }
        }
```

- double private specifier in `Value.hpp` at line 626

- Value move constructor `Value.cpp` line 177 could use std::exchange

- StaticArenaResource could use `std::align` or maybe even just be a wrapper around `std::pmr::monotonic_buffer_resource`

core/include/gnuradio-4.0/Tag.hpp
```diff
@@ -61,82 +72,112 @@ struct checked_access_ptr {
...
-                }
-            } else {
-                // If the key doesn't exist, just insert
-                dest.insert({key, value});
-            }
-        } else {
-            // If it's not a nested map, insert/replace the value
-            dest[key] = value;
-        }
+        dest.insert_or_assign(std::string_view{key}, value);
```
ivan-cukic, line 180: Is this intentional behavior change - instead of recursively merging nested maps, just doing the first level?

ivan-cukic (approved):

RalphSteinhagen:
@the-argus  thanks a lot for the careful review — much appreciated.  Status on each:

- [x] **Leading `_` on private/protected member functions** — this is a deliberate GR4 convention. We try to avoid excessive `public:`/`private:` sections to minimise duplicate getter/setter boilerplate, and add an access specifier only where a cross-constraint/invariant genuinely has to be protected across multiple member variables. The open policy also makes prototype extension and unit-testing of a class much easier. A leading `_xxx` in production code is a clear "treat-as-internal / code-smell" marker for us — we flag it, but deliberately don't suppress it categorically.
- [x] **`CountingResource` duplicated** — investigated; the three aren't interchangeable, so I've held off on a mechanical merge: `qa_BlockHouseKeeping`'s is `std::atomic` because that test is concurrent (the non-atomic canonical would be a data race), and `qa_MemoryAllocators`' (pre-existing) carries `reset()` + separate alloc/dealloc byte counters. The canonical `gr::allocator::pmr::CountingResource` already covers the simple case (new tests like `qa_Value` use it). I'll consolidate the safely-mergeable variants into the shared header as a follow-up.
- [x] **"Paranoid" debug-only bound check in `Value.hpp`** — agreed, converted both to `assert(...)`. `makeView` is documented INTERNAL/UNSAFE (the caller guarantees a verified record; untrusted bytes go through `from_blob`), so a short blob here is a precondition violation — `assert` states that intent, fails loudly in debug, and compiles out in release.
- [x] **Double `private:` in `Value.hpp`** — removed the redundant specifier.
- [x] **Move constructor `std::exchange` (`Value.cpp`)** — done; the body is now an empty member-init list using `std::exchange`.
- [x] **`StaticArenaResource` / `std::align`** — adopted `std::align` in `do_allocate` (it also subsumes the hand-rolled overflow check). Kept it as a dedicated resource rather than wrapping `std::pmr::monotonic_buffer_resource`, since the tests rely on its `used()` accounting and the `fatal`-on-exhaust diagnostic, neither of which `monotonic_buffer_resource` exposes.

core/include/gnuradio-4.0/Tag.hpp
RalphSteinhagen: @ivan-cukic good catch — yes, this is intentional. 

With the new byte-blob-resident `ValueMap`, an in-place recursive merge would require extract, materialise, mutate, re-serialise, reassign at each nesting level, so `updateMaps` is now a shallow `insert_or_assign` and recursive semantics are the caller's responsibility (noted in the comment). The only caller (`Settings`, folding unmatched parameters into `meta_information`) operates on flat leftovers, so it doesn't rely on the deep merge. If we find a real need for recursive merge later I'll add an explicit `mergeRecursive` helper rather than put that cost on the hot path. 

Thanks for the review/approval!

core/include/gnuradio-4.0/WireFormat.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+#include <cstddef>
+#include <cstdint>
+#include <cstring>
+#include <optional>
+#include <span>
+#include <string_view>
+
+namespace gr::wire {
+
```
drslebedev, line 13: Should we add this just to be on the safe side?
```cpp
static_assert(std::endian::native == std::endian::little, "gr::wire V1 requires little-endian hosts");
```

core/include/gnuradio-4.0/WireFormat.hpp
```diff
@@ -0,0 +1,156 @@
...
+ * skips and type-dispatches any element from the first 8 bytes alone. Fresh
+ * design, hard cutover; full spec
+ * and rationale in docs/CORE_WireFormat.md.
+ *
+ *   [0] u32 size           element → next element; == capacity (skip field)
+ *   [4] u8  valueType      kValueTypeEscape (0xFF) ⇒ precise type in typeName
+ *   [5] u8  containerType  Scalar|Complex|String|Tensor|Map|…
+ *   [6] u8  flags          kFlagReadOnly | kFlagChecksumPresent | kFlagAnnotationsPresent | …
+ *   [7] u8  payloadOffset  payload start; == kPrefixBytes when no optional region
+ *   [8] …   optional region (typeName iff escape; annotations iff flagged; checksum iff flagged)
```
drslebedev, line 32: Is this only partially implemented now?
I found only  annotations read/write.

core/include/gnuradio-4.0/WireFormat.hpp
```diff
@@ -0,0 +1,156 @@
...
+inline constexpr std::size_t kPrefixBytes      = 8UZ; // also the default payloadOffset
+static_assert(kPrefixBytes <= 255UZ, "payloadOffset is stored in a u8");
+
+[[nodiscard]] inline std::uint32_t elementSize(const std::byte* h) noexcept {
+    std::uint32_t s;
+    std::memcpy(&s, h + kOffSize, sizeof(s));
+    return s;
+}
+inline void setElementSize(std::byte* h, std::uint32_t size) noexcept { std::memcpy(h + kOffSize, &size, sizeof(size)); }
+
```
drslebedev, line 61: Should we do some extra validation when parsing bytes? For example:
- size >= kPrefixBytes
- payloadOffset >= kPrefixBytes
- payloadOffset <= size
- flags has no unknown bits set 
- valueType / containerType are known enum values

core/include/gnuradio-4.0/WireStream.hpp
```diff
@@ -0,0 +1,229 @@
...
+/// Serialise stream header into `out` (must be ≥ kStreamHeaderBytes; debug-asserted).
+inline void writeStreamHeader(std::span<std::byte> out, const StreamHeader& h) noexcept {
+    assert(out.size() >= kStreamHeaderBytes);
+    std::memcpy(out.data() + 0, h.magic.data(), 4);
+    std::memcpy(out.data() + 4, &h.version, sizeof(h.version));
+    std::memcpy(out.data() + 6, &h.flags, sizeof(h.flags));
+}
+
+/// Parse a stream header; returns nullopt on magic / version mismatch.
+[[nodiscard]] inline std::optional<StreamHeader> readStreamHeader(std::span<const std::byte> in) noexcept {
```
drslebedev, line 53: Same as for WireFormat: Should we do some extra validation when parsing bytes?

core/include/gnuradio-4.0/WireStream.hpp
```diff
@@ -0,0 +1,229 @@
...
+    /// `std::bad_alloc` from `out.resize()` on allocation failure.
+    void writeFrame(std::vector<std::byte>& out, std::span<const std::byte> payload) {
+        if (!_headerWritten) {
+            writeHeader(out);
+        }
+        const auto envPos     = out.size();
+        const auto payloadPos = envPos + kFrameEnvelopeBytes;
+        out.resize(payloadPos + payload.size());
+
+        FrameEnvelope env{.frame_size = static_cast<std::uint32_t>(payload.size()), .sequence = _sequence++, .crc32c = 0U};
```
drslebedev, line 128: I hope that we do not have payloads >4Gb

core/include/gnuradio-4.0/Value.hpp
```diff
@@ -110,88 +193,168 @@ constexpr std::string value_to_string(const Value&); // forward declaration
...
+ *
  * @warning Unchecked accessors have undefined behavior on type mismatch (assert in debug).
  *          Use holds<T>(), get_if<T>(), or value_or() for safe access.
+ *
+ * @par Wire Format
+ *
+ * A Value has no standalone wire format — it is a host-side handle. It acquires a wire form
+ * only when **stored inside a container**: a `PackedEntry` row of a `gr::pmt::ValueMap`, or a
+ * `PackedTensorElement` row of a `Tensor<Value>`. Both row types carry the type tag bytes
+ * (`valueType`, `flags`) plus an `inlineValue` slot (8 B) AND a `payloadOffset` / `payloadLength`
```
drslebedev: PackedEntry has no inlineValue - every value is stored in the payload, reached via payloadOffset/payloadLength. Only PackedTensorElement has inlineValue.

core/include/gnuradio-4.0/ValueMap.hpp
```diff
@@ -0,0 +1,3588 @@
...
+                }
+
+                if (e.payloadLength < kRecHeaderBytes) {
+                    return std::unexpected{DeserialiseError::CorruptOffset};
+                }
+                const std::uint32_t recordSize         = gr::wire::elementSize(bytes.data() + e.payloadOffset);
+                const std::uint32_t expectedRecordSize = alignToRecord(e.payloadLength);
+                if (recordSize != expectedRecordSize) {
+                    return std::unexpected{DeserialiseError::CorruptOffset};
+                }
```
drslebedev, line 2451: Add record's own `payloadOffset` check. 

```cpp
if (gr::wire::payloadOffset(bytes.data() + e.payloadOffset) != gr::wire::kPrefixBytes) {
     return std::unexpected{DeserialiseError::CorruptOffset};
}
```

core/include/gnuradio-4.0/ValueMap.hpp
```diff
@@ -0,0 +1,3588 @@
...
+// Sentinels — see range table above.
+inline constexpr std::uint16_t kIdUnknown    = 0x0000;
+inline constexpr std::uint16_t kInlineKeyId  = 0x8000;
+inline constexpr std::uint16_t kSpilledKeyId = 0x8001; // key bytes spilled to payload pool (>27 chars)
+inline constexpr std::uint16_t kEndMarkerId  = 0xFFFF;
+
+// Authoritative canonical key list. Names + C++ types pinned from
+// Tag.hpp:196-216 (DefaultTag<Name, Type, Unit, Description> declarations).
+// Value::ValueType is the YaS-inspired tag used in PackedEntry.valueType.
+inline constexpr std::array<CanonicalKey, 20> kCanonical = {{
```
drslebedev, line 100: Minor maintainability issue:
Use existing constants e.g. `tag::SAMPLE_RATE.shortKey()` 
In addition on can guard the count (and ideally the names) with a static_assert so a missing/extra entry fails the build:
```cpp
static_assert(keys::kCanonical.size() == gr::tag::kDefaultTags.size(), "kCanonical and kDefaultTags must list the same tags");
```

drslebedev (approved):
First of all, thanks for the PR.

  This is a large change, and I hope I understood the underlying idea well enough to review it properly in the limited time available.

  Overall, I think the direction and design look very solid. The key refactoring - storing Value/ValueMap as one flat blob that can be used directly both as the wire format and for data exchange between CPU and GPU - is a good architectural move. It also avoids per-tag heap allocations and makes tags cheap to copy.

  My main concern is validation at the raw-byte boundary. The implementation uses raw byte views and raw pointers heavily, so malformed input must be rejected before it becomes a usable in-process object. The common fix is to validate the full blob shape before accepting it: header, payload range, record metadata, and so on.

  I also like the new TagChunkBuffer implementation a lot, and the overall performance results look very impressive.

  I did not find any critical problems in the changes. Overall, the PR can be merged once the inline comments are resolved.

core/include/gnuradio-4.0/WireStream.hpp
RalphSteinhagen: We won't... at least not in one blob/single TCP frame... ;-)

core/include/gnuradio-4.0/ValueMap.hpp
RalphSteinhagen: Not use here because the definition would need to pull-in `Tag.hpp` which would then be a circulatory dependency.

RalphSteinhagen:
@drslebedev thanks a lot for the careful review 

What I've followed up on and merged:
 - [x] **Little-endian `static_assert` (WireFormat.hpp:12)**
 - [x] **Extra validation when parsing bytes (WireFormat.hpp:55)**
 - [x] **Same validation for WireStream (WireStream.hpp:53)** — `WireStreamReader` already validates magic / version / frame-extent and verifies CRC. There was one real gap: the frame-bounds test used `envEnd + frame_size > size()`, which wraps on 32-bit `size_t` (WASM) and lets a corrupt `frame_size` through. Switched to the overflow-safe version.
 - [x] **Payloads > 4 GB (WireStream.hpp:128)** — as noted in-thread, a single blob / TCP frame won't exceed 4 GB by design; the `u32` frame size is intentional.
 - [x] **`PackedEntry` has no `inlineValue` (Value.hpp:222)** — correct, the doc was wrong and self-contradictory with the `PackedEntry` definition. fixed.
 - [x] **Record's own `payloadOffset` check (ValueMap.hpp:2451)** — added your exact check (`gr::wire::payloadOffset(...) != gr::wire::kPrefixBytes → CorruptOffset`). While there I also corrected the `[size:4][vt:1][ct:1][flags:2]` comments
 - [x] **Use existing constants / `static_assert` count (ValueMap.hpp:100)** — as noted, referencing `tag::SAMPLE_RATE.shortKey()` would pull in `Tag.hpp` and form a cycle.

Merged the fixes. Will merge the PR once the first CI targets pass.

PR #806, opened 2026-07-03
Enable yaml importer and yaml messages on `-fno-exceptions`/embedded targets

### LINKED ISSUE #779: [1SP;3SP] Convert the remaining exception throwing in GR to expected
- Add error kinds to denote which class of errors happened. It should
contain Status, Warning, Error and Notification for the time being.
- Add kinds to currently existing errors
- Convert remaining exception throwing to errors

the-argus:
Loading graphs from yaml when using `-fno-exceptions` is the main change here.

This PR has been altered, initially it was intended to remove all (avoidable) internal uses of exceptions, now it is only removing exceptions from the YAML importer.

There is an additional commit which changes how the CI caches fetchcontent dependencies. Previously, `restore-keys` was used, but from a bit of research it seems that does not make sense, because FetchContent is not a progressive cache. ie. `restore-keys` would be used for something like ccache where an old or partial cache can still be used and then updated to be more recent. But FetchContent will not update anything if the deps are already cloned, so you just get an out of date cache entry.

~~Otherwise I just tried to remove all instances of `throw` and `#if __cpp_exceptions` as much as possible. The ones that remain should only be ones dedicated to maintaining strong exception safety in the case of exceptions thrown by the user.~~

~~One sort-of exception (ha) to that is in `SchedulerBase::processScheduledMessages()`:~~
```cpp
        if (this->msgOut.buffer().streamBuffer.n_readers() == 0) {
            // nobody is listening on messages -> escalate otherwise-ignored child errors
            for (const auto& msg : messagesFromChildren) {
                if (!msg.data.has_value()) {
#if __cpp_exceptions
                    throw gr::exception(std::format("scheduler {}: throwing ignored exception {:t}", this->name, msg.data.error()));
#else
                    std::ignore = gr::log::error(std::format("scheduler {}: ignored child error {:t}", this->name, msg.data.error()));
#endif
                }
            }
            return;
        }
```

~~Which I am treating as some sort of weak exception propagation (ie not aborting when exceptions are disabled). The only thing here is that the user might set a `message->data = Error {...}` and then the message gets ignored and turned into a thrown exception here, in which case GR is throwing exceptions when the user is not. I am considering just having this always use `gr::log::error`. Especially if in the future the user can register log handlers/hooks. Feedback appreciated.~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,10 +122,7 @@ jobs:
           cache-name: cache-fetchContent-cache
         with:
           path: ${{runner.workspace}}/build/_deps
-          key: fetchContent-${{ runner.os }}-${{ matrix.compiler.label || matrix.compiler.cc }}-${{ matrix.cmake-build-type }}-${{ hashFiles('CMakeLists.txt') }}
-          restore-keys: |
-            fetchContent-${{ runner.os }}-${{ matrix.compiler.label || matrix.compiler.cc }}-${{ matrix.cmake-build-type }}
-            fetchContent-${{ runner.os }}-${{ matrix.compiler.label || matrix.compiler.cc }}
+          key: fetchContent-${{ runner.os }}-${{ matrix.compiler.label || matrix.compiler.cc }}-${{ matrix.cmake-build-type }}-${{ hashFiles('CMakeLists.txt', 'patches/**') }}
 
       - name: Configure
         if: matrix.compiler.cmake_wrapper == null && matrix.compiler.cxx != 'acpp'
--- a/.github/workflows/latest.yml
+++ b/.github/workflows/latest.yml
@@ -54,9 +54,7 @@ jobs:
         uses: actions/cache@v5
         with:
           path: ${{ github.workspace }}/build-${{ matrix.preset }}/_deps
-          key: fetchContent-${{ matrix.preset }}-${{ hashFiles('CMakeLists.txt') }}
-          restore-keys: |
-            fetchContent-${{ matrix.preset }}
+          key: fetchContent-${{ matrix.preset }}-${{ hashFiles('CMakeLists.txt', 'patches/**') }}
 
       - name: Configure ccache
         run: |
--- a/.github/workflows/tag_builds.yml
+++ b/.github/workflows/tag_builds.yml
@@ -35,9 +35,7 @@ jobs:
         uses: actions/cache@v5
         with:
           path: ${{ github.workspace }}/build/_deps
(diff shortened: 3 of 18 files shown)


core/include/gnuradio-4.0/Logger.hpp
```diff
@@ -2,56 +2,20 @@
...
 #include <source_location>
 #include <string>
 #include <string_view>
 
+#include <gnuradio-4.0/meta/exception.hpp>
 #include <gnuradio-4.0/meta/formatter.hpp> // std::formatter<std::source_location>
 
 namespace gr {
 
-/**
```
RalphSteinhagen, line 132: This `gr::exception` needs to be kept. While the core code is supposed to be exception free, user-defined code may still (optional) throw exceptions. Some libraries and when using heap-allocating functions may still throw. I'd rather have these exceptions been converted into `gr::exception` which is instrumented.

Again, this for user-defined code only, not for the GR4 core.

core/include/gnuradio-4.0/Logger.hpp
```diff
@@ -67,8 +31,6 @@ struct Error {
 
     explicit Error(const std::exception& ex, std::source_location location = std::source_location::current()) noexcept : Error(ex.what(), location) {}
```
RalphSteinhagen: still needed, as indicated above.

core/include/gnuradio-4.0/MemoryAllocators.hpp
```diff
@@ -52,11 +52,7 @@ struct Aligned {
         if (void* p = ::operator new(bytes, std::align_val_t{alignment})) {
             return static_cast<T*>(p);
         }
-#if __cpp_exceptions
```
RalphSteinhagen, line 55: keep

core/include/gnuradio-4.0/PmtCollections.hpp
```diff
@@ -64,21 +64,13 @@ struct vector<T, false> { // trivially copyable, no lifetime management
 
     [[nodiscard]] constexpr T& at(std::size_t i) {
         if (i >= _size) [[unlikely]] {
-#if __cpp_exceptions
```
RalphSteinhagen, line 67: also keep.

RalphSteinhagen (approved):
@the-argus thanks for following-up minimising the usage of exceptions in the core. I think the ones you tackled are the right ones. 👍 

**N.B. yust for info: we shouldn't suppress OS/IO-specific exceptions but just forward them. They are largely Linux (or larger-OS) related and those blocks rarely called/used in an embedded, for example, MCU context.**

PR #807, opened 2026-07-09
feat(core): record-based logging as a fault/telemetry channel + first device-execution seam

RalphSteinhagen:
The previous logger was a thin wrapper that formatted eagerly at the call site into a `std::string` and threw on fatal. 
That shape did not yet include the targets the ValueMap[View]/PMR foundation (#804) was built for:

- a **real-time thread** cannot pay an unbounded `std::format` + heap allocation per record;
- a **freestanding MCU build** has no console, no exceptions, and no heap to queue into;
- a **producer that cannot allocate** (eventually: a device kernel) has nowhere to put the
  message at all.

The underlying decision: GR4 logging is a **fault and telemetry channel, not an operational logging framework**. Either there is a fault/warning worth a bounded record, or there is nothing — hence fixed-size records, a compile-time severity gate instead of runtime filtering, and no dependency on a logging thread to make records visible.

The same PR takes the first — deliberately inert — step towards heterogeneous execution, so that the upcoming GPU/SYCL work lands on a prepared seam instead of touching `Block::work` mechanics again.

## What's new:

**Logging (`gr::log`, `docs/USER_API_Logging.md`)**

- Emitters (`fatal`/`failure`/`error`/`warning` always-on; `info`/`debug`/`trace` compiled out
  unless `kDebugBuild`) take compile-time-checked format strings (`gr::log::runtime(...)` for
  the explicit runtime case) and publish a fixed-size, trivially-copyable `LogRecord`
  (512 B text, 96 B location, truncation flagged, drop-count carried in-band as
  `droppedBefore`) through a pluggable `Backend` seam (`setBackend`/`drain`/`flush`).
- The hosted default backend is **write-through**: publish pushes through a multi-producer
  ring and flushes to the console on the same thread — a `fatal` record is visible even if
  the process dies immediately after; no pump thread exists to fall behind.
  `error`+`warning`+`fatal` → stderr, rest → stdout (compile-time configurable).
- `EMBEDDED` builds default to `HistoryLoggerBackend`: zero heap, last 20 records retained
  in place, drainable/snapshotable post-mortem — the MCU "black box".
- `fatal` remains the single quarantine point: throws `gr::exception` on hosted targets,
  publishes a best-effort record and aborts under `-fno-exceptions`.

**Device-execution foundation (inert by design)**

- `gr::DeviceEligible` trait family (`AutoParallelisable` = const+noexcept `processOne`,
  `HasSyclBulk`, `HasShaderFragment`) and a dispatch seam in the work path: a device-eligible
  block whose `compute_domain` parses non-host warns once and continues on the CPU path.
  CPU-only and freestanding builds are bit-identical to before.
- `ValueMapView::formatAt` exposes the in-place wire-format initialisation the owning factory
  already performed, so a producer can format a map into borrowed storage (e.g. a USM slot)
  with bounded growth — allocation-, exception- and libc-free. Input validation rejects
  zero/oversized entry capacities and overflow-wrapping size requests rather than writing.
  Key comparison on the lookup path is now byte-wise (`detail::keyEquals`): the previous
  equality lowers to `bcmp`/`strlen`, which a CUDA JIT cannot resolve.
- `work::Status`/`work::Result` + accounting helpers moved to `WorkStatus.hpp` so
  freestanding/MCU TUs use them without pulling the full `Block.hpp` machinery.

Bottom-up by commit is probably the low-effort path for reviewing; 
each commit builds and CI tests on its own:

## Review notes / behavioural changes

- `gr::log::error(...)` now returns `void` (previously usable as a `gr::Error` factory);
  the two `thread_pool.hpp` call sites that relied on it construct the `Error` explicitly.
- Severity is a **compile-time** gate — there is deliberately no runtime log level.
- Formatted emitters still allocate transiently via `std::vformat`; the `string_view`
  overloads and the record path itself are allocation-free. A fully allocation-free
  formatted path (and device-side records: hashed source locations, clock-domain-tagged
  timestamps) is scoped for the device-integration follow-up.
- `compute_domain` currently doubles as IO-pool name and device selector; disentangling it
  is deferred to the follow-up that wires real dispatch.


core/src/Logger.cpp
```diff
@@ -0,0 +1,420 @@
...
+        return index;
+    } else {
+        return static_cast<std::size_t>(index);
+    }
+}
+
+} // namespace
+
+bool HistoryLoggerBackend::publish(const LogRecord& record) noexcept {
+    if (gr::atomic_ref(_busy).exchange(true)) {
```
the-argus, line 51: Is this acceptable? it seems to me that if many threads are logging at once, an undefined number of those logs will actually end up in the history.
(+we don't ever check the return value of this afaict)

core/include/gnuradio-4.0/Logger.hpp
```diff
@@ -1,20 +1,124 @@
...
 namespace gr {
 
+namespace log {
+
+enum class Level : std::uint8_t { fatal, failure, error, warning, info, debug, trace };
+
+inline constexpr std::size_t kLogTextCapacity     = 512UZ;
+inline constexpr std::size_t kLogLocationCapacity = 96UZ;
+
+struct LogRecord {
```
the-argus, line 30: nit: I think `Record` is meaningless, like putting `Data` in a typename. Maybe just `Log`, as it is the data representation of a `gr::log` call? Or `LogEntry`, perhaps. same for `RecordConsumer` below.

If you agree then maybe we could add something to CLAUDE.md about it, as I think claude really likes to use this word.

the-argus (commented):
Thanks for doing this, looks great! I'm glad to see logging maturing and am hopefully about those compile time improvements. I only have a couple comments, I am mostly just wondering about the synchronization strategy of cancelling parallel logs to history.

core/include/gnuradio-4.0/Logger.hpp
RalphSteinhagen: That might cause some name clashes... but I agree the LogRecord is generic but also telling what it es. `LogEntry` would be an alternate name.

**EDIT: `gr::log::Log` looks weird. Will leave it at `LogRecord` for now until we find a better name.**

core/src/Logger.cpp
RalphSteinhagen: Good catch — the non-blocking behaviour is intentional (the history is a bounded, best-effort last-N breadcrumb store, not a lossless audit log; `publish()` must stay wait-free and allocation-free so it is safe from RT/embedded contexts), but you are right that contention drops were silent while capacity overwrites were accounted for.

Addressed in a fixup commit on this branch:
- `publish()` now counts a failed `_busy` acquisition in a `_droppedContended` counter instead of silently returning `false`;
- the next successfully retained record folds `overwritten + contended` into `droppedBefore`, documented as a *lower bound* of records known to have been lost before that entry (the two loss sets are disjoint — stored-then-displaced vs never-stored — so the sum cannot double-count);
- `clear()` resets the counter; its busy no-op is now part of the documented contract (header comment + `docs/USER_API_Logging.md`);
- new deterministic test: a semaphore-parked `snapshot()` consumer holds `_busy`, the concurrent `publish()` is rejected, and the next retained record reports `droppedBefore == 1`.

With the loss accounted inside the backend itself, ignoring the return value in `FixedRecordBackend::publish` becomes legitimate rather than an oversight. A truly lossless multi-producer history (per-slot sequence ring à la Vyukov) would be a redesign — happy to do that as a follow-up if a hard requirement emerges; for now the dominant drop window (a reader holding the backend during `snapshot`/`drain` callbacks) is at least made visible via the lower-bound accounting.

core/include/gnuradio-4.0/Logger.hpp
the-argus: Sounds good.

PR #808, opened 2026-07-10
Fix `kRemoveEdge` creating ghost edges

the-argus:
The edge would still be in the Graph's `_edges`, so it would get reported via `kSchedulerInspect`, but the connection between the actual ports was not there.

Graph must remember to disconnect/connect edges and remove/add them to the `_edges` list at the same time.

### DIFF
--- a/core/test/qa_SchedulerMessages.cpp
+++ b/core/test/qa_SchedulerMessages.cpp
@@ -279,6 +279,46 @@ const boost::ut::suite TopologyGraphTests = [] {
         };
     };
 
+    "Edge removal tests"_test = [&] {
+        gr::Graph testGraph(context->loader);
+
+        // disconnect_on_done=false: dangling blocks would otherwise self-stop and tear down
+        // the freshly emplaced connection via disconnectFromUpStreamParents()
+        auto& blockOut = testGraph.emplaceBlock("gr::testing::Copy<float32>", {{"disconnect_on_done", false}});
+        auto& blockIn  = testGraph.emplaceBlock("gr::testing::Copy<float32>", {{"disconnect_on_done", false}});
+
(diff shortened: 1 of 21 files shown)


core/test/qa_SchedulerMessages.cpp
```diff
@@ -279,6 +279,46 @@ const boost::ut::suite TopologyGraphTests = [] {
...
+
+        // disconnect_on_done=false: dangling blocks would otherwise self-stop and tear down
+        // the freshly emplaced connection via disconnectFromUpStreamParents()
+        auto& blockOut = testGraph.emplaceBlock("gr::testing::Copy<float32>", {{"disconnect_on_done", false}});
+        auto& blockIn  = testGraph.emplaceBlock("gr::testing::Copy<float32>", {{"disconnect_on_done", false}});
+
+        TestScheduler scheduler(std::move(testGraph));
+
+        const property_map edgeData = {{std::pmr::string(gr::serialization_fields::EDGE_SOURCE_BLOCK), std::string(blockOut->uniqueName())}, //
+            {std::pmr::string(gr::serialization_fields::EDGE_SOURCE_PORT), "out"},                                                           //
```
RalphSteinhagen: Is the std::pmr::string cast still needed here? I thought I've removed it for const char literals and std::string_view arguments. 🤔

RalphSteinhagen (approved):

core/test/qa_SchedulerMessages.cpp
the-argus: Oh good point, I just copied that from the surrounding tests. I'll throw in a commit to remove uneeded std::pmr::strings from tests, claude would be good at that.

RalphSteinhagen (approved):

PR #810, opened 2026-07-20
Fix some TSAN output from qa_SchedulerMessages

the-argus:
For https://github.com/fair-acc/opendigitizer/issues/451 I needed to add functionality to the scheduler and in the process I got some ASAN output which looked like it was causing a data race. Switching to TSAN was not very helpful, though, because there was a lot of spam output. So this fixes the data races which were spamming that, so I could better debug things.

Most of these changes are pretty harmless, for example tests were reading from CountingSink's `count` while `work()` was writing to it, for that I created `AtomicCountingSink`. I considered setting up a TSAN blacklist flag for this, but I think I will sleep better at night doing it the defined behavior way. These simpler changes are in the first commit.

The second commit is more complex and contains changes to the SPMC and MPMC wait strategies. Please take time to review these changes as I can only think so hard about all the possible ways ordering can happen :). Both claude and I think it's alright, at least.

This introduces a bit of extra cost (atomic shared pointer operations) but seems like the old code was an actual potential source of crashes, it seems like the old `_cachedSingleReader` pointer might be dereferenced after another thread removed + destroyed a reader. Also, the MPMC was doing the same caching which does not make sense to me since there may be multiple threads reading/writing the cache afaict. But I think removing the cache there should not make things too much slower as there is still the `_cachedMinReaderCursor`. Let me know if I'm missing something there.


blocks/testing/include/gnuradio-4.0/testing/NullSources.hpp
```diff
@@ -224,6 +225,37 @@ Commonly used for testing scenarios and signal termination where output is unnec
     }
 };
 
+GR_REGISTER_BLOCK(gr::testing::AtomicCountingSink, [T], [ uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t, float, double, std::complex<float>, std::complex<double>, std::string, gr::Packet<float>, gr::Packet<double>, gr::Tensor<float>, gr::Tensor<double>, gr::DataSet<float>, gr::DataSet<double> ])
```
drslebedev: The implementation looks correct. 
For maintainability, could both variants share one implementation using an NTTP?

  ```cpp
  template<typename T, bool AtomicCount = false>
  struct CountingSink : Block<CountingSink<T, AtomicCount>> {
      // Select plain or atomic counter access with if constexpr (AtomicCount).
  };

  template<typename T>
  using AtomicCountingSink = CountingSink<T, true>;
  ```

  Both variants could retain distinct registered block names:

  ```cpp
  GR_REGISTER_BLOCK("gr::testing::CountingSink", gr::testing::CountingSink, ([T], false), [ /* supported types */ ])

  GR_REGISTER_BLOCK("gr::testing::AtomicCountingSink",  gr::testing::CountingSink, ([T], true), [ /* supported types */ ])
  ```

This is only an optional maintainability suggestion. If you prefer keeping two separate blocks for clarity, the current approach is also perfectly reasonable.

core/test/qa_SchedulerMessages.cpp
```diff
@@ -665,19 +661,26 @@ const boost::ut::suite MoreTopologyGraphTests = [] {
...
     for (const auto& edge : scheduler.graph().edges()) {
         std::println("edge in list({}): {}", scheduler.graph().edges().size(), edge);
     }
     expect(eq(scheduler.graph().edges().size(), 4UZ)) << "added three new edges, one previously registered with connect";
 
+    scheduler.run();
+    expect(awaitCondition(scheduler, [&scheduler] { return scheduler.state() == lifecycle::State::RUNNING; })) << "scheduler thread up and running w/ timeout";
+    expect(scheduler.state() == lifecycle::State::RUNNING) << "scheduler thread up and running";
+
     // FIXME: edge->connection is not performed
```
drslebedev: Why are these lines commented out? Do we still need this code?

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -760,11 +760,11 @@ struct SchedulerBase : Block<Derived> {
             // Process messages either when the ratio gate opens, or immediately when any entry-point port has
             // pending traffic. This keeps the ratio's amortisation of empty-queue checks while giving arriving
             // messages single-iteration latency (important for multi-hop sub-scheduler message paths).
-            const bool hasMessagesToProcess = msgToCount == 0UZ                //
-                                              || this->msgIn.available() > 0UZ //
+            const bool hasMessagesToProcess = msgToCount == 0UZ                                     //
```
drslebedev: Should `_fromChildMessagePort.available()` also be checked only by worker 0? I t looks to me that it  has the same issue as this->msgIn.available()?


  ```cpp
  const bool hasMessagesToProcess = msgToCount == 0UZ ||
      (runnerID == 0UZ && (this->msgIn.available() > 0UZ ||_fromChildMessagePort.available() > 0UZ));
  ```

drslebedev (commented):
Comment to the second commit (0201478355676444905c939183fdc9c148bba083) 

I think this addresses a real correctness issue.

  For SingleProducerStrategy, removing a reader concurrently with the producer can destroy its Sequence while _cachedSingleReader still points to it. The fix for this lifetime issue looks correctly implemented.

  For MultiProducerStrategy, the cached reader count and raw pointer are read by multiple producers and can be modified by a reader thread during connection or disconnection. Removing this unsynchronized cache therefore makes sense.

  My main concern is the additional cost on a hot path, since the cache was introduced as a performance optimization in #729. Could you add the benchmark results to the PR description so the trade-off is documented?

  The alternative would be to keep the old implementation and guarantee at the scheduler level that readers are only added or removed while all producers are quiescent. That would avoid synchronization inside the buffer, but the invariant would have to hold for every connection and disconnection
  path, including zombie cleanup, which today destroys blocks — and therefore their readers — while sibling blocks are still running. This is a much broader contract and is easier to violate, so I prefer keeping the protection inside the buffer.

  Overall, keeping the SP lifetime fix and removing the unsafe MP cache makes sense to me.

core/test/qa_SchedulerMessages.cpp
the-argus: Oh good point, this test passes now, I just uncommented it. I think I didn't look at it originally and just caught it in a find + replace.

core/include/gnuradio-4.0/Scheduler.hpp
the-argus: Oh yes, thanks for catching that.

blocks/testing/include/gnuradio-4.0/testing/NullSources.hpp
the-argus: Sure, I like this approach, it seems to fit the gnuradio style generally, I'll switch it to be that way

the-argus:
While making the benchmarks (`bm_Buffer`) initially, I discovered that the benchmark crashes before this change, here is a backtrace I get with ASAN enabled, although it can happen in different places:

[crashbefore_relwithdebinfo.txt](https://github.com/user-attachments/files/30597355/crashbefore_relwithdebinfo.txt)

Benchmarks from before applying these changes (it doesn't always crash so I can get some benchmark):
[before.txt](https://github.com/user-attachments/files/30597327/before.txt)

Benchmark from afterwards:
[after.txt](https://github.com/user-attachments/files/30597325/after.txt)

It definitely regresses performance somewhat, but by a fairly consistent margin it looks like. The loss is worse in the areas which were strongest before (1 consumer 1 producer), but at high producer and consumer counts it actually appears to start to perform better. Though perhaps there are some inaccuracies due to UB.

drslebedev (approved):

PR #811, opened 2026-07-21
Add groupBlocks and ungroupBlocks graph API and scheduler messages

the-argus:
Related issue, from opendigitizer: https://github.com/fair-acc/opendigitizer/issues/451

This PR adds two new functions to `gr::Graph`, `groupBlocks` and `ungroupBlocks`. The former will take a list of blocks in the graph, spawn a new subgraph block, then move the blocks in the list to the child graph. The latter will accept a subgraph block and take all of its child blocks, moving them into the parent. Both operations use port exporting/unexporting to maintain the same connections between blocks before and after the operation occurs.

There are a few additions to the SchedulerBase + SchedulerModel beyond the graph implementation and scheduler message property callbacks:
- When doing `disconnectAllEdges()`, the graph now looks at its `_edges` and disconnects the receiving input ports. It used to look at `_blocks`, and disconnect all ports of every block, but this would disconnect connections that the parent graph had to those blocks via exported ports.
- The scheduler now has `_movedBlocks` which is a list of blocks that should not be touched again / removed from the workers' localBlockLists. Workers always check this list before traversing their block list. `_movedBlocks` is behind a mutex, though the mutex will virtually never has anyone else holding it. Let me know if that is a performance concern as I think there are some easy ways to reduce the amount the workers grab that mutex.
- Work quiescence is now doing a lot more to stop workers from doing work, because there were a lot of data races from things like housekeeping resizing port buffers while `connectPendingEdges` is also modifying ports, or zombie cleanup causing ports to be destroyed at the same time as `connectPendingEdges`. Now worker threads basically do nothing but spin during work quiescence.

And some additional fixes w/ tests:
- quickly adding and removing a block would cause it to leak and sit in the zombie list forever
- adopting a singlethreaded scheduler block would cause the worker thread that adopted it to be blocked doing work for that scheduler forever

### DIFF
--- a/core/include/gnuradio-4.0/Graph.hpp
+++ b/core/include/gnuradio-4.0/Graph.hpp
@@ -137,6 +137,20 @@ std::expected<std::shared_ptr<BlockModel>, Error> findBlock(const GraphLike auto
     return std::unexpected(Error(std::format("Block '{}' not found in:\n{}", uniqueBlockName, format(graph)), location));
 }
 
+template<std::convertible_to<std::string_view> SVConvertible>
+std::expected<std::vector<std::shared_ptr<BlockModel>>, Error> findBlocks(const GraphLike auto& graph, const std::vector<SVConvertible>& blockUniqueNames, std::source_location location = std::source_location::current()) noexcept {
+    std::vector<std::shared_ptr<BlockModel>> output;
+    output.reserve(blockUniqueNames.size());
+    for (const auto& uniqueName : blockUniqueNames) {
+        std::expected<std::shared_ptr<BlockModel>, Error> block = graph::findBlock(graph, uniqueName, location);
+        if (!block.has_value()) {
+            return std::unexpected(block.error());
+        }
+        output.emplace_back(std::move(*block));
+    }
+    return output;
+}
+
 std::expected<gr::Edge, Error> findEdge(const GraphLike auto& graph, std::string_view edgeName, std::source_location location = std::source_location::current()) noexcept {
     for (const auto& edge : graph.edges()) {
         if (edge.name() == edgeName) {
@@ -466,6 +480,8 @@ public:
     }
 
     std::pair<std::shared_ptr<BlockModel>, std::shared_ptr<BlockModel>> replaceBlock(std::string_view uniqueName, std::string_view type, const property_map& properties);
+
+    [[nodiscard]] std::expected<std::shared_ptr<BlockModel>, Error> groupBlocks(std::span<const std::shared_ptr<BlockModel>> targetBlocks, std::string_view schedulerTypename, std::source_location location = std::source_location::current());
 
     [[nodiscard]] std::expected<void, Error> emplaceEdge(std::string_view sourceBlock, std::string sourcePort, std::string_view destinationBlock, //
         std::string destinationPort, [[maybe_unused]] const std::size_t minBufferSize, [[maybe_unused]] const std::int32_t weight, std::string_view edgeName) {
--- a/core/include/gnuradio-4.0/Graph.hpp
+++ b/core/include/gnuradio-4.0/Graph.hpp
@@ -483,6 +483,10 @@ public:
 
     [[nodiscard]] std::expected<std::shared_ptr<BlockModel>, Error> groupBlocks(std::span<const std::shared_ptr<BlockModel>> targetBlocks, std::string_view schedulerTypename, std::source_location location = std::source_location::current());
 
+    /// Take blocks in a subgraph and move them into this graph, and delete the subgraph block. Does not touch blocks or ports at all, it only modifies the contents
+    /// of _blocks and _edges of this graph and the subgraph. Some edges are left in a WaitingToBeConnected state.
+    [[nodiscard]] std::expected<void, Error> ungroupBlocks(std::shared_ptr<BlockModel> subgraphBlock, std::source_location location = std::source_location::current());
+
     [[nodiscard]] std::expected<void, Error> emplaceEdge(std::string_view sourceBlock, std::string sourcePort, std::string_view destinationBlock, //
         std::string destinationPort, [[maybe_unused]] const std::size_t minBufferSize, [[maybe_unused]] const std::int32_t weight, std::string_view edgeName) {
         auto sourceBlockIt = std::ranges::find_if(_blocks, [&sourceBlock](const auto& block) { return block->uniqueName() == sourceBlock; });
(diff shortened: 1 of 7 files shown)


core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -1109,6 +1312,174 @@ struct SchedulerBase : Block<Derived> {
...
+        if (category != block::Category::TransparentBlockGroup && category != block::Category::ScheduledBlockGroup) {
+            message.data = std::unexpected(Error{std::format("block '{}' is not a sub-graph", uniqueName)});
+            return message;
+        }
+
+        // if the subgraph is unmanaged, then we were already managing these blocks, this is relatively simple, just change the graph structure
+        if (category == block::Category::TransparentBlockGroup) {
+            {
+                WorkQuiescenceGuard quiescence(this);
+                if (auto result = _graph->ungroupBlocks(std::shared_ptr<BlockModel>(*subGraphBlock)); !result.has_value()) {
```
drslebedev: subGraphBlock was resolved inside targetGraph, but this calls ungroupBlocks() on the root graph.  Should this use targetGraph->ungroupBlocks(...)?

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -579,6 +656,30 @@ struct SchedulerBase : Block<Derived> {
...
+    /*
+     * Removes blocks which may currently be owned by this scheduler. While the
+     * block may not be destroyed immediately, it is guaranteed to never be
+     * dereferenced again by this scheduler after this function returns.
+     *
+     * This only makes sense to call while worker threads are not working, so
+     * one or both of the scheduler being stopped or work quiescence requested.
+     */
+    void removeBlocks(std::span<const std::shared_ptr<BlockModel>> blocksToRemoveSpan) {
+        if (lifecycle::isActive(this->state())) {
```
drslebedev, line 670: Please check if it does the right thing for INITIALISED state (which is not isActive state). The problem is that _executionOrder is already built, and start() will not rebuild it. Clearing it here is therefore unrecoverable and leaves the parent with no work.

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -258,19 +275,31 @@ struct SchedulerBase : Block<Derived> {
...
-    void requestWorkQuiescenceAll() {
+    /// Invokes blockUntilWorking(), do not call when scheduler is not running or being started
+    /// as it will block until it sees a worker thread/call to poolWorker
+    /// isRootScheduler should be true if this is a scheduler which is actually requesting work
+    /// quiescence. It is not correct to blockUntilWorking() in this case because the request is
+    /// issued from the property callbacks, which are issued from poolWorker or an external
+    /// driver thread.
+    void requestWorkQuiescenceAll(bool isRootScheduler = true) {
         requestWorkQuiescence();
+        if (!isRootScheduler) {
```
drslebedev: Should lines 295–297 be removed and the calls below reordered?
 sm->blockUntilWorking(); 
 sm->requestWorkQuiescenceAll(false);

core/include/gnuradio-4.0/Graph.hpp
```diff
@@ -787,24 +807,26 @@ struct Graph : Block<Graph> {
...
-                for (auto& port : ports) {
-                    if (auto* p = std::get_if<gr::DynamicPort>(&port)) {
-                        std::ignore = p->disconnect();
-                    } else {
-                        std::ignore = std::get<BlockModel::NamedPortCollection>(port).disconnect();
-                    }
-                }
-            };
+        for (auto& edge : _edges) {
+            if (edge.state() != Edge::EdgeState::Connected) {
```
drslebedev, line 811: Previously all edges were reset. This new filter leaves failed edges stuck, for example. Could we still reset every edge here?

core/include/gnuradio-4.0/Scheduler.hpp
```diff
@@ -950,6 +1077,21 @@ struct SchedulerBase : Block<Derived> {
 
         auto runnerIndex = std::hash<BlockModel*>{}(newBlock.get()) % nBatches;
         _adoptionBlocks[runnerIndex].push_back(newBlock);
+
+        // start scheduler via the SchedulerWrapper::start() so it spins up a
+        // thread, instead of potentially blocking this scheduler
+        if (auto* schedulerModel = detail::asSchedulerModel(*newBlock)) {
+            assert(newBlock->blockCategory() == ScheduledBlockGroup && "scheduler found with incorrect category");
+
+            if (gr::lifecycle::isActive(newBlock->state())) {
```
drslebedev, line 1088: isActive() also includes REQUESTED_PAUSE and PAUSED, so a new child is started into RUNNING while the parent is paused. Please review if it is acceptable.

drslebedev (approved):
Thanks for the PR and the substantial work on block grouping and ungrouping. 

Overall, the implementation looks good. 
I found several issues and left inline comments. 

Once those are addressed, the PR should be ready to merge.

core/include/gnuradio-4.0/Graph.hpp
the-argus: Good point. Added a test and copied what used to be there.

core/include/gnuradio-4.0/Scheduler.hpp
the-argus: Ah, yes, thank you.

core/include/gnuradio-4.0/Scheduler.hpp
the-argus: Hm, I think things were previously this way, because of this at the start of the block:

```cpp
        if (!lifecycle::isActive(this->state())) {
            return;
        }
```

So a scheduler stops adopted blocks even if it is `PAUSED`. I am not sure if this has consequences, though I think it should probably be fixed since otherwise the `start()` and `pause()` and `resume()` will recursively apply that to children, so that seems like a good invariant to uphold. Do you agree / is this what you are referring to?

Also, now that I am looking closer, this particular return statement shouldn't really matter because `SchedulerModel::start()` will only do something if the block is currently `INITIALISED` anyways- and then `blockUntilWorking()` will do nothing if the scheduler is not working. Although I would like to keep it, I guess hypothetically the user could implement `SchedulerModel` themselves and adopt that, and maybe their implementation does not follow this rule.

core/include/gnuradio-4.0/Scheduler.hpp
the-argus: Ah this is a good point, looking closer at this I think it makes sense to just remove from adoption blocks and execution order when not active.

core/include/gnuradio-4.0/Scheduler.hpp
the-argus: Oh, yes. yep. I removed that `isRootScheduler` thing.

the-argus (commented):
Thanks for the review, especially for catching the targetGraph error and leaving edges in error state.

I left your comment open about the lifecycle state of child schedulers, as it seems like I am at least doing the same behavior that already existed, whether it is exactly correct I am not so sure.

PR #812, opened 2026-07-28
feat(core): load block plugins as shared libs and WASM side modules

dantti:
Add gr::SharedLibrary and an opt-in gnuradio-core-dynload target so hosts can load plugins at runtime without forcing MAIN_MODULE on every Emscripten binary. PluginHandler uses the new loader; WASM plugins build as SIDE_MODULE.

#800


RalphSteinhagen (approved):
@dantti thanks for making the WASM dynamic loading modules/library happen. I guess this leaves the integration into the OpenDigitizer WASM application. Also, there should be a test-case (there not here) to include plugins that are reloaded from the web/file. Thanks again. This looks good and promising.

PR #814, opened 2026-08-06
When reading YAML, try both key and shortKey variants

ivan-cukic:
As keys also have their short form, YAML reading needs to support both the full and short forms.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -264,12 +264,11 @@ jobs:
         run: cmake --build . --target coverage
 
       - name: Run sonar-scanner
-        shell: bash
+        uses: SonarSource/sonarqube-scan-action@v7
         env:
           SONAR_HOST_URL: https://sonarcloud.io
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -267,12 +267,12 @@ jobs:
         run: cmake --build . --target coverage
 
       - name: Run sonar-scanner
-        shell: bash
+        if: matrix.compiler.cc == 'gcc-14' && matrix.cmake-build-type == 'Debug'
```
drslebedev: It looks like this was implemented in the same way as in gr-digitizers, but there the scanner runs inside a matrix job. Here it runs in the standalone coverage job, so the matrix.* values are unset and the step is always skipped. Please remove the if, this job already uses GCC-14 Debug.

drslebedev (approved):
Thanks for the PR!

Looks good overall.

 There is just one small CI issue with the matrix.* condition on the Sonar scanner step, which I noted separately.

 Otherwise, can be merged.

PR #819, opened 2026-08-18
Fix schedulers never starting all work when a pool thread is busy (#813)

### LINKED ISSUE #813: [Bug] Schedulers may never begin all work if any thread in their threadpool is in use elsewhere
In some tests, we take a thread pool, spawn a thread to call `runAndWait()` on a scheduler, and then use that same pool for the scheduler itself / the scheduler's workers. However, our scheduler implementations decide how many lists to put into `_executionOrder` based on the maximum number of threads in the thread pool. So if one thread in the pool is working, and two are available, the scheduler will try to spawn three threads (the max number of threads) and then block.

In the tests where this happens, we do not try to verify that work is being done on all blocks, so it seems to never crop up as a problem.

mohitt31:
### What

The `Simple`, `BreadthFirst` and `DepthFirst` schedulers size their parallel job-lists by `pool.maxThreads()`. Because each `poolWorker` occupies its thread for the scheduler's entire run, this over-subscribes whenever a pool thread is already in use: one job-list is queued and never starts, its blocks never `work()`, back-pressure stalls the rest of the graph, and `runAndWait()` never returns.

This is exactly the situation in #813: a thread from the pool is used to call `runAndWait()`, and the scheduler runs its workers on that same pool — so with an N-thread pool only N-1 threads remain for N job-lists.

### Fix

A small helper `SchedulerBase::availableThreadCount()` returns `max(1, maxThreads() - numTasksRunning())`, and the three schedulers use it in place of `maxThreads()` when choosing how many job-lists to create. Sizing to the threads that are actually free means every job-list is guaranteed a worker.

- It only ever **reduces** the number of job-lists, never increases it, so no worker can be left permanently queued.
- Clamped to **at least one**, so a fully-busy pool still runs the graph on a single job-list rather than doing nothing.
- When the pool is idle (`numTasksRunning() == 0`) the result equals `maxThreads()`, i.e. **no change** to existing behaviour.

### How to test

`core/test/qa_Scheduler.cpp` gains a regression test, run for all three schedulers, that launches a `multiThreaded` scheduler via `executeScheduler` (which occupies one thread of the `[2, 2]`-pinned CPU pool) and asserts the graph finishes and every block did work. On `main` these stall — only the source block runs — and hang until the test's timeout; with the fix they pass.

### Note

Commit fabe56d worked around this by forcing a `singleThreaded` scheduler in `qa_ManagedSubGraph.cpp`, with a TODO to check how many pool threads can be reserved. That workaround can now be reverted; I kept it out of this PR to stick to one concept and am happy to send it as a follow-up.

Closes #813

### DIFF
--- a/core/include/gnuradio-4.0/Scheduler.hpp
+++ b/core/include/gnuradio-4.0/Scheduler.hpp
@@ -735,6 +735,14 @@ protected:
         return {max_work_items, performedWorkAllBlocks, unfinishedBlocksExist ? work::Status::OK : work::Status::DONE};
     }
 
+    // threads currently free in the pool (at least one), used to bound the parallel job-lists so none
+    // is left permanently queued when a pool thread is busy elsewhere (issue #813)
+    [[nodiscard]] std::size_t availableThreadCount() const {
+        const std::size_t poolMaxThreads = static_cast<std::size_t>(_pool->maxThreads());
+        const std::size_t busyThreads    = _pool->numTasksRunning();
+        return poolMaxThreads > busyThreads ? poolMaxThreads - busyThreads : 1UZ;
+    }
+
     void init() {
         [[maybe_unused]] const auto pe = _profilerHandler->startCompleteEvent("scheduler_base.init");
         base_t::processScheduledMessages(); // make sure initial subscriptions are processed
@@ -1923,7 +1931,7 @@ struct Simple : SchedulerBase<Simple<execution, TProfiler>, execution, TProfiler
         case ExecutionPolicy::singleThreaded:
         case ExecutionPolicy::singleThreadedBlocking:
         case ExecutionPolicy::externalStep: break; // single job list; the application drives step()
-        case ExecutionPolicy::multiThreaded: n_batches = std::min(static_cast<std::size_t>(this->_pool->maxThreads()), nBlocks); break;
+        case ExecutionPolicy::multiThreaded: n_batches = std::min(this->availableThreadCount(), nBlocks); break;
         default:;
         }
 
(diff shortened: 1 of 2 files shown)


RalphSteinhagen (approved):
@mohitt31 thanks for your interest and tackling this.

This looks like the right pragmatic fix for #813. Since scheduler parallelism is only a preference, sizing the job-lists to currently available pool capacity is better than letting one long-lived job-list remain queued forever.

I would not call this a full reservation model: availability is sampled during init/reset, so callers should still avoid driving a scheduler from the same fixed-size worker pool for the graph lifetime unless capacity is reserved or the pool is sized accordingly.
For the reported runAndWait()/executeScheduler path, though, this addresses the actual liveness failure.

Before merging, could you please trim the long explanatory code comment in Scheduler.hpp and keep the detailed rationale in the PR/issue? Also, in the regression test, please consume and check `future.get()` so we assert that `runAndWait()` returned success, not only that the future became ready.

Functionally this is a go from me after those small cleanups.

mohitt31:
Thanks @RalphSteinhagen! Addressed both: trimmed the comment in `Scheduler.hpp` to two lines (rationale kept in the PR and #813), and the regression test now consumes `future.get()` and asserts `runAndWait()` returned success for all three schedulers.

mohitt31:
The two red checks look unrelated to this change:

- **GCC-14 Coverage** — all tests pass; the job only fails at the final SonarCloud step (`SONAR_TOKEN ... Not authorized`), which isn't available to fork PRs.
- **AdaptiveCpp | Debug** — the sole failing test is `qa_SubGraphAssets` (120 s timeout). It uses the `singleThreaded`/`singleThreadedBlocking` policies, which this PR doesn't touch (the change only affects the `multiThreaded` batching path), and it passes on every other config including AdaptiveCpp | Release. `qa_Scheduler` itself passes on all configs.

Happy to rebase for a clean re-run if that's useful.

RalphSteinhagen (approved):
@mohitt31 thanks for the adjustments. I'll squash-merge this as soon as the relevant CI-items passed.

_N.B. Some of the failing CI steps are flukes... the other require private tokens/authorisation. Don't worry._

PR #820, opened 2026-08-19
feat: self-contained gzip/DEFLATE codec and gzip file-io integration

RalphSteinhagen:
Adds a dependency-free gzip/DEFLATE codec to `core` and wires it into the file-io layer, so `.gz` sources — local files, downloads, in-memory URIs and HTTP — are transparently inflated, and compression can be requested explicitly on write.

Two commits, reviewable independently:

1. `feat(core): add gzip/DEFLATE codec` — the codec itself (`core/include/gnuradio-4.0/Compression.hpp`)
2. `feat(algorithm): add gzip file-io integration` — `FileIo`, `BasicFileIo`, `HttpBlock` integration

## Codec

Self-contained C++23 text-book/RFC-based implementation of raw DEFLATE, zlib and gzip (RFC 1950/1951/1952) with **no external compression dependency**. Encoder and decoder share one `constexpr` implementation, so a compile-time resource and a run-time call exercise identical code.

### Compile-time resources

Built during constant evaluation and proven to decode back to their source by a `static_assert` at the point of definition — an embedded blob cannot silently be corrupt. `CompressedText` and `CompressedBlob` decode lazily into persistent storage, and compressed text plugs into the existing formatter integration:

```cpp
using namespace gr::compression::literals;

static constexpr auto help    = "embedded help text"_gzip;
static constexpr auto compact = gr::compression::makeCompressedText<
    "embedded help text", gr::compression::CompressionLevel::best>();
const auto rendered = std::format("{}", help);
```

Binary resources take the same route:

```cpp
inline constexpr std::array source{std::byte{0x01}, std::byte{0x02}};
inline constexpr auto packed = gr::compression::gzip<
    source, gr::compression::CompressionLevel::best>();
```

### Run-time buffers

The allocating decode paths take a decompressed-size cap, so a hostile stream cannot become an unbounded allocation; `decompressedSize` answers the same question without writing anything. Decoding optionally targets a `std::pmr` resource:

```cpp
const auto packed = gr::compression::compress(payload, gr::compression::Format::gzip);
const auto plain  = gr::compression::decompress(*packed, gr::compression::Format::gzip);
const auto pooled = gr::compression::decompress(*packed, gr::compression::Format::gzip, &arena);
```

### Streaming

`StreamCompressor` / `StreamDecompressor` handle chunked input and output. `StreamResult::status` reports whether the codec wants more input, more output space, or is done, so callers need not infer progress from byte counts. The decoder rewinds to a symbol boundary when it runs out of bits, so the unconsumed tail of a chunk has to be offered again:

```cpp
gr::compression::StreamDecompressor inflate{gr::compression::Format::gzip};
while (!inflate.finished()) {
    const auto state = inflate.process(pending, output, atEnd ? Flush::finish : Flush::none);
    pending = pending.subspan(state->consumed); // keep what was not consumed
    consume(output.first(state->produced));
    if (state->status == gr::compression::StreamStatus::needInput) {
        pending = refill(pending);
    }
}
```

The encoder picks the smallest of stored, fixed-Huffman and dynamic-Huffman per block and carries its LZ77 window across block boundaries, so a small block size bounds latency without wrecking the ratio. `fast`, `balanced` (default) and `best` tune bounded LZ77 search effort and lazy matching.

## File-io integration

`CompressionMode { automatic, none, gzip }` on the file-io config, surfaced as a reflected `compression` setting on `BasicFileSource`, `BasicFileSink`, `HttpSource` and `HttpSink`, plus `compression_level` on the sinks and `max_decompressed_bytes` on the sources.

**Read and write interpret `automatic` differently, on purpose:**

| | `automatic` | `none` | `gzip` |
| --- | --- | --- | --- |
| read | inflate if the URI ends in `.gz` | raw bytes | force inflate |
| write | store verbatim | store verbatim | compress |

A `.gz` source is inflated because that hands the caller the bytes it asked for. Writing stores the payload verbatim even when the name ends in `.gz`, because a file name must not silently change what lands on disk — gzip output requires asking for `CompressionMode::gzip`. This also keeps formats where compression is unusual, such as `.wav`, from acquiring it by accident.

Also in this commit: reader offsets apply to the *decoded* bytes, gzip expansion is bounded during both preflight and streaming, and empty writes remain empty touch/truncate operations.


RalphSteinhagen:
Need to push again. The older Clang has a different stack limit.

RalphSteinhagen:
Needed to force-push again to fix some GCC14 specific warnings and WASM unit-test execution (no `gzip` execution in node/browser).

drslebedev (approved):
Overall, the PR looks good. 

I added three small fix commits for issues we discussed privately.

One remaining possible issue  is HTTP gzip content coding: a .gz response served with `Content-Encoding: gzip` may already be decoded by CPR or the browser before reaching `FileIO`, causing a second decode. Please think if we need to  add a regression test covering both content-coded responses and raw gzip responses without `Content-Encoding`.

```cpp
"FileIO - Native HTTP content-coded gzip is not decoded twice"_test = [&] {
      const std::string plain      = "http-content-coding:" + createTestString();
      const auto        compressed = gr::compression::compress(std::as_bytes(std::span<const char>(plain.data(), plain.size())), gr::compression::Format::gzip);
      expect(compressed.has_value());
      if (!compressed) {
          return;
      }

      const auto gzip = toUint8Bytes(*compressed);

      httplib::Server server;
      server.Get("/content-coded.gz", [&](const httplib::Request&, httplib::Response& res) {
          res.set_header("Content-Encoding", "gzip");
          res.set_content(reinterpret_cast<const char*>(gzip.data()), gzip.size(), "application/octet-stream");
      });

      auto serverThread = std::thread{[&server] { server.listen("localhost", 8080); }};
      server.wait_until_ready();

      auto readerExp = fileio::readAsync("http://localhost:8080/content-coded.gz", {});

  #if GR_HTTP_ENABLED
      expect(readerExp.has_value());
      if (readerExp) {
          auto reader  = std::move(readerExp.value());
          auto results = getReadResult(reader);
          expect(eq(results.errorCounter, 0uz));
          expect(eq(plain, joinBytesToString(results.allData)));
      }
  #else
      expect(!readerExp.has_value());
  #endif

      server.stop();
      serverThread.join();
  };
```
This test fails now for me. We can discuss if it is a valid case for us and need to be fixed.


  Apart from that, I found no further blockers.

PR #821, opened 2026-08-20
Fix sample rate forwarding for resampling blocks

drslebedev:
Keep sample_rate as the input rate in block settings. Apply the resampling ratio only when publishing output tags.

This prevents decimating blocks from storing and saving their output rate, which caused the rate to decrease again after each iteration.

Tests: Cover decimation, interpolation, blocks without a sample_rate setting, and GRC save/load behavior.

### DIFF
--- a/core/include/gnuradio-4.0/Block.hpp
+++ b/core/include/gnuradio-4.0/Block.hpp
@@ -1085,6 +1085,29 @@ public:
         }
     }
 
+    void insertOutputValue(property_map& destination, const auto& wireKey, const auto& value) const {
+        const std::string_view fieldKey          = gr::tag::settingsKey(std::string_view{wireKey});
+        const bool             convertSampleRate = ResamplingControl::kEnabled && input_chunk_size != output_chunk_size && fieldKey == gr::tag::SAMPLE_RATE.shortKey();
+        if (convertSampleRate && input_chunk_size != 0U) {
+            if (const float* inputRate = value.template get_if<float>()) {
+                const float ratio = static_cast<float>(output_chunk_size) / static_cast<float>(input_chunk_size);
+                destination.insert_or_assign(wireKey, ratio * (*inputRate));
+                return;
+            }
+        }
+        destination.insert_or_assign(wireKey, value);
+    }
+
+    [[nodiscard]] property_map toOutputTags(const property_map& parameters) const {
+        auto toWireKey = [](std::string_view bareKey) -> std::string { return std::ranges::contains(gr::tag::kDefaultTags, bareKey) ? std::string(gr::GR_TAG_PREFIX.view()) + std::string(bareKey) : std::string(bareKey); };
+
+        property_map wireTags;
+        for (const auto& [key, value] : parameters) {
+            insertOutputValue(wireTags, toWireKey(std::string_view{key}), value);
+        }
+        return wireTags;
+    }
+
     /// default tag forwarding — called by workInternal unless the user provides forwardTags()
     template<typename TInputSpans, typename TOutputSpans>
     void forwardInputTags(TInputSpans& inputSpans, TOutputSpans& outputSpans, std::size_t processedIn) noexcept {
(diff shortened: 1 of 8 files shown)


core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1085,6 +1085,37 @@ class Block : public lifecycle::StateMachine<Derived>, public BlockBase {
         }
     }
 
+    /// A resampling block's settings hold its input rate; only its output stream carries the resampled rate.
```
RalphSteinhagen: We should avoid this redundant in-code documentation in favour of the 'nomen-est-omen' policy, i.e., naming methods/functions and parameters so the function signature and expected behaviour are self-explanatory. This makes the documentation more durable w.r.t. refactoring and avoids bloat ... although -- specifically here -- it's not much but other developers (and even more LLM) will take this as a code-reference w.r.t. style.

Only, if you touch this code again: see if this could be eliminated.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1118,11 +1149,17 @@ class Block : public lifecycle::StateMachine<Derived>, public BlockBase {
                     if (cachedSettings == nullptr) {
                         cachedSettings = &settings().activeParameters();
                     }
-                    // only substitute when the active setting differs from the tag value; an equal
-                    // value (the steady-state constant-tag case) needs no rebuild → verbatim pass-through
-                    if (auto it = cachedSettings->find(fieldKey); it != cachedSettings->end() && !((*it).second == kv.second)) {
+                    if (shouldConvertOutputSampleRate(fieldKey)) {
                         anySubstitute = true;
+                    } else {
+                        // only substitute when the active setting differs from the tag value; an equal
```
RalphSteinhagen: same here.

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1085,6 +1085,37 @@ class Block : public lifecycle::StateMachine<Derived>, public BlockBase {
...
+        } else {
+            return false;
+        }
+    }
+
+    void insertOutputValue(property_map& destination, const auto& wireKey, const auto& value) const {
+        const std::string_view fieldKey = gr::tag::settingsKey(std::string_view{wireKey});
+        if (shouldConvertOutputSampleRate(fieldKey)) {
+            if (const float* inputRate = value.template get_if<float>()) {
+                const float ratio = static_cast<float>(output_chunk_size) / static_cast<float>(input_chunk_size);
```
RalphSteinhagen, line 1093: We should guard against division by zero and at least warn here if this happens (by accident).

core/include/gnuradio-4.0/Block.hpp
```diff
@@ -1085,6 +1085,37 @@ class Block : public lifecycle::StateMachine<Derived>, public BlockBase {
...
+                destination.insert_or_assign(wireKey, ratio * (*inputRate));
+                return;
+            }
+        }
+        destination.insert_or_assign(wireKey, value);
+    }
+
+    [[nodiscard]] property_map toOutputTags(const property_map& parameters) const {
```
RalphSteinhagen, line 1101: Could this function be inlined as a lambda? Otherwise we create a larger than necessary API surface.

core/include/gnuradio-4.0/Block.hpp
RalphSteinhagen: Maybe make this also a inline lambda?!?

RalphSteinhagen (approved):
Just minor comments. If you could re-review, adjust ... could be merged afterwards once the CI passes. Thanks for the new unit-test catching this specific bug!

core/include/gnuradio-4.0/Block.hpp
drslebedev: I removed shouldConvertOutputSampleRate and make inline const bool.  
I would keep toOutputTags() for now. Its callers are about 1,000 lines apart, and the logic is too large to copy and maintain twice.

PR #822, opened 2026-08-28
Feature: ONNX integration

### LINKED ISSUE #629: [3SP, 20SP] ONNX PoC: Integrate Basic Inference Block for Signal Classification and Peak Detection
## ONNX PoC:<br>   Minimal Inference Block for Time-Series Classification (e.g. Peak Detection, Modulation Labelling)

This task explores the integration of ONNX Runtime into GNU Radio 4.0 by implementing a minimal inference block capable of applying pretrained models to real-time or buffered time-series input.

### Scope
 * Create a basic ONNX inference block that accepts 1D float input vectors (or `std::complex<float>` magnitudes).
 * Load and evaluate a fixed ONNX model at runtime (e.g. MLP or CNN).
 * Apply this to minimal signal processing MVPs such as:
   - Multi-peak detection in noisy FFT spectra
   - Modulation classification
   - Anomaly detection / signal presence
 * Use ONNX Runtime with CPU backend only (no GPU offload or EP integration).
 * Accept static shape models (initially), with input validation logic to match.
 * Optional: visualise output via debug log or TagEmitter.

### Design Notes
 * Avoid complex tokenisation or postprocessing — focus on direct tensor-in, tensor-out logic.
 * Reuse existing GR4 block model infrastructure.
 * Support importing `.onnx` model path as a parameter.
 * Add a `onnx::Block` namespace for future extension (e.g. quantised models, streaming RNNs).
 * Potential model sources:
   - [Whisper-tiny] (as a stretch goal or toy speech demo)
   - [Coqui-TTS] (for future inverse/response model)
   - Small 1D CNN/MLP trained in scikit-learn, PyTorch, or Keras and exported to ONNX.

### Rationale

This PoC is meant to establish a minimal, testable bridge between GNU Radio and ONNX Runtime, enabling future DSP+ML hybrid pipelines. Rather than integrating large and complex models up front, the focus is on:
 - Simplicity of integration
 - Practical MVPs in the RF/timeseries domain
 - Documenting and resolving typical integration pitfalls (e.g. shape inference, memory reuse, ONNX session lifecycle)

Large models such as Whisper or TTS are **not** the primary goal here — they serve as long-term examples, not initial targets.

RalphSteinhagen:
Adds [ONNX](https://onnx.ai/) runtime as an optional dependency and a generic inference block on top of it. The
governing constraint is that the whole algorithm lives in the `.onnx` file — normalisation, any
post-processing, and even iteration are in-graph, so there is no required host step and swapping
the model swaps the topology. The C++ resamples, runs the session, and hands back tensors.

```cpp
gr::Graph graph;
auto&     inference = graph.emplaceBlock<gr::blocks::onnx::OnnxInference<float>>({
    {"model_path", "models/frame_delta_N64_M8.ort.gz"}, // .gz inflated transparently
    {"history_depth", 8U},                              // M frames per inference
    {"normalise_mode", "LogMAD"},
});
```

## What can be run

`OnnxInference` covers the four rate configurations a GR4 block can have, selected by the model's
own shape rather than by a setting: 1:1 elementwise, N:1 decimating, an M x N sliding window, and
multi-output.

A model carries its own integration contract in ONNX metadata: input size, history depth, the
normalisation it expects, and — where it emits structured results — the column layout of its
output tensor. The block reads that at load and configures itself. An explicit setting always
wins, so metadata only fills in what the caller left at its default; otherwise loading a
re-exported model would silently retarget someone who had chosen deliberately.

## Supported model structures

- **Input rank 1, 2 or 3** — `[N]`, `[batch, N]`, `[batch, frames, N]`. The batch dimension may be
  symbolic, so fixed- and dynamic-batch exports both load unmodified.
- **Single-frame or temporal.** `history_depth = 1` feeds one spectrum per inference;
  `M > 1` requires a rank-3 input and is rejected at load otherwise, with the declared shape named
  in the message.
- **`history_stride = 1`** slides 1:1 with a warm-up pass-through until the window fills;
  **`S > 1`** decimates S:1, one output per S inputs.
- **Single- or multi-output.** Extra outputs surface on `meta_out`; symbolic output dims are
  pinned by one zero-filled probe inference at load.
- **`DataSet<T>` or `Tensor<T>`**, independently for input and output — `DataSet` carries axes and
  metadata with the samples, `Tensor` is bare values.
- **`T` = `float` or `double`**; ORT tensors are float32, so a double block narrows at the session
  boundary and widens on return.
- **Normalisation in-graph or host-side** — `None`, `LogMAD`, `MinMax`, `ZScore`, or an ExprTk
  `Expression` over the resampled frame, with `n`/`median`/`mad`/`min_val`/`max_val`/`mean_val`/
  `std_val` pre-computed. A model that normalises internally declares `InGraph` and the host stays
  out of the way.
- **Input resampling** — `Linear` to the model's N, or `None` to require an exact match.
- **Overridable initializers** become `config_in` ports and `model_overrides` entries, so a baked
  operating point is tunable without re-export.
- **Structured outputs** may declare a column layout in metadata; the host validates the declared
  prefix at load rather than misparsing it silently.

Models should be exported as a `.ort` flatbuffer to be deployable for the WASM and on TPU-class targets where `.onnx` often does not work (e.g. due to the lack of floating-point ALUs). This is not important for native/x86/Linux targets that work fine with `.onnx`. Sizes read from model metadata are bounded before they reach an allocation, and a malformed or hostile file is rejected at load with an actionable message rather than throwing out of an exception-free block.

Models ship gzipped; the file-io layer inflates `.gz` transparently, so a path is a path.

## Demo: RF/Schottky Peak detection

`OnnxPeakDetector` is the non-trivial consumer: a learned spectrum peak detector publishing the same `timing_events` key set as the existing classical `PeakDetector`, so it can replace it in a graph without downstream changes. Its cascade is a rolled ONNX `Loop` — ten stages, each subtracting the peak it promotes — not a host-side loop.

image

Widths are published as `fwhm`, `hwhm_l`, `hwhm_r` rather than a "sigma", because only one of the seven generated shapes has one: a Lorentzian has a gamma, a sinc-squared and a parabola have neither. `w68`/`w96`/`w99` are energy-containment widths of the profile the model itself fitted, integrated in-graph against an analytic total. 
A fixed multiple of sigma would be wrong in both directions — the true `w99` runs from 4.20 sigma at eta=0 to 10.27 sigma at eta=1, against the 6 sigma a constant assumes.

There is a new UTF-8-based spectrum ImChart display that can be used for developing and debugging in the console only:

```text
spectrum: 1024 bins, 4 reference peaks, fixed seed

   +5.5865 ┤single-stage [a.u.]                                                                 
           │                                                                                    
           │           O                                                                        
           │           ★                                                                        
           │           ⠠                                                                        
           │                                                                                    
           │           ⠂                                                                        
           │           ⠐                                                                        
           │                                                                                    
  +3.57845 ┤           ⠂                                                                        
           │           ⠠                                                                        
           │                                 ★                                                  
           │           ⡀                     ⡞⡀                                                 
           │                                 ⡂⡁                                                 
           │           ⠐                     ⠂⡂                                                 
           │                                ⢈ ⠂                                                 
           │          ⠈                     ⠠ ⢈                                                 
           │           ⠐                    ⢈ ⠠                                                 
  +1.57039 ┤                                ⠠ ⢐                     ★⡀                          
           │          ⠈                     ⡐ ⠠                    ⢰⠃⢱⡀                         
           │           ⠈                    ⠂  ⡂                   ⡇  ⡇                         
           │          ⠐                     ⠅  ⡂                  ⢸   ⢸          ⢘★             
           │            ⠁                   ⠅  ⡂                  ⡇    ⣇         ⢐⠆             
           │          ⠈ ⠄                  ⢘   ⠰                 ⡸     ⠸⡀        ⠅⠰             
           │          ⢈ ⠄                  ⡸   ⠸⡀               ⣰⠃      ⢣        ⠅⢘             
           ├────────────────────┬────────────────────┬────────────────────┬────────────────────┐
           │                   256                  512                  768                1024
  -0.43766 ┘ █: spectrum █: reference █: detected            [..]/ex05_ml_peak_detector.cpp:130]
  single-stage — expected vs measured:
       pos     pos'    Δpos |     amp    amp'   Δamp% |   fwhm   hwhmL'/R'  Δfwhm% | stage
     150.3   151.18   +0.88 |    5.00    5.02   +0.3% |    9.4   5.8/4.0     +3.8% |     0
     420.7   421.32   +0.62 |    3.00    3.01   +0.2% |   28.3  15.0/13.9    +2.2% |     0
     700.0   701.43   +1.43 |    1.50    1.57   +4.7% |   58.9  32.9/30.4    +7.7% |     0
     860.5   861.10   +0.60 |    1.00    1.03   +2.8% |   14.1   8.8/7.1    +12.0% |     0
  summary: 4/4 reference peaks detected, 0 spurious
```

The ML detector recovers substantially more of the ground truth; the classical one is better on parameter accuracy and emits far fewer duplicates. Neither dominates.

Two caveats on those numbers. `SyntheticPeakSpectrum` matches the shape repertoire the ML model was trained on and its ground-truth definitions are the training targets, so this is in-distribution for the ML side and is not an out-of-distribution real-data measurement.

Comparing the classic- vs ML(ONNX)-based peak detectors figure of merits:
```text
=== ML vs classical peak detection ===
2000 independent spectra, scored per spectrum

max peaks: 10 (compiled-in default) (both detectors)
ML model: /home/rstein/git/graph-prototype/blocks/onnx/models/ex05_peak_detector_cascaded.onnx.gz (compiled-in default)
ML gate:  0.45 (compiled-in default)
Peak detection: classical vs ML   (2000 spectra, seed=42, single-frame scoring)

Detection quality
  metric                                   classical          ML      better
  ground-truth peaks                            7852        7852           -
  total detections                              4803        8839           -
  matched peaks                                 4387        5815          ML
  false negatives (missed)                     44.1%       25.9%          ML
  fabrications (no true peak)                   4.4%       13.2%   classical
  duplicates (extra hit on matched)             4.3%       21.1%   classical
  position MAE [bins]                           0.94        8.86   classical
  width MAE [bins]                              3.28       15.54   classical
  amplitude error [relative]                   0.174       0.418   classical

Parameter accuracy by peak width   (classical / ML)
   fwhm band               n        pos [bins]        pos [fwhm]       fwhm_est/gt    amp [relative]
         0-3     2298 / 2625       0.09 / 0.45       0.04 / 0.20       1.33 / 1.56       0.16 / 0.21
        3-10     1121 / 1388       0.22 / 0.40       0.04 / 0.08       1.23 / 1.37       0.16 / 0.23
       10-40       593 / 839       1.52 / 3.70       0.07 / 0.16       1.02 / 1.15       0.20 / 0.55
      40-120       288 / 526      5.04 / 19.63       0.08 / 0.27       0.81 / 0.78       0.21 / 0.84
        >120        87 / 437     14.87 / 83.26       0.08 / 0.32       0.77 / 0.51       0.60 / 1.47

  width ratio targets 1.00; rates are per detection except false negatives, which are per ground-truth peak
```

## A note on timing

Session creation dominates a one-shot run, and inference dominates a batch run; both stem from model size rather than the integration.

| | session create | per spectrum |
| --- | --- | --- |
| classical `PeakDetector` | - | <= 46 ms (measured together with spectrum generation) |
| ML single-stage | ~11 s | 39 ms |
| ML cascade | ~13 s | 843 ms |

Measured on a loaded developer machine in a Debug build, so treat the absolutes as an upper
bound; the ORT figures come from a release runtime and are unaffected by the build type.

For interactive use, load once and keep the session: the cost is per session, not per inference.


drslebedev:
The PR fully compiles and runs successfully on my system. Only one test in qa_OnnxPeakDetector fails:

```text
Running test "int8 tracks fp32's detections, is far smaller, and costs per-inference time to get there"... FAILED
in: /home/slebedev/Dev/gnuradio4/blocks/onnx/test/qa_OnnxPeakDetector.cpp:939 - test condition:  [false]
int8 deliverable failed to load: /home/slebedev/Dev/gnuradio4/blocks/onnx/models/ex05_peak_detector_single_stage_int8.ort.gz
```

  The executable also prints hundreds of duplicate ONNX schema-registration errors at startup, for example:

```text

Schema error: Trying to register schema with name Sin (domain:  version: 22) from file ./onnx/defs/math/defs.cc line 1403, but it is already registered from file ./onnx/defs/math/defs.cc line 1403

Schema error: Trying to register schema with name Cos (domain:  version: 22) from file ./onnx/defs/math/defs.cc line 1426, but it is already registered from file ./onnx/defs/math/defs.cc line 1426

```

The test currently reports only isModelLoaded() == false, so the actual ONNX Runtime load error is hidden.

Could you please have a look?

algorithm/include/gnuradio-4.0/algorithm/fileio/FileIo.hpp
drslebedev, line 437: `firstChunkSize` should be probably also limited by `gr::pmt::kMaxTensorElements`?  you did it below for the case when pendingBytes are empty.

blocks/onnx/include/gnuradio-4.0/onnx/OnnxPeakDetector.hpp
```diff
@@ -0,0 +1,371 @@
...
+        output.signal_values.resize(4 * inputSize);
+        auto dest = output.signal_values.begin();
+        std::copy_n(inData.signal_values.begin(), inputSize, dest);
+        dest = std::copy(heatmapOut.begin(), heatmapOut.end(), dest + static_cast<std::ptrdiff_t>(inputSize));
+        dest = std::copy(reconOut.begin(), reconOut.end(), dest);
+        std::copy(residOut.begin(), residOut.end(), dest);
+
+        output.meta_information = {{}, {}, {}, {}};
+
+        const float posScale = (inputSize != modelN) ? static_cast<float>(inputSize) / static_cast<float>(modelN) : 1.f;
```
drslebedev: It looks like a mismatch in scaling formula with OnnxPreprocess.

Should this be ` (inputSize - 1) / (modelN - 1)`?

drslebedev (approved):
Thanks for the PR! 

I compiled and tested it locally, and everything worked correctly except for one test failure, which I described in a separate comment.

  My overall impression is very positive. The implementation is clearly structured and well defined, and I was able to follow the overall design and logic.

  I especially appreciate the extensive unit-test coverage and the large number of examples. They make up a significant part of the PR and are very useful for understanding and validating the implementation. I think the examples will also provide a good starting point for users who want to work with the new functionality.

  I left 2 additional comments that are worth taking a look at. 

Apart from those points and the separately reported test failure, I think the PR is ready to be merged.

RalphSteinhagen:
**Force-pushed**: fixes were amended into the commits that introduced them rather than added as fixups. Scope is unchanged, apart from one new final commit (`basic:`).

Most changes address CI performance on 4-core / 16 GB runners:

* **Build time:** Release builds were effectively serial with Makefiles and hit the 6 h timeout. CI now uses Ninja, enabling the existing `GR_BUILD_PARALLEL_LEVEL` job pools.
* **Memory:** `ConverterBlocks` previously generated one ~7.2 GB translation unit. `GrBasicBlocks` now emits one unit per registration (~1.3 GB each). Parallelism is limited to GCC 3, clang/AdaptiveCpp 4, and emcc 2 jobs based on measured peak RSS.
* **Local/CI differences:** `ModelPath.hpp` now gets the models directory from CMake rather than deriving it from `__FILE__`. The FetchContent cache key now includes the generator to avoid Makefiles/Ninja conflicts.
* **WASM:** Added missing `--embed-file` fixtures. Rank and multi-format tests use `.ort` under Emscripten because the minimal ORT build cannot load `.onnx`. `emcc | Release` now passes 83/83 tests.

Also addressed the two review comments (`FileIo` chunk clamp, `posScale`), dropped `int8`, and added retries for flaky tests via `ctest --rerun-failed --repeat until-pass:3`.

drslebedev:
Thanks for the update. The changes look good to me. I’ll merge the PR.

PR #823, opened 2026-08-28
Fix scheduler hanging when requesting work quiesence

the-argus:
Using `_nRunningJobs` to check if the scheduler had begun starting workers doesn't make sense if there is any scheduler which has no blocks and therefore no workers.

I wanted to avoid this because it creates a kind of extra possible state that the scheduler can be in, but I think something like it is unavoidable as we need to have guarantees whether a given scheduler is currently having its graph accessed by another thread.

Partially fixes an issue found in review of https://github.com/fair-acc/opendigitizer/pull/477


RalphSteinhagen (approved):
@the-argus Thanks for the fixes!

RalphSteinhagen:
@all-contributors please add @the-argus for code, bug, doc

PR #826, opened 2026-09-03
build: cut generated block-library compile time and peak memory

RalphSteinhagen:
The build commits attack redundant work in the generated block-registration translation units in sequence — each targets a distinct, additive cost, verified before-vs-after test-builds. Added compatibility with GCC16, Clang22 and libstdc++ 16 as a compatibility prerequisite (libstdc++ 16 makes `std::expected` `[[nodiscard]]`).

## Incremental changes

| Commit                                                         | Change                                                                                                                             | Measured on                                     | Before → after                                                                                 |
| -------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------- | ---------------------------------------------------------------------------------------------- |
| `compile with GCC 16, Clang 22 and libstdc++ 16`               | Fixes ~40 newly-`[[nodiscard]]` call sites; RTL2832Source now forwards the tuner's *achieved* values instead of the requested ones | —                                               | prerequisite, no build-cost delta                                                              |
| `stop scanning for C++20 modules`                              | `CMAKE_CXX_SCAN_FOR_MODULES` was only ever turned off in the Clang branch; the tree declares no module                             | GrBasicBlocks, 334 units, -j6                   | GCC: −23 s · Clang: −6 s                                                                       |
| `precompile the registry header for generated block libraries` | `target_precompile_headers` on `BlockRegistry.hpp`, which cost 7.2 s/unit before any block code                                    | GrBasicBlocksObject, 334 units, -j6, ccache off | GCC 16: 1377 s → 1095 s (1.29 → 1.19 GB/unit) · Clang 20: 1094 s → 660 s (0.80 → 0.50 GB/unit) |
| `batch block instantiations per generated translation unit`    | Groups instantiations so the per-unit fixed cost is paid 61× instead of 334×; default batch size 8                                 | GrBasicBlocksObject, GCC 16, -j6, ccache off    | 1095 s (334 units) → 373 s (61 units) · Clang 20 at batch 8: 660 s → 237 s                     |
| `compile each block instantiation once, not once per registry` | Plugin builds no longer recompile the same bodies the shared/static build already compiled — only a small forward-declared hook    | GrBasicBlocks plugin object, GCC 15, -j6        | 61 units / 373 s → 60 hooks / 10 s                                                             |

## Overall effect

Whole-project, virgin trees, `-j6`, ccache disabled:

| Metric                 | GCC 15              | Clang 20            |
| ---------------------- | ------------------- | ------------------- |
| Summed compile time    | 9.26 h → **3.85 h** | 7.01 h → **2.87 h** |
| Max single-TU peak RSS | 7.34 GB → 6.67 GB   | 4.31 GB → 4.34 GB   |
| Ninja edges            | 3111 → 842          | 3109 → 836          |

Wall-clock isn't directly comparable to a pre-branch baseline (before `-j3` vs after '-j6'). Peak memory is unchanged but is now a trade-off using the new batch parameter (`qa_Math.cpp` is a good litmus test for this).

## CLAUDE.md updates

`docs: make CLAUDE.md the single source of truth for agent guidelines` — restructures the project's `CLAUDE.md` (470 → 210 lines) and adds a working-agreement section it never had; unrelated to the build-cost work above. Reduces the minimum required number of tokens.

_N.B. further improvement options would require code-changes e.g. batch-size-per-registration or refactoring common strate from `Block<T>` into a common pod struct._


drslebedev (approved):
Thanks for the PR!

The changes look good and results are very promising.
I also tested it locally with -j46: I got like 10% improvements in time, but the peak memory consumption also drops down significantly.

The PR can be merged.

RalphSteinhagen:
:grinning: 

image

PR #827, opened 2026-09-03
ci: save ccache after Build, not only on full job success

RalphSteinhagen:
actions/cache@v5's bundled save has `post-if: success()` baked into the action, so ccache is only persisted when the whole job — ctest included — passes. Verified against #826: several full CI cycles produced zero saved ccache entries for that branch.

Splits `Cache ccache` into `actions/cache/restore@v5` (unchanged, top of job) and `actions/cache/save@v5` placed right after `Build`, gated on `!cancelled()` instead of `success()`. Applies to the compiler-matrix `build` job and the `coverage` job.

- A ctest (or later-step) failure no longer discards a successful compile's ccache.
- ccache records each translation unit as it compiles, not only at the end of `cmake --build`, so a build that fails partway through still keeps what did compile.

Does not help a run killed by this repo's `cancel-in-progress: true` — a cancelled runner executes no further steps at all.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,8 +102,8 @@ jobs:
           fetch-depth: 0
           lfs: true
 
-      - name: Cache ccache
-        uses: actions/cache@v5
+      - name: Restore ccache
+        uses: actions/cache/restore@v5
         with:
           path: ${{ github.workspace }}/.ccache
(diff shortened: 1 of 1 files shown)


drslebedev (approved):

RalphSteinhagen:
_N.B. consistent numbers, all compile targets and c-test pass._

image

PR #828, opened 2026-09-04
fix(PythonBlock): correctness, tag handling and script loading

RalphSteinhagen:
`qa_PythonBlock` aborted with `free(): invalid size`. Fixing that surfaced other defects in the C++/Python boundary that I felt needed fixing.

Suggested reviewing commit by commit; each is self-contained, builds with `-Werror` and
passes on its own. The commit messages carry the detail.

Two behaviours worth knowing before:
- The GIL was never released after `Py_Initialize()`, fine for the unit-test but not on a multi-threaded scheduled graph.
- `Block<>` already forwards `gr:`-prefixed tags before `processBulk`. The block forwards only the remainder;

**Breaking**: `pythonScript` -> `python_script`, and the separate `script_uri` setting is gone — `python_script` now holds either the script or its location.

### DIFF
--- a/blocks/basic/include/gnuradio-4.0/basic/PythonBlock.hpp
+++ b/blocks/basic/include/gnuradio-4.0/basic/PythonBlock.hpp
@@ -80,15 +80,15 @@ myBlock.processBulk(ins, outs);
 )"">;
     // optional shortening
     template<typename U, gr::meta::fixed_string description = "", typename... Arguments>
-    using A                = Annotated<U, description, Arguments...>;
-    using poc_property_map = std::map<std::string, std::string, std::less<>>; // TODO: needs to be replaced with 'property_map` aka. 'ValueMap'
-    using tag_type         = std::string;
+    using A                 = Annotated<U, description, Arguments...>;
+    using StringPropertyMap = std::map<std::string, std::string, std::less<>>; // TODO: replace with gr::property_map once pmt::Value is Python-wrapped
+    using tag_type          = std::string;
 
-    std::vector<PortIn<T>>                                                        inputs{};
-    std::vector<PortOut<T>>                                                       outputs{};
-    A<gr::Size_t, "n_inputs", Visible, Doc<"number of inputs">, Limits<1U, 32U>>  n_inputs     = 0U;
-    A<gr::Size_t, "n_outputs", Visible, Doc<"number of inputs">, Limits<1U, 32U>> n_outputs    = 0U;
-    std::string                                                                   pythonScript = "";
+    std::vector<PortIn<T>>                                                         inputs{};
+    std::vector<PortOut<T>>                                                        outputs{};
+    A<gr::Size_t, "n_inputs", Visible, Doc<"number of inputs">, Limits<1U, 32U>>   n_inputs     = 1U;
+    A<gr::Size_t, "n_outputs", Visible, Doc<"number of outputs">, Limits<1U, 32U>> n_outputs    = 1U;
+    std::string                                                                    pythonScript = "";
 
     GR_MAKE_REFLECTABLE(PythonBlock, inputs, outputs, n_inputs, n_outputs, pythonScript);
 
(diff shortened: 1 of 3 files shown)


blocks/basic/include/gnuradio-4.0/basic/PythonBlock.hpp
```diff
@@ -195,154 +247,379 @@ this_block = PythonBlockWrapper(capsule))p",
...
+        }
+        for (const std::string& directory : python_path.value) {
+            python::PyObjectGuard entry(PyUnicode_FromString(directory.c_str()));
+            if (!entry) {
+                python::throwCurrentPythonError(std::format("{}: cannot convert python_path entry '{}'", this->name, directory));
+            }
+            if (PySequence_Contains(sysPath, entry) == 1) { // re-configuration must not grow sys.path without bound
+                continue;
+            }
+            if (PyList_Insert(sysPath, 0, entry) != 0) {
```
drslebedev, line 342: Each entry is inserted at index 0, so python_path = [A, B] produces sys.path = [B, A]. This reverses the requested import priority. 
Please review if it is intended. For me it looks like a bug.

blocks/basic/include/gnuradio-4.0/basic/PythonBlock.hpp
```diff
@@ -142,49 +185,58 @@ this_block = PythonBlockWrapper(capsule))p",
...
-                        python::throwCurrentPythonError(std::format("{}(aka. '{}')::settingsChanged(...) - script parsing error", this->unique_name, this->name), std::source_location::current(), pythonScript);
-                    }
-
-                    python::PyObjectGuard pyFunc(PyObject_GetAttrString(_interpreter.getModule(), "process_bulk"));
-                    if (!pyFunc.get() || !PyCallable_Check(pyFunc.get())) {
-                        python::throwCurrentPythonError(std::format("{}(aka. {})::settingsChanged(...) Python function process_bulk not found or is not callable", this->unique_name, this->name), std::source_location::current(), pythonScript);
-                    }
-                },
-                pythonScript);
+        if (new_settings.contains("python_script") || new_settings.contains("python_path")) {
```
drslebedev, line 188: This makes python_path look runtime-reconfigurable, but old paths are never removed and previously imported modules remain. Should we make this setting startup-only?

blocks/basic/include/gnuradio-4.0/basic/PythonBlock.hpp
```diff
@@ -195,154 +247,379 @@ this_block = PythonBlockWrapper(capsule))p",
...
+        // N.B. tags() is relative to the start of this span, which is exactly what publishTag() expects; rawTags() would be absolute
+        if constexpr (requires(TInputSpan span) { span.tags(); }) {
+            for (TInputSpan& span : ins) {
+                for (const auto& [relativeIndex, mapRef] : span.tags()) {
+                    const std::size_t offset = relativeIndex < 0 ? 0UZ : static_cast<std::size_t>(relativeIndex); // a tag of already-consumed samples belongs at the chunk start
+                    TagView           view{offset, {}};
+                    gr::property_map  forwarded;
+                    for (const auto& [key, value] : mapRef.get()) {
+                        view.map[std::string(key)] = toScriptText(value);
+                        if (!std::string_view(key).starts_with(gr::GR_TAG_PREFIX.view())) {
```
drslebedev, line 364: I think this is intended but still worth to note the framework already forwards the gr: part of this tag. Forwarding only the remaining keys here splits one mixed tag into two separate output tags.

blocks/basic/include/gnuradio-4.0/basic/PythonBlock.hpp
```diff
@@ -195,154 +247,379 @@ this_block = PythonBlockWrapper(capsule))p",
...
+    // N.B. the `requires` guards keep the block usable with plain std::span, which the unit tests and the usage example pass directly
+    template<typename TInputSpan>
+    void collectInputTags(std::span<TInputSpan> ins) {
+        _inputTags.clear();
+        _forwardTags.clear();
+        _nextInputTag = 0UZ;
+        _pendingOutputTags.clear(); // a throwing script must not leak its tags into the next invocation
+        // N.B. tags() is relative to the start of this span, which is exactly what publishTag() expects; rawTags() would be absolute
+        if constexpr (requires(TInputSpan span) { span.tags(); }) {
+            for (TInputSpan& span : ins) {
```
drslebedev, line 357: This loop does not deduplicate tags across inputs, this could easily publish identical tag twice. I think we need similar deduplication as in the base Block class.

drslebedev (approved):
Thanks for the PR! 

Overall it looks very good. It significantly improves PythonBlock correctness and adds  tag-handling functionality. 

I left a few inline comments concerning python_path ordering and tag forwarding. 

The PR can be merged after comments are resolved.

blocks/basic/include/gnuradio-4.0/basic/PythonBlock.hpp
RalphSteinhagen: Correct, and intended. `Block<>::forwardInputTags` forwards only `gr:`-prefixed keys plus the `autoForwardParameters()` allow-list, so a mixed tag arrives split: the framework publishes the gr: half, the block the rest, both at the same index.

Forwarding at all is necessary because a scripting block's tags are arbitrary by nature: the default drops every non-`gr:` key, so without this a `PythonBlock` silently destroys user tag metadata for everything downstream.

RalphSteinhagen:
> Thanks for the PR!
> 
> Overall it looks very good. It significantly improves PythonBlock correctness and adds tag-handling functionality.
> 
> I left a few inline comments concerning python_path ordering and tag forwarding.
> 
> The PR can be merged after comments are resolved.

@drslebedev thanks a lot for the quick review. Much appreciated. Aside the fixes, this PR was also supposed to test the latest `ccache` integration which worked reasonably well.

Both bugs confirmed and fixed, with a regression test each that fails without the fix (reversed order gave [2, 4, 6]. 
The duplicate showed as 2 tags. Amended into the existing commits and force-pushed.

RalphSteinhagen:
for reference:

image

PR #830, opened 2026-09-07
fixes Pre-Device Integration

RalphSteinhagen:
This fixes defects that predate the device-integration branch and are independent of it.

Suggested reviewing commit by commit; each commit is self-contained, builds with `-Werror` and passes on its own. The commit messages carry the detail.

Three worth knowing before:

- **`double_mapped_memory_resource` leaked one ring per destroyed buffer** — `munmap` released half of what `mmap` reserved. Invisible in RSS or other leak-checkers (that usually add hooks to `new`/`delete` only) because the mirrored half is never resident, which is why the test reads `/proc/self/statm` rather than resident size. Watch system `shmem` if this needs to be reproduced on `main`.
- **`ValueMap::from_blob` rejected every nested map** (`AlignmentViolation`). The 16-byte blob alignment was applied recursively, but a child sits inside its parent's allocation at a fixed offset and can never satisfy it. The pre-existing `CorruptOffset || AlignmentViolation` hedge in `qa_ValueMap` was the symptom.
- **A scalar read trusted a record's declared type and length.** A tensor and a nested map both record their _element_ type in the same field, so `get_if<float>` on a `Tensor<float>` returned a live pointer into the tensor's own header; and a string overwritten with a shorter one returned the slot length, not the string.
- `gr::complex` and `std::complex` now share one wire tag: a value is stored and read back under either, and the format is unchanged.


core/include/gnuradio-4.0/ValueMap.hpp
```diff
@@ -2471,8 +2626,8 @@ class ValueMap : public ValueMapView {
         if (bytes.size() < sizeof(Header)) {
             return std::unexpected{DeserialiseError::TooSmall};
         }
-        if ((reinterpret_cast<std::uintptr_t>(bytes.data()) & (kBlobAlignment - 1UZ)) != 0UZ) {
-            return std::unexpected{DeserialiseError::AlignmentViolation};
+        if (depth == 0U && (reinterpret_cast<std::uintptr_t>(bytes.data()) & (kBlobAlignment - 1UZ)) != 0UZ) {
```
drslebedev: Should we still check 8-byte alignment for nested maps? Header and PackedEntry require it.

core/include/gnuradio-4.0/ValueMap.hpp
```diff
@@ -444,23 +451,24 @@ constexpr decltype(auto) dispatchValueType(Value::ValueType vt, F&& f) {
...
+                       || std::same_as<std::remove_cvref_t<T>, std::int32_t>        //
+                       || std::same_as<std::remove_cvref_t<T>, std::int64_t>        //
+                       || std::same_as<std::remove_cvref_t<T>, std::uint8_t>        //
+                       || std::same_as<std::remove_cvref_t<T>, std::uint16_t>       //
+                       || std::same_as<std::remove_cvref_t<T>, std::uint32_t>       //
+                       || std::same_as<std::remove_cvref_t<T>, std::uint64_t>       //
+                       || std::same_as<std::remove_cvref_t<T>, float>               //
+                       || std::same_as<std::remove_cvref_t<T>, double>              //
+                       || std::same_as<std::remove_cvref_t<T>, std::complex<float>> //
+                       || std::same_as<std::remove_cvref_t<T>, gr::complex<float>>;
```
drslebedev, line 473: Should we also update the ValueMapView writers when adding this type? They currently label complex values as Scalar, but the reader expects Complex.

core/include/gnuradio-4.0/PmtTypeHelpers.hpp
```diff
@@ -352,6 +356,28 @@ template<class T, bool strictCheck = false, typename From>
...
+            std::vector<TElem> converted;
+            converted.reserve(srcValue.size());
+            for (const auto& elem : srcValue) {
+                auto conv = convert_safely<TElem>(elem);
+                if (!conv) {
+                    return std::unexpected(std::format("tensor element failed to convert: {}", conv.error()));
+                }
+                converted.push_back(*conv);
+            }
+            return T(converted.begin(), converted.end());
```
drslebedev: Do we expect complex tensors to always be 1D?

core/include/gnuradio-4.0/ValueMap.hpp
```diff
@@ -1481,6 +1491,77 @@ struct ValueMapView {
...
+            return {};
+        }
+        const std::uint32_t childCapacity = static_cast<std::uint32_t>(childBytes);
+        const std::uint32_t recSize       = std::max<std::uint32_t>(static_cast<std::uint32_t>(recSize64), kRecMinSize);
+
+        // A record's content sits at recordStart + kRecHeaderBytes, and a blob must start on kBlobAlignment. The
+        // record start is ours to choose -- readers reach content through the entry's payloadOffset -- so bias it
+        // by the header size and the child lands aligned without any reader knowing.
+        const std::uint32_t curUsed      = _header->payloadUsed;
+        const std::uint32_t contentStart = static_cast<std::uint32_t>((_header->payloadOffset + curUsed + kRecHeaderBytes + (kBlobAlignment - 1U)) & ~static_cast<std::uint32_t>(kBlobAlignment - 1U));
```
drslebedev: I think we should include _blob in the alignment calculation. An existing nested map may only be 8-byte aligned, so rounding contentStart to 16 bytes does not guarantee that _blob + contentStart is 16-byte aligned.

drslebedev (approved):
Thanks for the PR! 
The changes look good overall. 

I found a few issues that I think should be addressed, but otherwise this looks ready to merge.

core/include/gnuradio-4.0/ValueMap.hpp
RalphSteinhagen: fixed. I initially lazily dropped this because I felt it wasn't as critical for nested structures. However, safety aside, this could/should have some performance impact. Not measured yet, though.

core/include/gnuradio-4.0/ValueMap.hpp
RalphSteinhagen: Confirmed and fixed.

core/include/gnuradio-4.0/PmtTypeHelpers.hpp
RalphSteinhagen: No — good catch, that was a real bug. The copy was correct but the shape was missing, so a n-dim Tensor would have been silently be converted into a 1dim one. Fixed and tested.

core/include/gnuradio-4.0/ValueMap.hpp
RalphSteinhagen: Agreed, fixed.

PR #831, opened 2026-09-14
Flatten out port serizliation yaml

ivan-cukic:


RalphSteinhagen (approved):
OK. Before doing further changes to the (Sub-)Graph API we should discuss the general design changes. As discussed in the channel, there are some UX/API issues with the present implementation which need to be addressed more sustainably.

PR #832, opened 2026-09-16
Fix subgraph settings loading from YAML

drslebedev:
Fix subgraph settings loading from YAML

- Share parameter and context loading function across normal blocks and subgraphs.
- Add regression coverage for subgraph positions and conflicting names.


RalphSteinhagen (approved):
Thanks for the fix.

PR #836, opened 2026-09-21
Feat(core): same GR4 block on CPU, GPU (via SYCL, CUDA, ROCm), and MCU + Core Performance Improvements

### LINKED ISSUE #162: [5SP, 5SP] Implement up-/downconversion filter (i.e. 'Frequency XLating' filter)
Now tackled as a MVP spike/proof-of-concept of: https://github.com/fair-acc/gnuradio4/issues/628

### LINKED ISSUE #628: [10SP,25SP] SYCL PoC: Multi-FFT Kernel for Batched Spectral Processing

RalphSteinhagen:
GNU Radio 4.0 blocks and infrastructure ran until now only on either a CPU or MCU.  This PR adds the capability to dispatch GR4 flow-graphs efficiently to GPUs with minimal overhead. The default GPU integration is via SYCL (AdaptiveCpp): 
a block says where it would like to run and the framework puts it there, so the source, the tags, the settings and the tooling stay the same.

Nothing in a block names a vendor, and none of it is imposed — a build without a backend neither compiles nor links any of this, and still targets an MCU. The block developer keeps writing ordinary C/C++ code ... that's all. For the experienced, there are escape hatches for SYCL (default, favoured), CUDA and ROCm. _N.B. The latter backends are tested but not further followed-up by this repo and should be expanded elsewhere._

Of the ~30k lines, most are unit tests, benchmarks and new block implementations, written alongside the core API rather than after it to keep changes to the core-API cheap.
This was to ensure that the new core-API extensions survive the contact with real-world block applications (here: resamplers, channelizers, filters, FFTs, etc.) -- which I believe it did.

This PR is best read in this order:
 1. user-facing API (branch tip), 
 2. measured benchmark figures, 
 3. new device-related core invariants, and then 
 4. the commit-by-commits last — these curated to be read one at a time rather than as one large diff.

## TL;DR

**Host vs. Device Compute-Domain is a Setting**, not a port and not a rewrite:

```cpp
auto& fft = flow.emplaceBlock<gr::blocks::fft::FFT<std::complex<float>>>({
    {"compute_domain", "gpu:sycl"}, {"fft_size", 4096U}
});
```

**The domains the runtime registers:**

| spelling     | meaning                                             |
| ------------ | --------------------------------------------------- |
| `host`       | default — no device, no SYCL                        |
| `host:sycl`  | SYCL CPU device (fallback to multi-threaded OpenMP) |
| `gpu:sycl`   | SYCL GPU device                                     |
| `gpu:sycl:1` | that device, by index                               |
| `gpu:sycl!`  | required — the block stops rather than falls back   |

Without the `!` the ladder is `gpu:sycl → host:sycl → native`.

**Three block entry points.** The first two are unchanged; the third is new and is the only one that names a device:

```cpp
// 1:1 sample transform — host or device, vectorised where the target allows
[[nodiscard]] constexpr T processOne(T sample) const noexcept;

// span access, resampling, variable rate — host or device
[[nodiscard]] gr::work::Status processBulk(InputSpanLike auto& in, OutputSpanLike auto& out);

// the hatch: the block is handed the context and does its own consume/publish
[[nodiscard]] gr::work::Status processBulk(gr::device::DeviceContext& ctx,
                                           InputSpanLike auto& in, OutputSpanLike auto& out);
```

_N.B. the functional code inside remains standard C++ code regardless of whether being run on CPU or GPU._

**A chain crosses the host<->device memory boundary once**, not once per hop and only when needed (thanks to USM):

```cpp
auto group = gr::makeDeviceGroup(std::move(innerGraph));  // -> DeviceGroupHandle
// every port no interior edge claims is exported as "<member>:<port>"; interior edges stay device-resident
```

## What it adds

**Compute-Domain.** A block declares `compute_domain` (`gpu:sycl`, `host:sycl`, `host`).
Whoever drives the block — the scheduler, or a sub-graph — resolves that once per run and hands the answer to every `work()` call.
A domain spelled with a trailing `!` is required: if it cannot be served the block stops rather than quietly running on the host.
The fallback ladder is `gpu:sycl → host:sycl → native`.

**A device runtime** behind one contract: USM memory, queue-ordered transfers, an error channel, and diagnostics from inside a kernel rendered on the host.
Device-to-device edges carry device-only memory with no host round trip, including fan-out.

**Three ways to write a device block**, in increasing order of control: an ordinary `processOne`, a `processBulk` over spans, or a `processBulk(ctx, …)` hatch that receives the `DeviceContext` and does its own accounting.
Nothing falls back silently.

**A group of blocks moved across the boundary as one**, so a chain pays one transfer rather than one per hop.
`gr::makeDeviceGroup(graph)` returns a `DeviceGroupHandle`, exporting every port no interior edge claims under a `"<member>:<port>"` name; `BridgedDeviceGroup` and the `TransferBlocks` pair cover the explicit crossings.

**DSP on the new interfaces** — a device FFT, one FFT block for host and device, transform-domain convolution, rational and drift resampling, correlation, span reductions, Savitzky-Golay, and the FIR/IIR filters restated as a window and a recursion.

The user-facing blocks and algorithms this adds, named so a reviewer can find them:
 * `PolyphaseChannelizer`,
 * `PolyphaseSynthesizer`, 
 * `HierarchicalPolyphaseChannelizer`,
 * `PolyphaseResampler`,
 * `PolyphaseArbitraryResampler`, 
 * `HermiteResampler` (the drifting ratio), 
 * `AllpassHalfBand`,
 * `FrequencyXlatingFilter`, 
 * `InterleavedStreamDecimator`, 
 * `FrameStatistics` (e.g. sliding-window RMS computation),
 * `FastConvolutionFilter`,
 * `Correlator`, 
and the `Reduce`/`SyclReduce` span reductions.
`DifferenceEquation` and `FilterForms` carry the IIR structures; `WindowGeometry` carries the window/stride geometry the resamplers and the FFT block share.

**Reduced Framework Overhead paid previously by every graph.** Three fixes that are independent of the device work and apply to `main`'s own hot path: the scheduler no longer rebuilds a hash index of moved blocks on every iteration, a port cache no longer re-derives a compile-time-constant answer per `work()` call, and the `processOne` loop tests one flag per sample instead of a flag and an atomic state reload.
Measured below.

## For reviewers

**Read the tip for the benchmarks and the spike examples.** For the code, read the series: the history is ordered as a functional composition — contract, runtime, dispatch, then blocks — so each commit is one digestible piece rather than the whole surface at once.

**The last ten or so commits are the DSP spikes.** They are the worked examples of the API: what a block author writes against `processOne`, `processBulk`, and the hatch.
Read them for the usage patterns — the numbers behind each one are in its commit message and its code.

**The first three commits are not device work.** They fix framework overhead `main` pays too, apply to `main` unchanged, and can be reviewed or taken on their own.

<details>
<summary>Commit-by-commit reading order (26 commits)</summary>

```
 1. perf(scheduler): stop rebuilding the moved-block index when nothing moved
 2. perf(core): decide a single port's sample limits at compile time
 3. perf(core): read a tag's settings from the tag, not from a copy of it
 4. fix(core): let a memory resource survive the containers that were given it
 5. fix(core): give a block emplaced into a graph the graph's compute domain
 6. test(core): drop a tag assertion that cannot fail
 7. feat(core): a compute domain, its requirement marker, and one parser for both
 8. feat(core): a leading underscore keeps a reflected member off the settings surface
 9. feat(device): a device runtime, its memory and its dispatch, behind one contract
10. feat(core): let a scheduler place a block on a device, and dispatch it there unchanged
11. feat(core): give each device edge the memory its two ends actually need
12. feat(basic,core): move a group of blocks across the host/device boundary as one
13. feat(device): a discrete Fourier transform that runs on a device
14. feat(fourier): one FFT block that runs on the host or a device
15. build,test,bench,docs(device): compile the device benchmarks only when asked, and measure them honestly
16. feat(filter,algorithm): filtering by transform, for filters long enough to earn it
17. refactor(filter): state the FIR as a window and the IIR over a span, so that both stay on a device
18. feat(filter): say where BasicFilter's coefficients come from, run its cascade over a span, and let the FIR choose its own domain
19. perf(algorithm,filter): make the transform-domain convolution fast, not merely correct
20. perf(algorithm,filter): one difference equation per shape, and state that follows the design
21. perf(filter): give the tap form the loop shape its target wants
22. feat(algorithm,filter): resampling at a rational ratio and at a drifting one
23. feat(algorithm,filter): reductions over a span, and two blocks alongside them
24. feat(filter): name the Savitzky-Golay modes by their enumeration, not by string
25. perf(algorithm): let the Savitzky-Golay window reduction vectorise
26. feat(device): a second and third backend behind the same device contract
```

</details>

## Measured

Single-threaded, Ryzen 9 5900X at 3.66 GHz.
`%max` is against the attainable rate (`3660 MHz / (kernel cycles/sample + 3.7)`), not against an unreachable ceiling.
In _The numbers_ below it is instead against each benchmark's own measured chain ceiling, quoted in the same run.

| subject      | where it pays                                                                                                                                                                      |
| ------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `fir_filter` | time domain below ~128 taps; the transform domain overtakes above, holding 87–96 % of attainable to 32 k taps                                                                      |
| FFT          | `gpu:sycl` overtakes the host at large transforms; streaming mode sustains 56–66 % of the framework's own empty-chain ceiling                                                      |
| IIR          | the kernel beats a hand-written direct-form-II by 20 % — a recursion is sequential by nature and is on a device for residency, not for speed                                       |
| channelizers | FIR banks hold 42–50 % of the GPU ceiling; a 715-tap xlating FIR reaches 84 % (10x the acpp host, 2.8x the gcc-15 host); every IIR structure is the opposite and belongs on a host |
| SimdFFT      | 87–101 % of pffft on the same machine                                                                                                                                              |
| framework    | per-`work()` dispatch overhead cut against `main`: +13.2 % streaming, +41.3 % at one tag per sample                                                                                |

### The numbers

| benchmark / scenario                                   | unit     | `main` | branch gcc-15 | branch acpp |   Δ vs main | %max (gcc-15 / acpp) |
| ------------------------------------------------------ | -------- | -----: | ------------: | ----------: | ----------: | -------------------- |
| **qa_PerformanceMonitor** — no tags                    | MS/s     |    532 |       **602** |           — | **+13.2 %** | —                    |
| **qa_PerformanceMonitor** — moderate tag rate          | MS/s     |    457 |       **524** |           — | **+14.7 %** | —                    |
| **qa_PerformanceMonitor** — one tag per sample         | kS/s     |    849 |      **1200** |           — | **+41.3 %** | —                    |
| bm_Filter — chain ceiling, host                        | MS/s     |      — |        1015.1 |      1043.4 |         new | 100 % / 100 %        |
| bm_Filter — `fir_filter` AUTO, 64 taps                 | MS/s     |      — |         428.2 |       405.2 |         new | 42.2 % / 38.8 %      |
| bm_Filter — `fir_filter` AUTO, 8192 taps               | MS/s     |      — |         198.9 |       136.0 |         new | 19.6 % / 13.0 %      |
| bm_Filter — `iir_filter`, 2nd order                    | MS/s     |      — |         423.3 |       158.8 |         new | 41.7 % / 15.2 %      |
| bm_HChannelizer — chain ceiling, host                  | MS/s     |      — |        1201.5 |       540.4 |         new | 100 % / 100 %        |
| bm_HChannelizer — Xlating FIR 715, host                | MS/s     |      — |         119.9 |        33.5 |         new | 10.0 % / 6.2 %       |
| bm_HChannelizer — Xlating FIR 715, **gpu:sycl**        | MS/s     |      — |    host-only¹ |   **339.3** |         new | — / **84.2 %**       |
| bm_HChannelizer — FIR flat bank 16 ch, host            | MS/s     |      — |          70.7 |       127.1 |         new | 5.9 % / 23.5 %       |
| bm_HChannelizer — FIR flat bank 16 ch, gpu:sycl        | MS/s     |      — |    host-only¹ |       169.1 |         new | — / 42.0 %           |
| bm_HChannelizer — IIR allpass 16 ch, gpu:sycl          | MS/s     |      — |    host-only¹ |         3.6 |         new | — / 0.9 %            |
| bm_DeviceDispatch — multiply 1 flop, 64k chunk, host   | MS/s     |      — |    not built² |       190.1 |         new | —                    |
| bm_DeviceDispatch — same, host:sycl                    | MS/s     |      — |    not built² |       151.3 |         new | —                    |
| bm_DeviceDispatch — same, gpu:sycl                     | MS/s     |      — |    not built² |       145.9 |         new | —                    |
| bm_DeviceDispatch — fixed cost per dispatch, host:sycl | µs       |      — |    not built² |         2.4 |         new | —                    |
| bm_DeviceDispatch — fixed cost per dispatch, gpu:sycl  | µs       |      — |    not built² |        69.9 |         new | —                    |
| bm_FFT_backends — SimdFFT, N=4096                      | µs/xform |      — |            13 |          15 |         new | —                    |
| bm_FFT_backends — SyclFFT:CPU, N=4096                  | µs/xform |      — |            14 |          16 |         new | —                    |
| bm_FFT_backends — SYCL GPU (RTX 3070), N=4096          | µs/xform |      — |    host-only¹ |          46 |         new | —                    |

¹ gcc-15 has no SYCL backend, so `host:sycl`/`gpu:sycl` resolve to the host path — those rows measured within noise of `host` and are omitted rather than repeated. ² `bm_DeviceDispatch` is gated on a device backend and is not generated at all under gcc-15.

Only `qa_PerformanceMonitor` has a `main` counterpart — its source is byte-identical on both sides, so it is the one row group that is a _change_ rather than a new capability.
Everything else is added by this branch.

Two things: a GPU earns its place where the arithmetic is dense and independent (a 715-tap xlating FIR at 84 % of its own ceiling) and loses badly where the recursion is sequential (an IIR allpass tree at 0.9 %).
And dispatch has a floor — 2.4 µs host, 69.9 µs GPU — which is why the small-chunk columns collapse.
To note: FIR is often touted as the linear-phase option.
IIRs, properly designed, are sufficiently linear-phase (errors in the 0.001° range) and much more performant.
The IIR cases are shown to be often much faster on the host than the next-best FIR-based implementation on the GPU.
Correct assumptions of physical invariants still beat numerics by a mile!

<details>
<summary>Full DSP tables: channelizers, resamplers, distance from attainable</summary>

These tables are from an earlier measurement run than _The numbers_ above; where a row appears in both, the figures there are the later ones.
Input MSample/s (a channelizer's cost is per input sample, so it is not flattered by the channels it fans out to), over 2^24 samples, AdaptiveCpp Release on CUDA, idle machine.
The graph is driven by `NullSource` into `CountingSink` — the cheapest pair available, so what is measured is the block rather than the harness.
`chain ceiling` is the same graph with a bare copy in place of the filter, and each figure is quoted against the ceiling **for its own domain**.

**`host:sycl` is held to one OpenMP thread** (`OMP_NUM_THREADS=1`, set by the benchmark before the runtime starts).
GR4's host path is single-threaded by design and `host:sycl` is a _host_ device, so it must be held to the same rule — otherwise that column reports a thread count rather than a throughput.
Unpinned it read up to 2.6x faster than plain `host` on the banked rows; pinned to one core, plain `host` won every one of them.
An explicit `OMP_NUM_THREADS` in the environment still wins, so the threaded figure stays measurable.

2^24 rather than 2^20 matters: `runAndWait()` calls `start()`, which connects the edges, and a pinned device edge allocates there — inside the timed region.
That cost is per run and per EDGE, so at 2^20 it penalised the 16-port rows far more than the single-edge ceiling and understated the device columns roughly twofold.

| ch  | structure         |  host |  %ceil | host:sycl |  %ceil | gpu:sycl |  %ceil |
| --- | ----------------- | ----: | -----: | --------: | -----: | -------: | -----: |
| –   | chain ceiling     | 440.1 |      – |     400.6 |      – |    354.7 |      – |
| 1   | Xlating FIR (715) |  25.9 |  5.9 % |       5.1 |  1.3 % |    304.0 | 85.7 % |
| 1   | Xlating FFT (715) |  84.3 | 19.2 % |       5.1 |  1.3 % |    317.8 | 89.6 % |
| 1   | Xlating FIR (91)  | 128.7 |      † |      32.6 |      † |    424.6 |      † |
| 1   | Xlating IIR (3)   |  56.2 | 12.8 % |      36.0 |  9.0 % |      1.0 |  0.3 % |
| 1   | Drift, Hermite    | 343.3 |      † |     300.5 |      † |    532.3 |      † |
| 1   | Drift, polyphase  |  40.1 |      † |      32.9 |      † |    530.2 |      † |
| 8   | FIR polyphase 2x4 |  33.5 |  7.6 % |      26.2 |  6.5 % |    124.2 | 35.0 % |
| 8   | IIR allpass tree  | 113.3 | 25.7 % |     101.2 | 25.3 % |      3.2 |  0.9 % |
| 8   | FIR flat bank     |  96.5 | 21.9 % |      71.2 | 17.8 % |    173.5 | 48.9 % |
| 16  | FIR polyphase 4x4 |  42.7 |  9.7 % |      31.3 |  7.8 % |    109.3 | 30.8 % |
| 16  | IIR allpass tree  |  91.0 | 20.7 % |      81.7 | 20.4 % |      3.6 |  1.0 % |
| 16  | FIR flat bank     | 100.3 | 22.8 % |      49.7 | 12.4 % |    144.2 | 40.7 % |

The host xlating rows moved a long way since the earlier revision of this table, from two changes that only touch the host path: the NCO now rotates a phasor instead of evaluating a sine and a cosine per sample (6.05x on the mixer alone), and the overlap-save convolution runs its transforms unordered.
Those rows were 18.3 / 34.2 / 42.2 / 29.7; the gain tracks how large a share the mixer had, so 91 taps gains 3.03x and 715 taps 1.42x.
`host:sycl` and `gpu:sycl` are unchanged by construction — a work item still computes its own angle, and the device tap spectrum is still natural-order.
No channelizer row moved, which also bounds this machine's run-to-run noise at about two percent.
Rows repeat to within half a percent inside one binary but swing four to five across rebuilds of untouched code, so a single-digit move on a row nothing touched is code layout, not a result — `Xlating IIR (3)` reads between 54.9 and 59.6 across eight builds of this branch.

The GPU channelizer rows carry a transfer change: the staged device path scattered its result to the channel ports with one BLOCKING copy per channel, and at `outputs_per_frame = 64` each of those moves only 512 bytes, so the transfer's fixed cost was paid `n_channels` times over and the payload never amortised it.
The copies are now queued and awaited once.
Measured, best of three: 8ch flat bank 157.6 -> 173.5, 16ch flat bank 124.9 -> 144.2, 8ch polyphase 114.7 -> 124.2, 16ch polyphase 97.2 -> 109.3 — between 8 % and 16 %, growing with the channel count, exactly as a per-transfer-overhead story predicts.
`host:sycl` is untouched, which is the check that it is the staging path doing it: that domain has no remote memory and never stages.

It does NOT fix the 8-to-16-channel fall-off, and the numbers say so plainly: the flat bank goes from 48.9 % of ceiling to 40.7 %, where before the change it went 45.4 % to 36.0 %.
Both levels lift; the slope barely moves.
That residual is the O(M²) cross-channel DFT, analysed below.

Cross-tested on AMD by a second machine — an RX 7700S (gfx1102) under ROCm 6.4 and AdaptiveCpp 25.10, base and modified binaries built from the same source and run interleaved, six rounds with the first discarded as JIT warm-up.
It reproduces, larger: **+17 %, +25 %, +51 % and +60 %** on the 8ch polyphase, 8ch flat, 16ch polyphase and 16ch flat rows.
Two qualifications the AMD run's own author asked be attached rather than dropped: those deltas were measured with `defaultMinBufferSize` at the 16384 the branch carried at the time, and a smaller ring makes the fixed per-copy cost a larger share of each row — so they are an upper-ish bound for the shipped 65536 default, where the NVIDIA figures were taken.
The qualitative claim is what survives both: the gain tracks how transfer-dominated the path is, and grows with the channel count because the number of synchronisations removed grows with it (N queued copies and one wait, in place of N waits).

The three channelizer suites also pass on AMD with the change in, and produced no HSA memory-access fault — which is what ROCm raises when a queued copy touches freed USM.
That is the strongest signal either side has that the `wait()` sits in the right place, though it is absence of the failure mode rather than proof: it says the ordering holds for these frame sizes and this allocator, not that it was probed adversarially.

`Drift, polyphase` on the host first went the other way, 33.8 -> 31.6, which was the cost of correcting the polyphase arm interpolation (see _Resampler accuracy_): it reads one sample further and its guard is tighter, and it buys +12.8 dB.
It now reads 40.1, because the arms run over separated real and imaginary parts: a polyphase arm weights its samples with real taps, so each arm becomes a real dot product that the vectorised routine can take, where interleaved complex cannot use it at all.
Splitting the window and recombining the outputs is 2 % of the path, measured.

That row is 4.4x faster under gcc and only 1.2x under clang, and the reason is worth stating because it bounds several other rows too. gcc has libstdc++'s `<experimental/simd>` behind `vir::simd`; clang has no standard-library simd and falls back to vir's own, which costs 9.2 ns against 4.2 on the same 32-tap dot (clang 20 and 21 alike, so it is not AdaptiveCpp).
Every clang FIR row pays it — the two `Xlating FIR` rows share the routine.
The table above is an AdaptiveCpp build, so it reports the clang figure.

† marks the **decimating** rows, where the ceiling is not a bound: it is measured with a 1:1 copy, but these blocks pass only a fraction of their input downstream — 1/8 for `Xlating FIR (91)`, 1/2 for the two `Drift` rows at `ratio = 0.5`.
A decimating figure above the ceiling is not an error; the chain below it simply carries less data.

The bracketed number on the xlating rows is the **tap count**, and it is the whole story of those rows: `filter_order` sets the transition width (`0.1/order`), not the length, so order 32 asks for a narrow transition and the Kaiser estimate answers with 715 taps while order 4 gives 91.

Reading it:

- **A long FIR is what a device is for.** The 715-tap xlating filter goes 18.3 → 298.2 MSample/s, **16x**,
  and reaches 86 % of the GPU's own ceiling. Every output is an independent dot product, which is exactly the
  shape that relocates.
- **On the host the same design is a transform, not a loop.** `Xlating FFT` is `Xlating FIR` with
  `filter_domain` left at its default: at 715 taps `Auto` evaluates it by overlap-save instead of a tap per
  sample and the row goes 18.3 → 34.2, **1.87x**, for the same filter. The transform's tap spectrum and FFT
  plan are host objects, so a device domain keeps the tap form — which is why the two rows agree on
  `host:sycl` and `gpu:sycl`, and why nothing there falls back.
- **An arbitrary ratio parallelises after all.** `DriftResampler` owns its own accounting because its output
  count per input is not a chunk pair — but within one call the ratio is fixed, so output `m` reads
  `phase + m / ratio` and the outputs are independent. Evaluating them as one flat parallel loop takes the
  polyphase kernel from 33.8 to **495.0** MSample/s on `gpu:sycl` (**15x**) and the cubic from 343.4 to 551.5
  (**1.6x**) — that A/B left the host `processBulk` untouched, and its two rows stayed within 0.2 % of their
  previous figures; the host polyphase row has since been vectorised separately, as described above. The
  earlier note that this block "stays sequential" confused the accounting with the arithmetic.
- **A recursion is what a device is not for.** The same block's IIR setting is 0.9 on the GPU against 29.7 on
  the host — 33x _worse_. It runs in a single work item because a recursion has no parallelism to offer, and
  it is on the device for residency (no host round trip when it sits between two device blocks), never for
  speed. The IIR allpass tree's 2 % says the same thing, and both are device figures, not fallbacks.
- **The FIR banks sit in between**, holding 28–45 % of the GPU ceiling. The flat bank falls from 45.4 % to
  36.0 % of ceiling between 8 and 16 channels on the GPU, where the host barely moves: the device kernels write each channel to its own pinned host array with the channel index
  fastest, so the write scatters further as the channel count grows. That is the next thing to fix.
- **`host:sycl` is slower than plain `host` almost everywhere**, once both are held to one thread — 49.3
  against 99.0 on the 16-channel flat bank. Two causes, and they are different: AdaptiveCpp's host codegen is
  weaker than gcc's on the same source, and the blocks that own a
  `processBulk(ctx, …)` hatch run their _device_ kernel there, which is deliberately scalar so it can be
  emitted for a GPU. `Xlating FIR` at 5.1 against 18.3 is that second effect: the host path takes
  `dotReversedSimd`, the hatch takes the scalar `dotReversed` in a single work item. Neither figure is a
  defect; `host:sycl` exists to run device code on a CPU, not to be the fast host path.

### Resampler accuracy

`DriftResampler` carries two interpolation kernels and the setting is only meaningful if the difference is measurable.
A resampler delays what it interpolates by a fraction of a sample that depends on the kernel, so comparing sample against sample measures the alignment rather than the interpolation — which is what made an earlier attempt at this table unpublishable.
For a single complex tone the whole ambiguity, delay and gain together, is one complex scalar, so fitting it by least squares removes it exactly and what is left is the error the kernel made.
Interpolating (ratio 1.37), never decimating, so the figures describe interpolation rather than aliasing.

| f / f_in | Hermite dB | Polyphase dB |
| -------: | ---------: | -----------: |
|     0.02 |       83.2 |         46.0 |
|     0.05 |       67.9 |         51.6 |
|     0.10 |       50.0 |         43.3 |
|     0.15 |       38.6 |         42.3 |
|     0.20 |       30.2 |         45.7 |
|     0.25 |       23.5 |         43.4 |
|     0.30 |       17.8 |         42.8 |
|     0.35 |       12.9 |         37.7 |
|     0.40 |        8.3 |         34.0 |

The two kernels answer different questions, and neither is the default "for quality".
A cubic through four points is near-exact on a slow tone and falls away as the fourth power of frequency; the bank holds a floor across the band.
They cross at about `0.12 f_in`, so a narrow signal wants Hermite (and is ~10x faster for it), a wide one wants the bank.
The default stays Hermite, which is the existing behaviour.

No absolute figure is asserted in the test — only that ordering.
A dB threshold chosen to pass would measure nothing, and GR 3.10's 8-tap/128-arm MMSE bank is not a bar this can claim parity with, since none of its code was taken.

What sets the bank's floor turned out to be a defect rather than a property.
The arm interpolation stopped at the last arm instead of carrying on into arm 0 of the _next_ sample, so the top `1/n_phases` of every sample interval was flat — a dead zone that more arms merely made narrower.
That is invisible in a sweep across a sample boundary, because the discontinuity is at the far end of the arm, and it is why the floor appeared to follow the arm count:

| n_phases | n_taps | before |    after |
| -------: | -----: | -----: | -------: |
|       16 |    512 |   36.3 | **49.1** |
|       32 |   1024 |   43.4 |     49.1 |
|       64 |   2048 |   47.5 |     49.0 |
|      128 |   4096 |   48.7 |     49.0 |

With the wrap correct the arm count stops mattering (0.04 dB across 16–128) and 16 arms reach what previously took 128 — the same floor from a bank **8x smaller** to design, store and ship to a device.
What the floor actually follows is the length of each arm, ~6 dB per doubling, which is the stopband of the filter doing the work: at 16 arms, 4 taps/arm gives 25.6 dB, 8 → 38.2, 16 → 43.3, 32 → 49.1, 64 → 54.9.

`qa_DriftResampler` now asserts that shape — invariance in the count, ordering in the length — instead of the artefact.
`n_taps` is still the length of the _whole_ prototype, with each arm getting `n_taps / n_phases` of it; the two `Doc` strings say so, where they previously described `n_taps` as giving "sharper channel edges", which is the channelizer's story, not an interpolator's.

### Staleness of the DSP tables

Every channelizer, resampler, FFT and filter figure above was measured **before** the three framework commits at the tip of this branch.
Those commits remove per-`work()`-call overhead that every block pays, so the DSP numbers are a lower bound on what the branch now delivers, by an amount that depends on the chunk each benchmark hands the framework — negligible for the large-chunk device rows, more for the small-chunk host rows.
They have not been re-measured, and the cross-over points and ratios they are quoted for are unaffected, because both sides of each comparison move together.

### Distance from attainable

The DFT across the arms was evaluating `std::cos`/`std::sin` inside its inner loop.
Measured in isolation that cost ~6x the entire rest of the transform and dominated the whole filterbank.
The twiddle for the (channel, arm) pair `(k, p)` depends only on `(k*p) mod nChannels`, so `nChannels` precomputed entries cover the whole grid; stepping that index beats both the live evaluation and a full nChannels x nChannels table:

| M    | live angle | NxN table | N-entry, stepped |
| ---- | ---------: | --------: | ---------------: |
| 8    |    60.6 ns |   11.9 ns |   10.2 ns (5.9x) |
| 16   |   123.2 ns |   24.8 ns |   19.4 ns (6.4x) |
| 1024 |  9218.6 ns | 1788.7 ns | 1128.1 ns (8.2x) |

Applied, that moved the host figures 2.0–4.9x (flat bank at 16 channels, measured on the same harness) and left the GPU unchanged — a GPU evaluates sine in hardware, so it was never bound by this.
Host and device agree bit-for-bit on every channel of an 8- and 16-channel bank.

What remains is a formulation change, and it was measured rather than counted.
The transform is still an O(nChannels^2) direct DFT; an FFT across the arms is O(log nChannels) per sample.
Operation counts suggest 8x at 16 channels, but a radix-2 FFT carries a dependency chain and a bit-reversal that the flat strided sum does not, so the measured figure for producing **all** channels of one output set is:

| channels | direct DFT | radix-2 FFT | speedup |
| -------- | ---------: | ----------: | ------: |
| 8        |    57.3 ns |     41.7 ns |   1.38x |
| 16       |   226.0 ns |     84.7 ns |   2.67x |
| 64       |  3454.0 ns |    462.1 ns |   7.48x |
| 256      | 56304.4 ns |   2448.6 ns |  22.99x |

Those figures are the transforms ALONE, and that turns out to matter: `transformSetPlanar` keeps real and imaginary parts in separate arrays precisely so it vectorises, while `SimdFFT`'s complex transform wants them interleaved, so a real implementation pays a planar-to-interleaved conversion per output set.
Re-measured WITH that conversion (g++-15 -O3 -march=native, 20 000 reps after warm-up):

| M   | direct planar DFT | FFT + conversion | FFT alone | speed-up, conversion included |
| --- | ----------------: | ---------------: | --------: | ----------------------------: |
| 8   |           20.5 ns |                — |         — |  unusable: below `minSize` 16 |
| 16  |           50.4 ns |          53.3 ns |   31.9 ns |             **0.95x, slower** |
| 32  |          123.4 ns |         110.1 ns |   65.8 ns |                         1.12x |
| 64  |          423.0 ns |         135.6 ns |  105.0 ns |                     **3.12x** |
| 256 |         9768.1 ns |         486.6 ns |  427.8 ns |                    **20.07x** |

The FFT-alone column agrees with the table above (31.9 ns against 84.7 ns is the same ordering at a different scale), so the two measurements are consistent — the conversion is what changes the verdict.
**The crossover is M ~ 32-64, not 16**, and at the 8 and 16 channels this PR benchmarks there is no win available at all: `SimdFFT<float, Complex>::minSize()` is 16, so 8 cannot use it, and 16 comes out slower.

Where it does pay is high channel counts, which `n_channels` admits (up to 1024) — and above `kMaxPlanarChannels = 256` the planar direct path is not even built, so the FFT is the only option there.
Named as the next step, not taken here.

How much of that would actually reach the block depends on what share of it the transform is, which was measured by replacing each stage of the flat bank's loop with a stand-in of the same shape and taking the delta (at the benchmark's own `n_taps = 32` and `outputs_per_frame = 64`, so the working set matches):

| stage                       | 8 channels | 16 channels |
| --------------------------- | ---------: | ----------: |
| per-arm FIR                 |     57.9 % |      34.5 % |
| cross-channel DFT           |     30.5 % |  **49.4 %** |
| scatter to the output ports |      8.2 % |       4.3 % |

The transform overtakes the arms between 8 and 16 channels, which is what the complexities predict: the DFT is O(M²) per set while the arms are O(M · taps-per-arm), and a fixed _total_ tap count means each arm gets shorter as M grows.

An earlier revision of this section folded the transform-only speed-ups into these shares and claimed roughly 1.09x at 8 channels and 1.45x at 16 for the whole block.
**That was wrong and is withdrawn**: it costed the FFT without the planar-to-interleaved conversion a real implementation needs.
With the conversion counted there is no block-level win at either size — the transform is unusable at 8 and slower at 16.
The shares above still say where the time goes; what they do not say is that an FFT can claim it at these M.

Worth noting the scatter is **not** the problem (4–8 %), which retires the obvious suspicion that writing one sample at a time into M separate port rings is what costs.

Fuller per-backend and block-level tables are in the working notes.

</details>

## Breaking changes

- **`work()` takes the compute backend by reference.** `BlockModel`'s vtable changes, so **plugins compiled
  against the previous core must be rebuilt**, and any out-of-tree block overriding `work()` will not compile
  until updated. Six in-tree blocks needed that change.
- **`fir_filter` changes sample semantics** — it is a window rather than a 1:1 `processOne`, so its output
  stream is shorter by `b.size() - 1`. This is the one a user sees in their data.
- A required domain inside a sub-graph now **refuses** where it previously ran on whatever answered.
- `Stride` is applied per window for the view form of `processBulk`.
- A chunk larger than the edge carrying it is refused at start rather than stalling.
- `compute_domain` is frozen while a block runs.
- `EdgeParameters` resources default to `nullptr` ("unset, inherit") rather than to a concrete resource.
- Per-block ABI grows by a device shadow and two pointers; the exact figure needs re-measuring.

<details>
<summary>Verification</summary>

| lane                  | build            | tests   | sanitizer |
| --------------------- | ---------------- | ------- | --------- |
| AdaptiveCpp Debug     | clean            | 151/151 | n/a       |
| gcc-15 Release + ASAN | 0 errors, 0 warn | 148/148 | 0 reports |
| gcc-15 Debug + ASAN   | 0 errors         | 148/148 | 0 reports |
| Emscripten Release    | 0 errors         | 96/96   | n/a       |

**Scope of the sanitized lanes, stated rather than implied:** AdaptiveCpp is incompatible with ASan/UBSan/TSan, so the gcc-15 lanes build `GR_DEVICE_HAS_SYCL == 0` and exercise the _host fallbacks_ of the device headers.
Device dispatch has no sanitized lane anywhere.
ASAN reported nothing at runtime; all eight defects were compile-time.

Two hardware SDR tests (`qa_RTL2832Source`, `qa_SoapyIntegration`) are excluded from the 148: ASAN's 2–4× slowdown breaks their real-time capture assertions — stated as lost coverage rather than hidden.

</details>

<details>
<summary>AOB</summary>

- Issue #161 yield is narrow and should not be overstated: this crosses off **Rational Resampler**, and
  arguably **Fractional Resampler** (`DriftResampler` fills the role with cubic Hermite rather than GR's MMSE
  polyphase interpolator, so it is not numerically equivalent). **No Channelizer is crossed off** — but the
  polyphase arithmetic and overlap-save convolution the whole category needs now exist.
- A separate build-system defect was found and written up, unrelated to this branch: the block-lib generator
  needs two configure→build cycles to converge from a cold directory, because generated sources land after
  CMake has fixed each target's source list. Reproduced in two independent build directories.
- `ci.yml` passes the deprecated misspelled `GNURADIO_PARSE_REGISTRATIONS_TOOL_CXX_COMPLILER`; the project's
  own CMake warns and asks for the corrected spelling.

</details>

---

Runtime and compile-time block merging, and a JIT path, build directly on these interfaces.
What they remove is the per-block bookkeeping the tables above quantify — the reason a chain of individually cheap blocks reads well below its own ceiling.

Subtle understatement having apparently left the building:

image

<sub>30k lines, most of them tests. The monolith is doing a lot of work in this metaphor, and so is the word "beta".</sub>


core/include/gnuradio-4.0/device/DeviceContext.hpp
```diff
@@ -0,0 +1,201 @@
...
+    /// count, not a duration -- "one barrier per group `work()`, not one per block" -- so it is observable rather
+    /// than inferred from a stopwatch. Implementations bump it wherever they block. Atomic because a context is
+    /// process-global and every scheduler thread dispatching through it bumps the same counter.
+    [[nodiscard]] std::uint64_t syncCount() const noexcept { return _syncCount.load(std::memory_order_relaxed); }
+
+protected:
+    std::atomic<std::uint64_t> _syncCount{0UZ};
+
+public:
+    // true when `ptr` is memory this context's device can dereference directly (USM); conservatively false
```
the-argus: this comment should be on below function, isDeviceAccessible?

core/include/gnuradio-4.0/device/DeviceContext.hpp
```diff
@@ -0,0 +1,201 @@
...
+    [[nodiscard]] std::string   name() const override { return "CPU fallback"; }
+    [[nodiscard]] std::string   version() const override { return "host"; }
+
+    void wait() override {}
+
+    static constexpr std::size_t kAlign = alignof(std::max_align_t);
+
+    // invalid-on-failure rather than null-on-failure, so every backend shares one contract
+    [[nodiscard]] DeviceBuffer allocate(std::size_t bytes, std::size_t align, Residency wanted) override {
+        if (wanted != Residency::host && wanted != Residency::shared) {
```
the-argus: Hm, I imagine a scenario where I want to write a block which has some wrapper functions such that it can compile either for software or for cuda. In this case it might be desirable for the cpu context to pretend to be capable of supplying host and shared memory.

core/include/gnuradio-4.0/device/DeviceContextRegistry.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+namespace gr::device {
+
+/**
+ * Singleton registry mapping `compute_domain` strings to device contexts.
+ *
+ * Contexts are keyed by canonical name, so every spelling of one device names one context. A domain that no
+ * rung of the ladder serves yields nullptr — the caller decides whether that is a skip or an error.
+ */
+class DeviceContextRegistry {
```
the-argus: This class has a mutex, it seems the only use case that would be needed for is if new compute domains appear dynamically throughout a program's lifetime (but a compute domain shutting down/disappearing is not supported by this).

If it is okay to say all domains are know up front, maybe there could be `void setInstance(DeviceContextRegistry*) { assert(!instance); ... }` and a public constructor `DeviceContextRegistry(Map<string, unique_ptr<DeviceContext>>)`, and then no mutex.

drslebedev (approved):
Thanks a lot for this PR! This is a huge amount of work, both in the implementation itself and in bringing CPU and GPU processing together within the same framework.

I’ve gone through the main architectural and API changes, particularly the changes in Block and work(), how device execution is selected and dispatched, and how buffers and CPU/GPU transfers are handled. I also looked through several of the new DSP blocks to understand how these mechanisms are used in practice.

From what I’ve reviewed, the overall architecture looks consistent. I particularly like the new processBulk(DeviceContext&, ...) overload: it gives block authors explicit control over device execution when needed, while simpler blocks can use
automatic dispatch. The API feels reasonably straightforward given how much flexibility it provides. 

Being able to keep data on the GPU across several processing blocks also looks useful for real DSP chains.

I also ran the relevant tests and benchmarks on my machine. That really helped me get a practical impression.

Given the size of this PR and the time available, I haven’t reviewed every line or explored every execution path in depth. My review has mainly focused on understanding the design, the API, and how the main pieces fit together. I expect some of the finer details and trade-offs to become clearer through more extensive use in different flowgraphs.

With that review scope in mind, I’m happy to approve this PR. Thanks again for all the effort that went into it!

